PR #11, opened 2016-07-06
Fix typos

pyk:


waruqi:
Oh, thanks for your commit!

I have merged this commit into branch: sitesrc

The branch gh-pages is readonly : )

PR #38, opened 2016-10-15
Fixed a problem about name-clash of object file

avaicode:
There is a not serious problem in your build system when building projects blending with both c and c++ ,I think.
It is that when c source and cpp source have the same base name,one of their object file will be overwrited by another,so the build will fail because of some missing reference.I've prepared a example to make it clearer: [testXmake.tar.gz](https://github.com/waruqi/xmake/files/531298/testXmake.tar.gz).
And because it's a small problem.I fixed it myself instead of writing an issue.It may work and help.

### DIFF
--- a/tests/merge_object/xmake.lua
+++ b/tests/merge_object/xmake.lua
@@ -41,5 +41,5 @@ target("test")
 
     -- add files
     add_files("src/test.c") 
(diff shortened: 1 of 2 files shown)


waruqi:
Oh, I have merged this patch.
 
thanks very mush! 👍

PR #61, opened 2017-03-26
safer 'install' task

TitanSnow:
This pull request is trying to fix a problem about *security*

When installing something, users take it for granted to give command higher permission. I think most users use <code>sudo xmake install</code> instead of <code>xmake install</code>, because installing will put files into paths like <code>/usr/local</code> and root permission is required.

That is the problem. I found xmake will always build the target before installing even if it has been built before. And the building is a *latent danger*. For example, a bad man replace the compiler like gcc, and while installing, the fake gcc will get root permission to do something bad.

There are two solutions to avoid this. One is to set permission lower by process self before actual installation. Another is to do *least* things with high permission. My solution is to let task 'install' avoid building and use 'safe_config' which will not reconfig instead of 'config'. When users directly do task 'install' without building before, they will get a note to do building.

Also, except security, the build by root will generate result files with wrong owner. On my computer, after installation, I couldn't even delete 'build' dir without <code>sudo</code>

My solution may be not perfect enough. *Please do not merge easily*. But the secure problem must be fixed.

### DIFF
--- a/xmake/actions/install/main.lua
+++ b/xmake/actions/install/main.lua
@@ -85,15 +85,28 @@ function main()
     -- init finished states
     _g.finished = {}
 
-    -- build it first
-    task.run("build", {target = targetname})
+    task.run("config", {target = targetname})
 
     -- install all?
-    if targetname == "all" then
-        for _, target in pairs(project.targets()) do
-            _install_target_and_deps(target)
-        end
-    else
-        _install_target_and_deps(project.target(targetname))
-    end
+    try{
+        function ()
+            if targetname == "all" then
+                for _, target in pairs(project.targets()) do
+                    _install_target_and_deps(target)
+                end
+            else
+                _install_target_and_deps(project.target(targetname))
+            end
+        end,
+        catch
+        {
+            function (errors)
+                -- print user-friendly notes
+                cprint("${bright red}error: ${default red}installation fail. may it hasn't built before or permission denied")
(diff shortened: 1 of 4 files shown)


waruqi:
Ok, I'll find a better solution to fix this problem in the next few days. 

Thanks. : )

TitanSnow:
I have found an example: if run linuxbrew as root, linuxbrew will give a note then stop

```console
$ sudo brew
W: be careful as root.

========================================================================
You may want to update following environments after installed linuxbrew.

  PATH, MANPATH, INFOPATH 

(example: /usr/share/doc/linuxbrew-wrapper/examples/profile)
========================================================================
Don't run this as root!
```

waruqi:
Ok, thanks for your advice.

I am solving this problem, please wait some time. : )

waruqi:
You can try [235bb96](https://github.com/tboox/xmake/commit/235bb967883ab0686bab731c8e94a4a0bb7677ab) from dev branch first.

And I will continue to improve some details in future.

TitanSnow:
Would you add a note displaying like brew?

waruqi:
Ok, I will add it.

waruqi:
I have add it for checking root.

You can see the following tips:

```bash
$sudo xmake
```
```
error: Running xmake as root is extremely dangerous and no longer supported.
        As xmake does not drop privileges on installation you would be giving all
        build scripts full access to your system.
```

TitanSnow:
I see

Issue #60, opened 2017-03-26
couldn't self-build


TitanSnow:
```console
~/xmake/core$ xmake
error: /usr/local/share/xmake/core/base/option.lua:94: assertion failed!
stack traceback:
	[C]: in function 'assert'
	/usr/local/share/xmake/core/base/option.lua:94: in function '_context'
	/usr/local/share/xmake/core/base/option.lua:787: in function 'options'
	/usr/local/share/xmake/core/base/option.lua:745: in function 'get'
	/usr/local/share/xmake/core/base/utils.lua:37: in function '_print'
	/usr/local/share/xmake/core/base/utils.lua:78: in function 'cprint'
	/usr/local/share/xmake/core/base/utils.lua:118: in function 'error'
	/usr/local/share/xmake/core/main.lua:113: in function </usr/local/share/xmake/core/main.lua:105>
```

Is there any other work needed to do to self-build?

waruqi:
It works here.

Can you run `xmake --version` to tell me your xmake version? 

And run `xmake -v --backtrace` to get more verbose backtrace info, or please attempt to update the newest version.

Thanks!

TitanSnow:
Sorry, my mistake

waruqi:
Ok, I push a patch to master and improve error tips in order to locate problem more quickly in future.

TitanSnow:
I'm trying to let xmake self-build on CIs. But it seems that only Linux could do this.

Branch: [TitanSnow:self_build](https://github.com/TitanSnow/xmake/tree/self_build)

[Build on travisCI](https://travis-ci.org/TitanSnow/xmake/builds/215363101), Linux succeeded but Mac OS failed.

[Build on appveyor](https://ci.appveyor.com/project/TitanSnow/xmake/build/1.0.15), failed on Windows

waruqi:
I have fixed self-build on macOS, you can see thee patch [d3865a3](https://github.com/tboox/xmake/commit/f9ff6e5e50c26501964a3b35279aa4881d3865a3) from dev branch.

I need modify and rebuild libluajit.a to support vs201x compiler and it only support vs200x compiler now.

Please wait some time. I will fix it in future.

Thanks!

TitanSnow:
It seems that xmake uses binary library file to satisfy dependencies. Will it be better to build dependencies from source together?

waruqi:
I will consider this suggestion.

Thanks

waruqi:
[...]

PR #63, opened 2017-03-30
add global makefile intending to replace unsafe install script

TitanSnow:
Danger of binding building and installing together is explained clearly in #61 . The install script seems to take the same risk.
```console
$ sudo ./install
```
So a two-step makefile might be better
```console
$ make build
$ sudo make install
```

This pull also includes new self-build CI yaml file. It seems that you nearly finish #60 's work. @waruqi


waruqi:
Great,  I have merged. 👍 

Thanks!

PR #65, opened 2017-03-31
add global interface "set_default_target"

TitanSnow:
*This pull is a feature adding.*

Sometimes I don't want end-users to build all targets by default. Just like my libBG project, the "bg" target for building library is main, and other targets for building test program is meaningless to end-users. So I think it would be better to add a interface to let people like me choose default target manually.

This "set_default_target" is applied to task "build", "package", "install" and "run". I do not apply it to other task like "uninstall" because it might trouble end-user. (They always like to uninstall everything)

Design is that just do such work to <code>xmake.lua</code> just like <code>set_project</code> is OK:
```lua
set_default_target("bg")
```
Then the default target will be "bg" instead of "all".

### DIFF
--- a/xmake/actions/build/main.lua
+++ b/xmake/actions/build/main.lua
@@ -70,7 +70,7 @@ function main()
         function ()
 
             -- build 
-            builder.build(targetname or "all")
+            builder.build(targetname or project.default_target() or "all")
         
         end,
 
--- a/xmake/actions/build/xmake.lua
+++ b/xmake/actions/build/xmake.lua
@@ -49,7 +49,7 @@ task("build")
                 ,   {'j', "jobs",       "kv", "4",          "Specifies the number of jobs to build simultaneously"          }
                
                 ,   {}
-                ,   {nil, "target",     "v",  "all",        "Build the given target."                                       } 
+                ,   {nil, "target",     "v",  nil,          "Build the given target."                                       }
                 }
             }
 
--- a/xmake/actions/install/main.lua
+++ b/xmake/actions/install/main.lua
@@ -39,7 +39,7 @@ function main()
     task.run("build", {target = targetname})
 
     -- get the target name again
-    targetname = targetname or project.default_target()
+    targetname = targetname or project.default_target() or "all"
 
     -- trace
     print("installing to %s ...", option.get("installdir") or platform.get("installdir"))
--- a/xmake/actions/install/main.lua
+++ b/xmake/actions/install/main.lua
@@ -25,6 +25,7 @@
 -- imports
 import("core.base.option")
 import("core.project.task")
+import("core.project.project")
 import("core.platform.platform")
 import("install")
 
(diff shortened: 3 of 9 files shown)


waruqi:
You can pass the given target name to build or install it.

For example:

```bash
# only build target: bg
$ xmake [-b|--build] bg

# only rebuild target: bg
$ xmake [-r|--rebuild] bg

# only install target: bg
$ xmake install bg 

# only uninstall target: bg
$ xmake uninstall bg 
```

waruqi:
I have added `set_default` api to each target and you can use this interface to set multiple default targets.

Please try it from the lastest commit from dev branch.

For example:

```lua
target("test1")
    set_default(false)

target("test2")
    set_default(true)

target("test3")
    ...
```

If you run commands: `xmake`, `xmake install`, `xmake package`, `xmake run`, the behavior is

| target | behavior                             |
| ------ | -------------------------------- |
| test1  | will be not built and installed .. by default |
| test2  | will be built and installed .. by default      |
| test3  | will be built and installed .. by default        |

And you can also force to build and install all targets:

```bash
$ xmake [-a|--all]
$ xmake install [-a|--all]
$ xmake run [-a|--all]
```

I do not apply it to other task like "xmake uninstall", "xmake clean".

: )

TitanSnow:
Good but I think specifying for each target is a bit cumbersome. Why not a global setting interface? If multiple default has meanings, I think it would be simpler to be <code>set_default_targets</code> instead of specifying in each target. Some projects might have tons of targets, But the main targets are always not too many. :thinking:

waruqi:
I don't want to add too much apis to global scope and this interface name is a little long. `set_default` only affects the target scope. : ) 

You don't need to set default for each target. All targets will be built by default if don't call `set_default` to set default property.This behavior will be compatible with older versions.
 
You can only call `set_default` to disable the given target. For example:

```lua
-- it will be built and install by default
target("test1")
    ...

-- it will be built and install by default
target("test2")
    ...

-- only `test3` target will be not built and install by default
target("test3")
    set_default(false)
```

You need only to set one target.

TitanSnow:
I know your meaning, but I'm afraid that you don't get my point. My meaning is that *to disable target is used more usual than to enable a target,* so api to enable could only set once but to disable need to set many times.

waruqi:
I know, but I think that to enable target is used more usual than to disable a target. : )

TitanSnow:
All right, maybe you're right :smiley_cat:

TitanSnow:
I think <code>xmake build all</code> should be kept. It's simpler and I'm used to building all with this.

```console
$ xmake build all
error: assertion failed!
```

waruqi:
You can try run `xmake -a` or `xmake --all`.

TitanSnow:
I see

waruqi:
And then I thought about it. You can also disable all targets in the root scope and only enable target you need. for exampe:

```lua
-- disable all targets by default
set_default(false)

-- only `test1` target will be built and install by default
target("test1")
    set_default(true)

-- other targets will be not built and install by default
for _, name in ipairs({"test2", "test3", "test4" ... }) do
    target("testX")
        ...
end
```

TitanSnow:
:smile_cat:

PR #77, opened 2017-04-29
fix sandbox_utils.cprint no color reset eol

TitanSnow:
```console
$ xmake create -P test
$ cd test
$ xmake
$ xmake run
[100%]: linking.release hw
hello world!    # This line is magenta! Last line's color is not reset!
```

The bug is that sandbox_utils.cprint does not reset color at end of line


waruqi:
Thanks, this problem I have fixed on `repo` branch, but not yet merged to `dev` branch.

TitanSnow:
😺

PR #78, opened 2017-04-30
soft-wrap `help`

TitanSnow:
in standard 80x40 terminal, `xmake --help` print help info will overflow in horizontal which not looks nice

I add feature to get terminal size in runtime and soft-wrap the help printing

It has only tested on linux. If it's able to work in macos, please add the macos define check in `getwinsize.c` by directly push here


TitanSnow:
Is that available on macos?

waruqi:
@TitanSnow Great, it works on macos and I also implement it on windows. 😸

TitanSnow:
:+1:

uael:
love it :+1:

PR #84, opened 2017-05-10
replhistory

TitanSnow:
1. keep repl command history
2. add module `lib.readline`
3. add fields `host`, `arch`, `nuldev` to `os.versioninfo()`


TitanSnow:
Please have a wait, there's a problem on self-building. I'll fix it right away

xmake/plugins/lua/xmake.lua
```diff
@@ -56,8 +61,33 @@ task("lua")
...
+                -- load history
+                replhistory = history.load("replhistory") or {}
+                for _, ln in ipairs(replhistory) do
+                    readline.add_history(ln)
+                end
+            end
+
             -- enter interactive mode
             sandbox.interactive()
+
```
waruqi: Why not handle them in `sandbox.interactive()`?  I don't want to expose too many interfaces that can't cross the platform in sandbox.

TitanSnow: There are some reason for it
* I wanna reuse `history` API to store replhistory
* Duty should be clear. `sandbox.interactive()` should only make sandbox interactive but not history store. History storing is the job of upper level
* I have considered what you think. So I put readline into `lib.readline` but not in `core` as an optional library. In this way it's clear and give enough info that this is **optional** and not available always. That's why I open a new namespace `lib`

TitanSnow:
I'll fix build on macos today

xmake/plugins/lua/xmake.lua
waruqi: You can use `require("lib/readline")` and `require("base/history")` to handle history in [sandbox_core_sandbox.interactive()](https://github.com/tboox/xmake/blob/master/xmake/core/sandbox/modules/import/core/sandbox/sandbox.lua) and simplify the implementation in `lua action` as far as possible.

waruqi: If we don't look at implementation of `sandbox.interactive()`, we may not know that `readline` and `sandbox.interactive()` have a relationship in lua action sandbox.

TitanSnow: I'll have a try

waruqi: Thanks! 😄

xmake/core/sandbox/modules/import/core/sandbox/sandbox.lua
```diff
@@ -47,8 +48,34 @@ function sandbox_core_sandbox.interactive()
...
+        if #entries > #replhistory then
+            for i = #replhistory+1, #entries do
+                history.save("replhistory", entries[i].line)
+            end
+        end
+
+        -- clear history
+        readline.clear_history()
+    end
 end
```
waruqi: 👍

TitanSnow:
Done. I have also modified CI scripts -- make them simpler and better

waruqi:
great! 👍

core/src/xmake/readline/clear_history.c
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+// clear_history wrapper
+tb_int_t xm_readline_clear_history(lua_State* lua)
+{
+    // check
+    tb_assert_and_check_return_val(lua, 0);
+
+    // call clear_history
+    clear_history();
```
waruqi: It will crash on macOS here.

```bash
ruki:~ ruki$ lldb /usr/local/share/xmake/xmake 
(lldb) target create "/usr/local/share/xmake/xmake"
Current executable set to '/usr/local/share/xmake/xmake' (x86_64).
(lldb) run lua
Process 32656 launched: '/usr/local/share/xmake/xmake' (x86_64)
Process 32656 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007fffa48c6e13 libedit.3.dylib`history + 912
libedit.3.dylib`history:
->  0x7fffa48c6e13 <+912>: movq   (%r13), %rdi
    0x7fffa48c6e17 <+916>: movq   %r15, %rsi
    0x7fffa48c6e1a <+919>: callq  *0x48(%r13)
    0x7fffa48c6e1e <+923>: xorl   %r15d, %r15d
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007fffa48c6e13 libedit.3.dylib`history + 912
    frame #1: 0x00007fffa48bc569 libedit.3.dylib`clear_history + 31
    frame #2: 0x00000001000049dd xmake`xm_readline_clear_history(lua=0x0000000000084378) at clear_history.c:50
    frame #3: 0x00000001000a0eb6 xmake`lj_BC_FUNCC + 52
    frame #4: 0x0000000100024102 xmake`lua_pcall(L=0x0000000000084378, nargs=0, nresults=1, errfunc=-2) at lj_api.c:1052
    frame #5: 0x00000001000011b7 xmake`xm_machine_main(machine=0x0000000000400bc0, argc=2, argv=0x00007fff5fbff760) at machine.c:538
    frame #6: 0x0000000100000ad3 xmake`main(argc=2, argv=0x00007fff5fbff760) at xmake.c:22
    frame #7: 0x0000000100000a64 xmake`start + 52
(lldb) f 2
frame #2: 0x00000001000049dd xmake`xm_readline_clear_history(lua=0x0000000000084378) at clear_history.c:50
   47  	    tb_assert_and_check_return_val(lua, 0);
   48  	
   49  	    // call clear_history
-> 50  	    clear_history();
   51  	
   52  	    // pushnil
   53  	    lua_pushnil(lua);
(lldb) 

```

xmake/core/sandbox/modules/import/core/sandbox/sandbox.lua
```diff
@@ -47,8 +48,34 @@ function sandbox_core_sandbox.interactive()
...
+
+    if enable_readline then
+        -- save to history
+        local entries = readline.history_list()
+        if #entries > #replhistory then
+            for i = #replhistory+1, #entries do
+                history.save("replhistory", entries[i].line)
+            end
+        end
+
```
waruqi: I think maybe never be reached here, because I often quit `xmake lua` using `<C-C>`.

So I still feel that It should save the global history in the following place:
 
```c
#ifdef XM_CONFIG_API_HAVE_READLINE
    // get line
    tb_char_t const* line = readline(prompt);
    if (line)
    {
        // add line to history
        add_history(line);

       // save current history to file
       // ..
```

xmake/core/sandbox/modules/import/core/sandbox/sandbox.lua
```diff
@@ -28,6 +28,7 @@ local sandbox_core_sandbox = sandbox_core_sandbox or {}
 -- load modules
 local sandbox   = require("sandbox/sandbox")
 local raise     = require("sandbox/modules/raise")
+local history   = require("project/history")
```
waruqi: This project history module will generate `.xmake` directory in the current directory, but `xmake lua` is a global action. It does not depend on project.
 
I think it would be better to save repl history in the global directory. .e.g `~/.xmake/cache/history_repl`

core/src/xmake/readline/clear_history.c
TitanSnow: I'll look into it

xmake/core/sandbox/modules/import/core/sandbox/sandbox.lua
TitanSnow: Right

core/src/xmake/readline/clear_history.c
TitanSnow: Always or sometimes? @waruqi

waruqi: Always

Issue #69, opened 2017-04-10
Remote package management


uael:
> A package manager or package management system is a collection of software tools that automates the process of installing, upgrading, configuring, and removing computer programs for a computer's operating system in a consistent manner.

> A package manager deals with packages, distributions of software and data in archive files. Packages contain metadata, such as the software's name, description of its purpose, version number, vendor, checksum, and a list of dependencies necessary for the software to run properly. Upon installation, metadata is stored in a local package database. Package managers typically maintain a database of software dependencies and version information to prevent software mismatches and missing prerequisites. They work closely with software repositories, binary repository managers, and app stores.

> Source: [package manager](https://en.wikipedia.org/wiki/Package_manager)

Remote package management roadmap:
- [x] detect tools and environment and wrap them
  - [x] git
  - [x] curl
  - [x] wget
  - [x] unzip/tar/7z/gzip
  - [x] ping (test speed)
- [x] Extractor wrapper
  - [x] tar 
  - [x] unzip
  - [x] gzip
  - [x] 7z
- [x] Cache management
- [x] semver
  - [x] parse
  - [x] satisfies
  - [x] select
- [ ] Multi language
  - [x] asm/c/c++
  - [x] objc/c++
  - [x] swift
  - [x] rust
  - [x] dlang
- [x] Multi platform
  - [x] linux
  - [x] macOS
  - [x] windows 
    - [x] git
    - [x] downloader (http[s], ftp ..)
    - [x] extractor (zip, tar.gz, 7z .. )
- [x] Command line tools
  - [x] require task
    - [x] install
      - [x] parse require info
      - [x] parse and load packages from repositories
      - [x] download package
        - [x] optimization: select the fastest url if exists multiple urls
      - [x] extract package
      - [x] build package 
        - [x] xmake.lua
          - [x] linux/macos
          - [x] windows
        - [x] makefile
          - [x] linux/macos
          - [x] windows
        - [x] configure/make
          - [x] linux/macos
        - [x] cmakelists.txt
          - [x] linux/macos
          - [x] windows (in progress ..)
        - [x] *.vcproj
          - [x] windows
      - [x] install package
      - [x] install basic dependent tools (git, curl, unzip ...)
    - [x] search
    - [x] clear
    - [x] remove
    - [x] list
  - [x] repo task
    - [x] add repository
    - [x] remove repository
    - [x] list all repositories
    - [x] clear all repositories and cache
    - [x] global or local repositories
- [x] xmake lua api
[...]

PR #98, opened 2017-05-18
semver

uael:
- [x] semver:resolve
- [x] add more tests


uael:
#69

waruqi:
@uael I found `semver.select()` will return version 1.7.0 when I test `add_requires("https://github.com/glennrp/libpng.git@libpng >=1.6.28")`. But the version tags of libpng is `v1.7.0beta87 ...`.

How to get the original version string in the returned version, for exmaple:

```
local version, source = semver.select(">= 1.6.28", {}, {"v1.7.0beta87"}, {})

version = 
{
    version = 1.7.0
    origin = v1.7.0beta87  <- return raw version string
    major = 1
    minor = 7
}
```

waruqi:
@uael If I choose master but no branch, it would be better to choose the newest version, rather than simply returning the empty

```lua
local version, source = semver.select("master", {"1.8.0, 1.9.0"}, {"1.7.0", "1.6.0"}, {})
```

return version: 1.9.0

uael:
@waruqi hum 1.7.0beta87 isn't a valid version it should be 1.7.0-beta+87
Ok for master you're right 👍

waruqi:
@uael > hum 1.7.0beta87 isn't a valid version it should be 1.7.0-beta+87

But there's a lot of irregular version numbers in git tags. For an irregular version number, if it is not handled, it should return empty, and if it can be handled, the original version string should be returned. Otherwise `git checkout` or download url will fail because I don't know if the returned version is valid.

for example, the returned version is `v1.7.0`, but I `git checkout v1.7.0` fails.

If the empty version `{}` is returned, I can at least ignore it, or return to the original version string `"v1.7.0beta87"` and I can checkout it normally.

```
> semver.select(">=1.6.28", {}, {"v1.7.0beta87"}, {})

{
    patch = 0
,   minor = 7
,   version = v1.7.0
,   major = 1
,   build = 
    {
    }

,   prerelease = 
    {
    }

}
 tags
```

uael:
@waruqi I've tested and fixed most of the libsv, i will continue the xmake semver with latest changes once I release libsv as soon as possible.

uael:
I dunno how to hanlde bad tags but I think you're right, the goal is to get files with a version as close as possible to the user input, which include bad versions

waruqi:
@uael  Yes, and the returned version(`v1.7.0`)  maybe correct in this example, It's very similar to `v1.7.0beta87`.  Can you return this user input string `v1.7.0beta87` at same time when return `v1.7.0`

uael:
np :smiley:

uael:
About the resolver you should just concatenate range when you got a dependency twice like, a want d to be >1.5 and b want d to be <2 then you select:
```lua
semver.select(">1.5 <2", {...}, {...}, {...})
```

uael:
@waruqi 
```lua
> semver = import("core.base.semver")
> semver.select(">=1.6.28", {}, {"v1.7.0beta87"}, {})

{
    patch = 0
,   raw = v1.7.0beta87
,   version = v1.7.0beta87
,   major = 1
,   build = 
    {
    }

,   minor = 7
,   prerelease = 
    {
        beta87
    }

}
 tags
```

waruqi:
@uael 
> About the resolver you should just concatenate range when you got a dependency twice like, a want d to be >1.5 and b want d to be <2 then you select:

Ok, I will improve it. Thanks!

waruqi:
@uael Can I merge this PR now?

uael:
@waruqi Yep 😃

waruqi:
Ok 👌

waruqi:
@uael If you only use the bool type, this header file `<cstdbool>` does not seem to be necessary?When I use vs2008 to compile it as c++(`/TP`), this file not found!

```c
# ifdef __cplusplus
#  include <cstdbool>
# else
```

uael:
@waruqi If it work on 2010 I ca check for it
```
MSVC++ 14.1 _MSC_VER == 1910 (Visual Studio 2017)
MSVC++ 14.0 _MSC_VER == 1900 (Visual Studio 2015)
MSVC++ 12.0 _MSC_VER == 1800 (Visual Studio 2013)
MSVC++ 11.0 _MSC_VER == 1700 (Visual Studio 2012)
MSVC++ 10.0 _MSC_VER == 1600 (Visual Studio 2010)
MSVC++ 9.0  _MSC_VER == 1500 (Visual Studio 2008)
MSVC++ 8.0  _MSC_VER == 1400 (Visual Studio 2005)
MSVC++ 7.1  _MSC_VER == 1310 (Visual Studio 2003)
MSVC++ 7.0  _MSC_VER == 1300
MSVC++ 6.0  _MSC_VER == 1200
MSVC++ 5.0  _MSC_VER == 1100
```

waruqi:
@uael Is that okay? 

```c
# ifdef __cplusplus
#    if defined(_MSC_VER) && _MSC_VER >= 1600
#        include <cstdbool>
#    endif
# else
```

uael:
@waruqi Yep I think, but you're maybe right
```c
#ifndef __cplusplus
# if defined(_MSC_VER) && _MSC_VER < 1900
#   define bool	unsigned char
#   define true	1
#   define false	0
#   define __bool_true_false_are_defined	1
# else
#   include <stdbool.h>
# endif
#endif
```
should work

waruqi:
😄

uael:
#114

PR #104, opened 2017-05-20
specify statsfile's path

TitanSnow:
to avoid multi statsfile and not found

I found after ce27b8d coverage is decreased weirdly. That's because some `luacov.stats.out` are not found. I specified statsfile's path to make a single statsfile to be found


waruqi:
Great! 👍

PR #110, opened 2017-05-21
add tests for C part

TitanSnow:
Inc coverage of C part to 72.79%

There is problem with `process.open()` on travis-ci that I cannot reproduce local. That is, to open process with a nonexistent command, `tb_init_cmd` does not return `null`


waruqi:
> There is problem with process.open() on travis-ci that I cannot reproduce local. That is, to open process with a nonexistent command, tb_init_cmd does not return null

I will look at it.

waruqi:
It always return success and a valid pid when using `posix_spawnp` to open a nonexistent command on linux. 😢

TitanSnow:
But on my machine it passed tests

TitanSnow:
> RETURN VALUE
>
> Upon successful completion, posix_spawn() and posix_spawnp() place  the PID  of  the  child process in pid, and return 0.  If there is an error before or during the fork(2), then no child is created, the contents of *pid  are  unspecified,  and  these functions return an error number as described below.
>
> Even when these functions return a success status,  the  child  process may still fail for a plethora of reasons related to its pre-exec() ini‐tialization.  In addition, the exec(3)  may  fail.   In  all  of  these cases, the child process will exit with the exit value of 127.

From man page

Seems that just check if return code is 127 is ok. Or just use `fork` & `exec`

waruqi:
@TitanSnow 

> http://man7.org/linux/man-pages/man3/posix_spawn.3.html
>    When we try to execute a nonexistent command in the child, the
>     exec(3) fails and the child exits with a status of 127.

Yes, the cilld will return a status of 127. But we're not able to get this return value when opening process. Even with `fork` & `exec`, it seems we have no way to get the return state of a child process directly from the parent process.

So we can only get the status(127) of the child by `waitpid`. But this interface may be blocked for a while, so I can't call it in open process.

I think even if  the return value of `process.open` is not empty, it is normal. We just need to call `process.wait` to wait child process status.

[os.execv](https://github.com/tboox/xmake/blob/master/xmake/core/base/os.lua) has been handled well.

```lua
function os.execv(shellname, argv, outfile, errfile)

    -- open command
    local ok = -1
    local proc = process.openv(shellname, os.argw(argv), outfile, errfile)
    if proc ~= nil then
         ...
         waitok, status = process.wait(proc, 0)
         ...
```

So I don't think it's a bug and I'm going to remove the test code：

```lua
process.open("thismustnotbeaprogramname")
```

TitanSnow:
ok:ok_hand:

PR #112, opened 2017-05-22
better root checking & process redirect fixing

TitanSnow:
I write `os.uid()` to get uid instead of `id -u`

I modified the open mode of `process.open()` & `process.openv()`. Seems that the redirection has been fixed by this


TitanSnow:
Seems to have trouble on appveyor

waruqi:
Great! Some of the tests in the `tests/modules/os/` didn't go through on windows, but `assert` didn't give me more detail location information, so I tried to improve the `assert`.

waruqi:
`copy empty dir` seems to have some problems on windows. I will look at it.

TitanSnow:
It's better to let `assert` support context view

Issue #111, opened 2017-05-21
xmake root 权限安装问题


745275633:
xmake 有没有办法强制在 root 权限下运行？

TitanSnow:
See #61

745275633:
我需要使用 root 运行 xmake

TitanSnow:
#61 described why running xmake as root is deprecated

xmake *can* run as root. See `xmake/core/main.lua` line 163:

```lua
    -- check run command as root
    if not option.get("root") then
        local ok, errors = main._check_root()
        if not ok then
            utils.error(errors)
            return -1
        end
    end
```

That is, when option `root` is got, xmake will accept to run as root.

But, the option `root` has not been add to `common option`. That means we cannot use `xmake --root`. That should be a mistake. @waruqi 

After I do this modify, xmake accepts `--root`

```diff
diff --git a/xmake/core/project/task.lua b/xmake/core/project/task.lua
index 35c051ef..7bc08d6c 100644
--- a/xmake/core/project/task.lua
+++ b/xmake/core/project/task.lua
@@ -145,7 +145,8 @@ function task._translate_menu(menu)
                                                               , "    1. The Given Command Argument"
                                                               , "    2. The Envirnoment Variable: XMAKE_PROJECT_DIR"
                                                               , "    3. The Current Directory"              })
-        table.insert(options, 10, {})
+        table.insert(options, 10, {nil,"root","k",nil,"awd"})
+        table.insert(options, 11, {})
 
     end
 
```

**But, there are some problems. After I try running xmake as root, `/dev/null` becomes unwritable for common users. That might be a bug**

waruqi:
@745275633 我修改了下xmake，master上新版本支持临时允许root运行，通过命令： `sudo xmake --root f -c`, `sudo xmake --root` ，所有命令前加上 `--root`参数..

waruqi:
@TitanSnow 

> But, the option root has not been add to common option. That means we cannot use xmake --root. That should be a mistake. @waruqi

I have modified it as common option.

> But, there are some problems. After I try running xmake as root, /dev/null becomes unwritable for common users. That might be a bug

I have fixed it.

waruqi:
@745275633  see [#113](https://github.com/tboox/xmake/pull/113).

PR #113, opened 2017-05-23
privilege manage (escalation & degradation) when running as root

TitanSnow:
#111 

Privilege manage when running as root. Store the root privilege and degrade, when root privilege is needed, escalate privilege

With this, `sudo xmake install` is safe and ok

### DIFF
--- a/core/src/xmake/machine.c
+++ b/core/src/xmake/machine.c
@@ -83,6 +83,7 @@ tb_int_t xm_os_getwinsize(lua_State* lua);
 tb_int_t xm_os_versioninfo(lua_State* lua);
 #ifndef TB_CONFIG_OS_WINDOWS
 tb_int_t xm_os_uid(lua_State* lua);
+tb_int_t xm_os_gid(lua_State* lua);
 #endif
 
 // the path functions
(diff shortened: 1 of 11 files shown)


waruqi:
Great! 👍

waruqi:
@TitanSnow I do not understand why it is safe to do this, and now it seems that `sudo xmake` can always be passed.

TitanSnow:
@waruqi when running as root, module `privilege` will store the root privilege. That is

```
sudo xmake
uid: 0
gid: 0

privilege store it
uid: 1000 (for example)
gid: 1000

some actions need privilege, get
uid: 0
gid: 0
```

waruqi:
@TitanSnow This means that when the store privilege (uid, gid 1000), some operations that require root(uid,gid 0) will still fail. So it's safe when run `sudo xmake`. Is that right?

TitanSnow:
@waruqi The script inside xmake process could use `privilege.get()` to get root privilege. But the processes xmake spawns could not get root privilege. They run as normal user

TitanSnow:
@waruqi  When root privilege is needed, for example installation, call `privilege.get()` then after this xmake and spawned process will have root privilege

waruqi:
Got it! It's great!  👍 ❤️

TitanSnow:
launch xmake by `sudo xmake` -> has root privilege -> store it -> doesn't has root privilege -> spawn process like `gcc`, doesn't has root privilege, cannot get root privilege -> do installation -> `privilege.get()` -> has root privilege -> spawn process for installation, has root privilege

waruqi:
I understand it. Thanks! 😄

PR #114, opened 2017-05-23
semver update

uael:


waruqi:
👍

waruqi:
@uael When I run `xmake require "luajit >2.0.1"`, I found that the returned result seems to be incorrect.
 
```lua
semver.select(">2.0.1", {"2.0.0", "2.0.1", "2.0.2"}, {"v2.0.0", "v2.0.1", "v2.0.2", "v2.0.3", "v2.1.0-beta1", "v2.1.0-beta2"}, {"master", "v2.0", "v2.1"})
```

results:
```
{
    patch = 0
,   raw = v2.1.0-beta2
,   version = v2.1.0-beta2
,   major = 2
,   build = 
    {
    }

,   minor = 1
,   prerelease = 
    {
        beta2
    }

}
```

uael:
I'm writing this
```c
#include <stdlib.h>
#include <assert.h>

#include "semver.h"

#define STRNSIZE(s) (s), sizeof(s)-1

int main(void) {
  semver_t v0, v1, v2, v3, v4, v5, v6, v7, v8, v9, v10;
  semvers_t semvers = {0};

  semver_tryn(&v0, STRNSIZE("2.0.0"));
  semver_tryn(&v1, STRNSIZE("2.0.1"));
  semver_tryn(&v2, STRNSIZE("2.0.2"));
  v0.source = v1.source = v2.source = "version";
  semver_tryn(&v3, STRNSIZE("v2.0.0"));
  semver_tryn(&v4, STRNSIZE("v2.0.1"));
  semver_tryn(&v5, STRNSIZE("v2.0.2"));
  semver_tryn(&v6, STRNSIZE("v2.0.3"));
  semver_tryn(&v7, STRNSIZE("v2.1.0-beta1"));
  semver_tryn(&v8, STRNSIZE("v2.1.0-beta2"));
  v3.source = v4.source = v5.source = v6.source = v7.source = v8.source = "tag";
  semver_tryn(&v9, STRNSIZE("v2.0"));
  semver_tryn(&v10, STRNSIZE("v2.1"));
  v9.source = v10.source = "branch";

  semvers_push(semvers, v0);
  semvers_push(semvers, v1);
  semvers_push(semvers, v2);
  semvers_push(semvers, v3);
  semvers_push(semvers, v4);
  semvers_push(semvers, v5);
  semvers_push(semvers, v6);
  semvers_push(semvers, v7);
  semvers_push(semvers, v8);
  semvers_push(semvers, v9);
  semvers_push(semvers, v10);

  if (semvers.length != 11) {
    return EXIT_FAILURE;
  }
  if (semvers.capacity != 16) {
    return EXIT_FAILURE;
  }

  semvers_sort(&semvers);

  for (unsigned i = 0; i < semvers.length; ++i) {
    semver_fwrite(semvers.data + i, stdout);
    fprintf(stdout, ", %s\n", semvers.data[i].source);
  }

  semvers_rsort(&semvers);
  putc('\n', stdout);
  for (unsigned i = 0; i < semvers.length; ++i) {
    semver_fwrite(semvers.data + i, stdout);
    fprintf(stdout, ", %s\n", semvers.data[i].source);
  }

  return EXIT_SUCCESS;
}
```
which output
```bash
2.0.0, version
2.0.0, tag
2.0.0, branch
2.0.1, version
2.0.1, tag
2.0.2, version
2.0.2, tag
2.0.3, tag
2.1.0-beta1, tag
2.1.0-beta2, tag
2.1.0, branch

2.1.0, branch
2.1.0-beta2, tag
2.1.0-beta1, tag
2.0.3, tag
2.0.2, version
2.0.2, tag
2.0.1, version
2.0.1, tag
2.0.0, version
2.0.0, tag
2.0.0, branch
```
Did you think it's a good solution and if so what's the priority btw versions/branches/tags ?

uael:
We should add a stability setting to avoid using version with prerelease like:
```lua
target "a"
  set_minimum_stability "stable" -- or dev or something else
```

waruqi:
@uael 

> Did you think it's a good solution and if so what's the priority btw versions/branches/tags ?

I think their priority  is `versions > tags > branches`.

`versions` is defined in `xmake.lua` file of package. These versions are tested, so they are the most stable and their priority is highest.

If there is no `versions`, we will try to find the version from tags and branches.

> We should add a stability setting to avoid using version with prerelease like:

I think they should have priority when selecting versions. `stable > prerelease`. But selecting only one stable version is also required, we can use stability constraints`xmake require >2.0.1-stable` or `add_requires("luajit >2.0.1-stable")`. 

You can see [stability constaints of composer](https://getcomposer.org/doc/articles/versions.md#stability-constraints)

> If you are using a constraint that does not explicitly define a stability, Composer will default internally to -dev or -stable, depending on the operator(s) used. This happens transparently.

> If you wish to explicitly consider only the stable release in the comparison, add the suffix -stable.

PR #116, opened 2017-05-24
use SUDO_UID & SUDO_GID to know original uid & gid

TitanSnow:
sudo will set SUDO_UID & SUDO_GID
use this to go back original user


waruqi:
👍

PR #204, opened 2018-08-30
fix targetdir,objectdir not used in vsproject

xigalto:
fix targetdir,objectdir not used in vsproject


waruqi:
请先看下 https://github.com/tboox/xmake/blob/master/CONTRIBUTING.md 里面的贡献说明，请提交pr到dev分支

PR #221, opened 2018-10-09
fix makefile use of null device for windows

rivy:
.# Discussion

On the Windows platform, "NUL" is the null device which discards its input (analogous to "/dev/null" on the *nix platform).


waruqi:
Please commit code to dev branch. You can see https://github.com/tboox/xmake/blob/master/CONTRIBUTING.md

rivy:
Done.

waruqi:
Thanks! :+1:

PR #228, opened 2018-10-14
fix detection strategy for MSVC/VS2017+

rivy:
fix detection strategy for MSVC/VS2017+

* fixes the MSVC/VS 2017 detection problem discussed within <https://github.com/tboox/xmake/issues/225>

### Discussion

MSVC/VS versions >= 15.0 will no longer be setting registry entries for tool location purposes. But `vswhere` (included with MSVC/VS versions >= 15.2, and placed in a guaranteed location) can instead be used to find the needed path.

`os.runv(...)` is used as the tool to gather `vswhere` output instead of `($shell ...)` in order to avoid some `($shell ...)` limitations. Commands containing white space and/or special characters (eg, '[', ')') are difficult or impossible to pass through the current `($shell ...)` implementation, causing application exceptions. Both white space and the noted special characters are needed to construct the required `vswhere` command.

ref: <https://github.com/Microsoft/vswhere/blob/master/README.md> @@ <https://archive.is/mEmdu>

### DIFF
--- a/xmake/modules/detect/sdks/find_vstudio.lua
+++ b/xmake/modules/detect/sdks/find_vstudio.lua
@@ -24,6 +24,7 @@
 
 -- imports
 import("lib.detect.find_file")
+import("detect.tools.find_vswhere")
 
 -- init vc variables
 local vcvars = {"path", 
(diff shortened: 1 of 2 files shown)


waruqi:
Thanks! :+1:

PR #231, opened 2018-10-16
Fixes and minor improvement for detect.sdks.find_vstudio

rivy:
I've included two fixes to find_vstudio for older MSVC/VS builds, detailed in the commit messages.

I also included a commit which modifies the search logic to only use `find_tool("vswhere")` for MSVC/VS versions where it is supported.

### DIFF
--- a/xmake/modules/detect/sdks/find_vstudio.lua
+++ b/xmake/modules/detect/sdks/find_vstudio.lua
@@ -152,13 +152,40 @@ function main()
     ,   ["4.2"]  = "VS42COMNTOOLS"
     }
 
-    -- find vs from environment variables 
+    -- find the single current MSVC/VS from environment variables
     local VCInstallDir = os.getenv("VCInstallDir")
-    local VisualStudioVersion = os.getenv("VisualStudioVersion")
-    if VCInstallDir and VisualStudioVersion then
+    if VCInstallDir and (VCInstallDir ~= "") then
+        local VisualStudioVersion = os.getenv("VisualStudioVersion")
+        if not VisualStudioVersion or (VisualStudioVersion == "") then
 
-        -- find vcvarsall.bat
-        local vcvarsall = path.join(VCInstallDir, "Auxiliary", "Build", "vcvarsall.bat")
+            -- heuristic for VisualStudioVersion value (early MSVC/VS versions don't set VisualStudioVersion)
+            local VSInstallDir = os.getenv("VSInstallDir") or ""
+            VisualStudioVersion = VSInstallDir:match('(%d+[.]?%d*)\\?%s*$')
+            if not VisualStudioVersion then VisualStudioVersion = VCInstallDir:match('(%d+[.]?%d*)\\VC\\?%s*$') end
+            if not VisualStudioVersion then VisualStudioVersion = "0" end
+            if not VisualStudioVersion:match('[.]') then VisualStudioVersion = VisualStudioVersion .. '.0' end
+
+            -- find highest known version which is less than or equal to VisualStudioVersion
+            if not vsvers[VisualStudioVersion] then
+                local versions = {}
+                local count = 0
+                for k in pairs(vsvers) do table.insert(versions, tonumber(k)); count = count + 1 end
+                table.sort(versions)
+                local i = 0
+                local v = tonumber(VisualStudioVersion)
+                while ((i < count) and (versions[i+1] <= v)) do i = i + 1 end
+                VisualStudioVersion = versions[i] or "0"
+            end
+        end
+
+        -- find vcvarsall.bat or vcvars32.bat
+        local pathes =
+        {
+            VCInstallDir.."\\Auxiliary\\Build",
+            VCInstallDir.."\\bin",
+            VCInstallDir
+        }
+        local vcvarsall = find_file("vcvarsall.bat", pathes) or find_file("vcvars32.bat", pathes)
         if os.isfile(vcvarsall) then
 
             -- load vcvarsall
--- a/xmake/modules/detect/sdks/find_vstudio.lua
+++ b/xmake/modules/detect/sdks/find_vstudio.lua
@@ -93,7 +93,7 @@ function _load_vcvarsall(vcvarsall, arch)
     -- @note vcvarsall.bat maybe detect error if install WDK and SDK at same time (multi-sdk version exists in include directory).
     --
     local UCRTVersion = variables["UCRTVersion"]
-    if UCRTVersion and UCRTVersion ~= WindowsSDKVersion and WindowsSDKVersion ~= "" then
+    if UCRTVersion and WindowsSDKVersion and UCRTVersion ~= WindowsSDKVersion and WindowsSDKVersion ~= "" then
         local lib = variables["lib"]
         if lib then
             lib = lib:gsub(UCRTVersion, WindowsSDKVersion)


xmake/modules/detect/sdks/find_vstudio.lua
```diff
@@ -210,11 +207,14 @@ function main()
...
-            local vswhere_vrange = format("%s,%s)", version, (version+1))
-            local result = os.iorunv(vswhere.program, {"-property", "installationpath", "-products", "Microsoft.VisualStudio.Product.BuildTools", "-version", vswhere_vrange})
-            if result then 
-                vswhere_VCAuxiliaryBuildDir = result:trim() .."\\VC\\Auxiliary\\Build" 
+        if ((version + 0) >= 15) then
+            local vswhere = find_tool("vswhere")
+            if vswhere then
+                local vswhere_vrange = format("%s,%s)", version, (version+1))
+                local result = os.iorunv(vswhere.program, {"-property", "installationpath", "-products", "Microsoft.VisualStudio.Product.BuildTools", "-version", vswhere_vrange})
+                if result then
```
waruqi: But this condition `if ((version + 0) >= 15) then` will always be satisfied. 

`find_tool("vswhere")` will always be executed once or even multiple times if the version list contains 16.0 17.0 .. in future.

xmake/modules/detect/sdks/find_vstudio.lua
```diff
@@ -152,13 +152,40 @@ function main()
...
-    local VisualStudioVersion = os.getenv("VisualStudioVersion")
-    if VCInstallDir and VisualStudioVersion then
+    if VCInstallDir and (VCInstallDir ~= "") then
+        local VisualStudioVersion = os.getenv("VisualStudioVersion")
+        if not VisualStudioVersion or (VisualStudioVersion == "") then
+
+            -- heuristic for VisualStudioVersion value (early MSVC/VS versions don't set VisualStudioVersion)
+            local VSInstallDir = os.getenv("VSInstallDir") or ""
+            VisualStudioVersion = string.match(VSInstallDir, '(%d+[.]?%d*)\\?%s*$')
+            if not VisualStudioVersion then VisualStudioVersion = string.match(VCInstallDir, '(%d+[.]?%d*)\\VC\\?%s*$') end
```
waruqi: You can write `VSInstallDir:match'(%d+[.]?%d*)\\?%s*$')` to simplify code.

xmake/modules/detect/sdks/find_vstudio.lua
rivy: The condition should only be satisfied if the MSVC/VS version being tested for is greater than or equal to "15.0". If that's not correct, I need to rework it so that it is... the entire "vswhere" logic only applies to MSVC/VS versions at/after version 15.0/2017.

rivy: Ah, I see your point. It's only being used for the "15.0" version now, but in the future, the search will be run multiple times because it's not being cached. And it'll always run at least once because the search always includes "15.0".

Logical failure on my part.

I remove that commit and re-push it.

waruqi:
```lua
-- always only execute once
local vswhere = find_tool("vswhere") 
for _, version in ipairs({"17.0", "16.0", "15.0", "14.0", "12.0", "11.0", "10.0", "9.0", "8.0", "7.1", "7.0", "6.0", "5.0", "4.2"}) do
    -- ...
end
```

```lua
for _, version in ipairs({"17.0", "16.0", "15.0", "14.0", "12.0", "11.0", "10.0", "9.0", "8.0", "7.1", "7.0", "6.0", "5.0", "4.2"}) do
    -- will be executed three times
    if ((version + 0) >= 15) then
        local vswhere = find_tool("vswhere") 
    end
end
```

```lua
for _, version in ipairs({"15.0", "14.0", "12.0", "11.0", "10.0", "9.0", "8.0", "7.1", "7.0", "6.0", "5.0", "4.2"}) do
    -- always only execute once
    if ((version + 0) >= 15) then
        local vswhere = find_tool("vswhere") 
    end
end
```

waruqi: ok

xmake/modules/detect/sdks/find_vstudio.lua
rivy: I didn't know that... (I'm new to Lua). I'll make the changes here and re-push it.

waruqi:
The purpose of find_vstudio is to detect all vs environments on the system, not just the current one.
Because under the vs command prompt we also need to switch the vs version.

for example:

```console
$xmake f --vs=2015
$xmake
$xmake project -k vs2017
$xmake project -k vs2015
```

```lua
-- find the single current MSVC/VS from environment variables
local VCInstallDir = os.getenv("VCInstallDir")
if VCInstallDir and (VCInstallDir ~= "") then
           
        -- save results
        local results = {}
        results[vsvers[VisualStudioVersion]] = {version = VisualStudioVersion, vcvarsall_bat = vcvarsall, vcvarsall = {x86 = vcvarsall_x86, x64 = vcvarsall_x64}}
        return results 
    end
end
```

xmake/modules/detect/sdks/find_vstudio.lua
rivy: Updated.

waruqi:
you can use os.getenv(VisualStudioVersion) get the current vs version and save to results.

Then we can select this version first in

https://github.com/tboox/xmake/blob/master/xmake/platforms/windows/check.lua

rivy:
I'm not sure it's possible to detect the *correct* environments of all MSVC/VS environments on the system when you're within an already pre-prepared environment. I think that to do so would require scrubbing out the currently prepared versions environment variables. You could delete some, but scrubbing the PATH would be quite difficult.

Prior to this commit set, the results of find_vstudio were a muddled combination of the environments between the currently set environment and the next detected MSVC/VS version. And, I definitely have had problems when inadvertantly running multiple `vsvarsall` from differing MSVC/VS versions in the past.

And the original source returned after detecting just the currently set environment if it saw VCInstallDir and VisualStudioVersion.

waruqi:
We should also check other versions while acquiring the current VS environment.
Not just one version

waruqi:
> you'll need to, at least, delete all the environment variables except PATH. 

Why? We only need to detect one VS version in the current environment variable of command prompt and detect other versions from vswhere and registry

rivy:
> os.getenv(VisualStudioVersion) to get the current vs version and select this version first

I like the idea of selecting the currently setup version first, but, I believe that'll only work for versions of MSVC/VS which set that environment variable. I'm not sure when the transition happened, but MSVC/VS 8.0/2005 (and earlier) doesn't have it. I can see that MSVC/VS v11.0/2012 does have it. So it could have been added anywhere in the VC9/2008, VC10/2010, or VC11/2012 versions.

rivy:
> Why? We only need to detect one VS version in the current environment variable of command prompt and detect other versions from vswhere and registry

The results that I see from find_vstudio have all of the environment variables associated with that version listed (with values) in the return results table. But, with a pre-prepared environment, the environment variables from the newly discovered environment are mixed on top of the current settings. I think using the resulting mixed environment will cause later issues. It's definitely much different from the results returned when you have a non-prepared (normal CMD) environment.

It's possible that cleaning the listed environment variables would clean things up enough to avoid issues. Although I'm still worried that the value of PATH may be problematic.

waruqi:
The detection process does not affect the current environment

```lua
local file = io.open(genvcvars_bat, "w")
    file:print("@echo off")
    file:print("call \"%s\" %s > nul", vcvarsall, arch)
    for idx, var in ipairs(vcvars) do
        file:print("echo " .. var .. " = %%" .. var .. "%% %s %s", idx == 1 and ">" or ">>", genvcvars_dat)
    end
    file:close()
```

waruqi:
I will merge this patch first, then I will modify it slightly.

rivy:
Yes, I agree, the detection doesn't effect the current environment, but the current environment effects what results returns as values for environment variables. Using your current "dev" branch code, I've attached three files ("CMD.txt" which is the `xmake lua detect.sdks.find_vstudio` output from a normal CMD prompt without running `vsvarsall`, "CMD+VC8.txt" is run from the VC8 command prompt, and "CMD+2017.txt" is run from the VC2017 command prompt).

[CMD.txt](https://github.com/tboox/xmake/files/2481556/CMD.txt)
[CMD+VC8.txt](https://github.com/tboox/xmake/files/2481557/CMD%2BVC8.txt)
[CMD+VC2017.txt](https://github.com/tboox/xmake/files/2481558/CMD%2BVC2017.txt)

Note the variation in what is returned and how the values of VC2017 are contaminated by VC8.

waruqi:
Ok, I understand what you mean and agree with you, so let's do it first.

rivy:
So, after thinking about this, one way to get all versions even if one is already setup via a "VCX Command Prompt" or running `vcvarsall` would be to reset the environment prior to searching for the other versions:

1. save the current environment, then erase it
2. re-read and reconstruct the environment from the registry via the "HKLM\System\CurrentControlSet\Control\Session Manager\Environment" and "HKCU\Environment" keys (see <https://stackoverflow.com/a/32420542/43774>); (note: merging the PATH variables)
3. if the VCVARS_ARE_SET and/or VCInstallDir environment variables are missing (meaning the user environment has a "clean" version without MSVC/VS variables), then search for other versions, skipping any already defined version in the results array

As long as the MSVC/VS variables are not "permanently" set in the user environment, this would give you all available versions with the cleanest respective environment.

Along with this, I would default to use the current environment MSVC/VS version if it's already setup.

I can see how it needs to be done, but I don't know how to read all of the registry keys to find the variables. All of the examples I see accessing the registry are for specific values within a key. If you could guide me on that, I could work up a prototype of the above algorithm.

waruqi:
Thank you very much for taking the time to find a solution to this problem, I thought about it and I think the current processing is very good, we don't have to do too complicated processing on the command prompt environment.

If needs to switch the vs environment to compile, we can open a clean cmd.exe to do it.
And if a vs commad prompt has been opened, in most cases, we just want to compile in the current vs environment.

PR #235, opened 2018-10-18
improve find_vswhere (allow use of check/command options)

rivy:
* minor change
* allows use of check or command options to be passed into `find_vswhere`, similar to all the other `find_...` routines


rivy:
I also saw that find_dmd has find_vswhere notations in it now. I was going to correct that back to the original code. But I'm not sure what `dmd` is and thought any correction that I did might cause problems.

waruqi:
> I also saw that find_dmd has find_vswhere notations in it now. I was going to correct that back to the original code. But I'm not sure what dmd is and thought any correction that I did might cause problems.

This is indeed my mistake. I have reverted this file.

Thanks!

PR #247, opened 2018-11-02
fix find_program ~ don't report errors for failed simple checks

rivy:
When searching for tools, errors were emitted from the "check" function when a tool was missing from a searched path. That is expected behavior and I would say should have only an informational note or nothing at all.

This PR moves any check with a non-function `opt.check` together with the check done with a missing `opt.check`. That check never emits an error, so this change suppresses the, IMO unnecessary, error noted above.

### DIFF
--- a/xmake/core/sandbox/modules/import/lib/detect/find_program.lua
+++ b/xmake/core/sandbox/modules/import/lib/detect/find_program.lua
@@ -54,19 +54,15 @@ function sandbox_lib_detect_find_program._check(program, opt)
         return os.isfile(program) 
     end
 
-    -- no check script? attempt to run it directly
-    if not opt.check then
-        return 0 == os.execv(program, {"--version"}, {stdout = os.nuldev(), stderr = os.nuldev()})
+    -- no check function? attempt to run it directly
+    if not opt.check or type(opt.check) ~= "function" then
+        return 0 == os.execv(program, {opt.check or "--version"}, {stdout = os.nuldev(), stderr = os.nuldev()})
     end
 
     -- check it
     local ok = false
     local errors = nil
-    if type(opt.check) == "string" then
-        ok, errors = os.runv(program, {opt.check})
-    else
-        ok, errors = sandbox.load(opt.check, program) 
-    end
+    ok, errors = sandbox.load(opt.check, program)
 
     -- check failed? print verbose error info
     if not ok then


waruqi:
This is unnecessary, I need to add `-v` argument to run `xmake -v` for knowing why the detection tool failed.

And if you do not add the -v parameter, this error message is already suppressed.

rivy:
Ok, I understand that you might want to see information of the detection tool failed. But to me an "error" denotes something that is configured incorrectly and needs to be fixed. It's concerning when I see an "error" or "warning" during builds, even if only during a verbose (`-v`) build. And I think lots of other devs would see the "error" and spend time searching for a repair as well.

From my reading of the source code, the check function can be run several times while searching through paths for a tool and will, during normal operation, "fail" for earlier paths which don't contain the tool until finding it in a later path (driven by the find_program() function). I initially thought about adding a "warning" and/or "info" function, eg...

```lua
-- the verbose warning function
function utils.vwarning(format, ...)

    -- enable verbose?
    if option.get("verbose") and format ~= nil then
        utils.warning(format, ...)
    end
end

-- the verbose information function
function utils.vinfo(format, ...)

    -- enable verbose?
    if option.get("verbose") and format ~= nil then
        -- ... bright blue info message ...
    end
end

```

... and then changing check to issue a `vwarning` or `vinfo` instead of `verror`. But I thought it better to move the simple check up into the same context. when I saw that the initial exec can fail without issuing any error:

https://github.com/tboox/xmake/blob/d130b8049af3250f3504580d146e3c54ed052a5f/xmake/core/sandbox/modules/import/lib/detect/find_program.lua#L57-L59

It seems to me that given what you want (and I agree that knowing which paths failed for a tool search is useful), check should issue an informational note if it doesn't find the tool on a specific path (including in the above notes section of code), and maybe find_program() or some higher level function should issue an error if the tool is truly, finally, not found in all searched locations.

waruqi:
Ok, I understand and agree with you, as the error message is really not friendly during a verbose (-v) build. But I haven’t thought about how to improve the display detection failure information. Maybe `utils.vinfo`, add `xmake -v --more` or `xmake -V` arguments  to get more verbose info only for developers, and etc.

I need to take some time to think about it.

waruqi:
I add `-D` and `--diagnosis` arguments to get more diagnosis information only for developers.
 d3d05b0c620dbfc318d4c693bc2fb7b556d51d3c

```console
$ xmake -h
  -v, --verbose      Print lots of verbose information for users.
  -D, --diagnosis    Print lots of diagnosis information (backtrace, check info ..) only for developers.
                     And we can append -v to get more whole information.
                           e.g. $ xmake -v -D
```

And I have improved the check errors output:

```lua
    if not ok and option.get("diagnosis") then
        utils.cprint("${yellow}checkinfo: ${clear dim}" .. errors)
    end
```

PR #248, opened 2018-11-02
fix + minor improvements to `find_vstudio`

rivy:
When not specified, the most recent version of MSVC/VS is chosen. The choice is made based upon a sort of the versions release year. However, some earlier versions had version number instead of release year which caused some sorting irregularities when those earlier versions were present together with later versions.
All versions now have a specified year, fixing the sorting issue.

Years of release are based upon a [wikipedia source](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B) archived at <http://archive.is/OWqM3>.
And, as it is not referenced, the year of release for MSVC/VS 4.2 is estimated.

... plus, minor code improvements (in a separate commit)


xmake/modules/detect/sdks/find_vstudio.lua
```diff
@@ -136,10 +136,10 @@ function main(opt)
...
     ,   ["7.1"]  = "2003"
-    ,   ["7.0"]  = "7.0"
-    ,   ["6.0"]  = "6.0"
-    ,   ["5.0"]  = "5.0"
-    ,   ["4.2"]  = "4.2"
+    ,   ["7.0"]  = "2002"
+    ,   ["6.0"]  = "1998"
+    ,   ["5.0"]  = "1997"
+    ,   ["4.2"]  = "1996"
     }
```
waruqi: This will affect the following detection code.

```lua
local pathes = 
{
    format("$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\VisualStudio\\SxS\\VS7;%s)\\VC", version),
    format("$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\VisualStudio\\SxS\\VS7;%s)\\VC7\\bin", version),
    format("$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\VisualStudio\\SxS\\VS7;%s)\\VC", version),
    format("$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\VisualStudio\\SxS\\VS7;%s)\\VC\\Auxiliary\\Build", version),
    format("$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\VisualStudio\\SxS\\VS7;%s)\\VC\\Auxiliary\\Build", version)
}
if vsenvs[version] then
    table.insert(pathes, format("$(env %s)\\..\\..\\VC", vsenvs[version]))
end
if vswhere_VCAuxiliaryBuildDir and os.isdir(vswhere_VCAuxiliaryBuildDir) then
    table.insert(pathes, vswhere_VCAuxiliaryBuildDir)
end
```

For the lower version of vs, we need to read the registry and environment variables from '7.0', '5.0' instead of '2002', '1997', ...

rivy: Thanks, yes, I made a mistake in not following the usage through the function.

So, the alternative is to sort the versions numerically (instead of as strings) when choosing the preferred version.

This change seems to fix the problem as well.

https://github.com/rivy/xmake/commit/188b2bda739b339188111b331eb5b919f91a8044

If that looks good to you, I'll squash the commit to this change, removing the erroneous year changes, and update the PR.

PR #249, opened 2018-11-02
revise progress formatting to space-leading three digit percentages

rivy:
Formatting was a bit off for progress when hitting 100%. I thought the leading zeros were also distracting. I can put those back in if you prefer them. It's just not what I'm used to seeing.

### before

```shell
C:>rm -r build .xmake & xmake
checking for the architecture ... x64
checking for the Microsoft Visual Studio (x64) version ... 2015
building to: build\windows.release.cl.x64\bin\hello.exe
[00%]: compiling.release src\main.rc
[50%]: compiling.release src\main.c
[100%]: linking.release hello.exe
```

### after

```shell
C:>rm -r build .xmake & xmake
checking for the architecture ... x64
checking for the Microsoft Visual Studio (x64) version ... 2015
building to: build\windows.release.cl.x64\bin\hello.exe
[  0%]: compiling.release src\main.rc
[ 50%]: compiling.release src\main.c
[100%]: linking.release hello.exe
```

There was no single location for the change, so I changed it throughout the code base. I thought about localizing it to a single code location, but I wasn't sure where it would be best placed. Future localization would probably be a good idea though in order to plan for alternative formatting (different color choices, etc).

### DIFF
--- a/xmake/actions/build/kinds/binary.lua
+++ b/xmake/actions/build/kinds/binary.lua
@@ -77,7 +77,7 @@ function _build_from_objects(target, buildinfo)
     local verbose = option.get("verbose")
 
     -- trace progress into
-    cprintf("${green}[%02d%%]:${clear} ", (buildinfo.targetindex + 1) * 100 / buildinfo.targetcount)
+    cprintf("${green}[%3d%%]:${clear} ", (buildinfo.targetindex + 1) * 100 / buildinfo.targetcount)
     if verbose then
         cprint("${dim magenta}linking.$(mode) %s", path.filename(targetfile))
     else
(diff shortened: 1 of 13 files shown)


waruqi:
Ok, thanks and I will move to a single code location in these days.

waruqi:
> There was no single location for the change, so I changed it throughout the code base. I thought about localizing it to a single code location, but I wasn't sure where it would be best placed. Future localization would probably be a good idea though in order to plan for alternative formatting (different color choices, etc).

I have move it to the single location. You can see #310 issue about theme.

```lua
    set_text("build.progress_format", "[%3d%%]")
    set_color("build.progress", "green")
```

PR #282, opened 2018-12-10
Enable nvcc to work with -Xcompiler option.

wangii:
ref https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-passing-specific-phase-options

### DIFF
--- a/xmake/modules/detect/tools/nvcc/has_flags.lua
+++ b/xmake/modules/detect/tools/nvcc/has_flags.lua
@@ -31,7 +31,7 @@ function _islinker(flags, opt)
   
     -- the flags is "-Wl,<arg>" or "-Xlinker <arg>"?
     local flags_str = table.concat(flags, " ")
-    if flags_str:startswith("-Wl,") or flags_str:startswith("-Xlinker ") then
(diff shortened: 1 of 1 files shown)


xmake/modules/detect/tools/nvcc/has_flags.lua
```diff
@@ -31,7 +31,7 @@ function _islinker(flags, opt)
   
     -- the flags is "-Wl,<arg>" or "-Xlinker <arg>"?
     local flags_str = table.concat(flags, " ")
-    if flags_str:startswith("-Wl,") or flags_str:startswith("-Xlinker ") then
+    if flags_str:startswith("-Wl,") or flags_str:startswith("-Xlinker ") flags_str:startswith("-Xcompiler ") then
         return true
```
waruqi: Here is the invalid grammar, we need add `or` keyword, e.g. `or flags_str:startswith("-Xcompiler ")`.

`-Xcompiler xx` is not linker argument options, it is only for compiler. We should not judge it in `function _islinker(flags, opt)`.

waruqi:
And please commit code to dev branch and we will merge into master branch in future. You can see [CONTRIBUTING](https://github.com/tboox/xmake/blob/master/CONTRIBUTING.md).

Thanks!

wangii:
Got it, thanks!

PR #387, opened 2019-04-10
improve vcpkg support for linux.

ivechan:
### TLDR: 
When using vcpkg, machine architecture should change  **x86_64**  to **x64** not only on **macos** but also on **linux** .

### Detail:
Run `./vcpkg help triplet`
it show that the valid options of target machine architecture are as follows:
```
Available architecture triplets:
  arm-uwp
  arm-windows
  arm64-uwp
  arm64-windows
  x64-linux
  x64-osx
  x64-uwp
  x64-windows
  x64-windows-static
  x86-uwp
  x86-windows
  x86-windows-static
```
also refer to https://vcpkg.readthedocs.io/en/latest/users/triplets/#VCPKG_TARGET_ARCHITECTURE.

### DIFF
--- a/xmake/modules/package/manager/vcpkg/find_package.lua
+++ b/xmake/modules/package/manager/vcpkg/find_package.lua
@@ -48,11 +48,8 @@ function main(name, opt)
     local mode = opt.mode 
     if plat == "macosx" then
         plat = "osx"
-        if arch == "x86_64" then
-            arch = "x64"
-        end
     end
-    if plat == "linux" and arch == "x86_64" then
+    if arch == "x86_64" then
         arch = "x64"
     end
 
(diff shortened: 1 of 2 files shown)


xmake/modules/package/manager/vcpkg/find_package.lua
```diff
@@ -52,6 +52,9 @@ function main(name, opt)
             arch = "x64"
         end
     end
+    if plat == "linux" and arch == "x86_64" then
+        arch = "x64"
+    end
```
waruqi: I think we can simplify them if all platform architectures are not x86_64

```lua
    if plat == "macosx" then
        plat = "osx"
    end
    if arch == "x86_64" then
        arch = "x64"
    end
```

waruqi:
Thanks!

xmake/modules/package/manager/vcpkg/find_package.lua
ivechan: agree.

PR #392, opened 2019-04-11
makefile: fix relative paths; map amd64 to x86_64

0x17de:
In some build environments the PWD for recursive builds will point to the wrong directory causing a wrong include dir for the luajit arch specific headers. I therefore added $(CURDIR) to the relative path inside the luajit makefile.

Since luajit is missing a directory for amd64, this arch is also mapped to x86_64 (see mapping table arch.mak). This one is now used at two different locations and therefore i made to a separate file inside the core directory.

### DIFF
--- a/xmake/modules/detect/sdks/find_qt.lua
+++ b/xmake/modules/detect/sdks/find_qt.lua
@@ -53,6 +53,7 @@ function _find_sdkdir(sdkdir, sdkver)
         table.insert(subdirs, path.join(sdkver or "*", "android_*", "bin"))
     end
     table.insert(subdirs, path.join(sdkver or "*", "*", "bin"))
+    table.insert(subdirs, path.join("*", "bin"))
 
     -- init the search directories
     local pathes = {}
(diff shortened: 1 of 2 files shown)


core/src/luajit/makefile
```diff
@@ -1,5 +1,7 @@
 # prefix
 include $(PRO_DIR)/prefix.mak
+# architecture mappings
+include $(PRO_DIR)/arch.mak
```
waruqi: We need not include arch.mak file here, please move to core/makefile

https://github.com/xmake-io/xmake/blob/f4781b3cbc4d8e61eddb888ea4e7d258adc70c5b/core/makefile#L141-L176

0x17de:
I will provide a smaller update later today.
I finally have spotted my error.  This build environment (gentoo, portage via build script) behaved weird, but the actual reason was the arch being set to amd64 _only_ inside the build environment instead of x86_64 compared to when i called make manually.

PR #394, opened 2019-04-12
rename ARCH amd64 to x86_64: allows luajit to compile

0x17de:
Luajit has a directory for x86_64 but not for amd64 - to allow compilation the architecture variable is mapped.

### DIFF
--- a/makefile
+++ b/makefile
@@ -39,6 +39,9 @@ endif
 
 endif
 
+# map architecture amd64 to x86_64
+ARCH 		:=$(if $(findstring amd64,$(ARCH)),x86_64,$(ARCH))
+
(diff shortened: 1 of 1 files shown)


waruqi:
I didn't see where the `ARCH` variable was set to amd64. ?

0x17de:
The arch variable was set by the build environment (portage) - not by your script

waruqi:
Ok, I hope to explain this situation in the comments.

waruqi:
Because ARCH=arm64 is not explicitly set in the makefile, I just want to add a comment to explain the situation on your side about ARCH=amd64.

In fact, there is nothing wrong with it, amd64 usually has to deal with it.

waruqi:
I merged first here.

PR #402, opened 2019-04-16
makefile: describe mapping from amd64 to x86_64

0x17de:
Here is some description for the architecture mapping inside the make file - if the ARCH variable is set from the outside e.g. from the build environment provided by the OS.


waruqi:
Ok, thanks!

PR #551, opened 2019-08-26
fix error caused by the new vsDevCmd.bat of vs2019

ChanthMiao:
A specific block of the new vsDevCmd.bat may cause system error. Use env var 'VSCMD_SKIP_SENDTELEMETRY' to skip it.
Solution referrer: <https://developercommunity.visualstudio.com/content/problem/694847/vsdevcmdbat-developer-prompt-causes-the-input-line.html>


waruqi:
https://github.com/xmake-io/xmake/issues/549

PR #679, opened 2020-02-07
Speedup non-MSVC builds

rivy:
This PR bypasses the (relatively slow) `check_vstudio` routine for non-`cl` builds (eg, `clang` or `gcc` builds). The additional `cl` executables beyond `cl.exe` (`cl.com`, `cl.bat`, `cl.cmd`) may be a bit overkill.

I also normalized the color usage (success/ failure) for the "check_*" tasks.

### DIFF
--- a/xmake/modules/private/platform/check_arch.lua
+++ b/xmake/modules/private/platform/check_arch.lua
@@ -32,7 +32,6 @@ function main(config, default)
         config.set("arch", default or os.arch())
 
         -- trace
-        cprint("checking for the architecture ... ${green}%s", config.get("arch"))
+        cprint("checking for the architecture ... ${color.success}%s", config.get("arch"))
     end
 end
-
--- a/xmake/modules/private/platform/check_vstudio.lua
+++ b/xmake/modules/private/platform/check_vstudio.lua
@@ -87,15 +87,13 @@ function main(config)
         config.set("vs", vs, {readonly = true, force = true})
 
         -- trace
-        print("checking for the Microsoft Visual Studio (%s) version ... %s", config.get("arch"), vs)
+        cprint("checking for the Microsoft Visual Studio (%s) version ... ${color.success}%s", config.get("arch"), vs)
     else
         -- failed
-        print("checking for the Microsoft Visual Studio (%s) version ... no", config.get("arch"))
-        print("please run:")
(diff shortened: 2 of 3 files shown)


waruqi:
`xmake f -p windows` only for msvc builds. Please use cross-complation or mingw. 

e.g. 

```console
xmake f -p cross --cc=clang --ld=clang --ar=...
```

see https://xmake.io/#/guide/configuration?id=cross-compilation

rivy:
Thanks for the response.

That does compile for `CC=clang` and `CC=gcc`, but the clang version results in a final file without the trailing ".exe" necessary to run the file on Windows. `CC=gcc` works correctly (at least for 32-bits; I've not tried 64-bits).

The platform == "cross" seems a bit odd though. For `CC=clang` ... its a windows hosted compiler, compiling a windows executable for a windows target, not what I would consider a cross-compiler.

rivy:
Strangely, for `CC=gcc`, although a final target is generated with correct executable suffix (".exe"), the `target:targetfile()` value is missing that suffix.

waruqi:
> And, strangely, for `CC=gcc`, although a final target is generated with correct executable suffix (".exe"), the `target:targetfile()` value is missing that suffix.

you can set `set_filename("xxx.exe")` for target.

> The platform == "cross" seems a bit odd though. 

You can pass other custom platform name as cross-compilation. 

```console
xmake f -p win_clang --cc=clang.exe --ld=clang.exe ...
```

xmake.lua
```lua
if is_plat("win_clang") then
    ...
end
```

rivy:
Hmm, that seems like a lot of trouble, on an ongoing basis, for other projects.

Would it be possible to just copy the `windows` platform, removing the `check_vstudio` call and tweaking the other parameters, incorporating it into the mainline as say `windows_gcc`. and/or `windows_clang`?

waruqi:
I have merged this patch. https://github.com/xmake-io/xmake/commit/96757ac7da97b7205b91ca396eec1408b4e16a59

rivy:
Thanks!
I appreciate your work and thoughts to help simplify the alternate compiler windows use-cases.

waruqi:
we can use the following command to switch to clang on windows fastly now.

```console
xmake f --toolchain=clang
xmake
```

you can see https://github.com/xmake-io/xmake/issues/780

we need update to dev version first if try it.

xmake update dev

PR #682, opened 2020-02-09
Refactor ~ enable `rc` search when needed

rivy:
`clang` may require the use of MSVC `rc`.
This modification enable the needed `check_vstudio()` call when `rc` is requested.


rivy:
On windows, for the default "windows" platform, this patch does require using `--mrc=` or `--mrc=<ALTERNATE_RC>` to avoid the check_vstudio() call. I'm not sure if there's another way, but this markedly speeds up `clang` and `gcc` configurations when either no or alternate resource compilers are used.

PR #686, opened 2020-02-23
change ~ use 'success' theme more widely

rivy:
* replaces "${bright}...ok!" with "${color.success}...ok!${clear}" widely throughout the code base

### DIFF
--- a/xmake/actions/config/scangen.lua
+++ b/xmake/actions/config/scangen.lua
@@ -156,5 +156,5 @@ function main()
     end
 
     -- trace
-    cprint("${color.success}xmake.lua generated, scan ok!${clear}")
+    cprint("${color.success}xmake.lua generated, scan ok!")
 end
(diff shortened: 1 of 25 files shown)


xmake/actions/config/scangen.lua
```diff
@@ -156,5 +156,5 @@ function main()
     end
 
     -- trace
-    cprint("${bright}xmake.lua generated, scan ok!${clear}")
+    cprint("${color.success}xmake.lua generated, scan ok!${clear}")
```
waruqi: we need not add `${clear}` at tail, because cprint will automatically add it internally. 

https://github.com/xmake-io/xmake/blob/4a7f397b9036a3c31cad090732d758794555e40c/xmake/core/base/colors.lua#L326

we need only add `${color.success}`.

```lua
    cprint("${color.success}xmake.lua generated, scan ok!")
```

rivy: Sure, I'll look through the code base now.

rivy: Done ... added to the PR.

I removed all the `${clear}` elements which were the last portion of a `cprint(...)`.

waruqi:
thanks

PR #954, opened 2020-09-16
make find_package success log more explicit

paul-reilly:
I think something like this would have made the issue of using a vcpkg instead of an xmake package I encountered in https://github.com/xmake-io/xmake-repo/issues/68#issuecomment-693230739 a lot more obvious to the untrained eye.

**New log message:**
```
checking for xmake::luajit ... using xmake::luajit 2.1.0-beta3
configure
{
...
```
**Old log message:**
```
checking for xmake::luajit ... 2.1.0-beta3
configure
{
...
```


paul-reilly:
I also changed the incorrect use of the word "the" in front of what are proper nouns.

waruqi:
> checking for xmake::luajit ... using xmake::luajit 2.1.0-beta3

find_package is just to find the library, it may not be using the library. Although add_requires internally will use find_package() to find the library.

So I don’t think we should add `using`.

```console
xmake l find_package luajit
xmake l find_package xmake::luajit
xmake l find_package vcpkg::luajit
xmake l find_package system::luajit
```

In addition, the first `checking for xmake::xxx` has told us that it is looking for it from the `xmake::` package

Therefore, we only need the following to let us know which package is currently being checked.

```
checking for xmake::luajit ... luajit 2.1.0-beta3
                          |
             from xmake-repo
```

waruqi:
I think it is only necessary to add the source of the package to the result when the package namespace is not specified

for example:

```console
checking for luajit ... vcpkg::luajit 2.0.5-3
```

If the package source (`xmake::`) has been specified when checking, there is no need to add it.

```console
checking for xmake::luajit ... 2.0.5-3
```

waruqi:
We need get package namespace from `result, manager = find_package("...")` https://github.com/xmake-io/xmake/blob/master/xmake/modules/package/manager/find_package.lua

paul-reilly:
Yes, that's a good idea and good point about using "using", I'll update the PR in a bit. I think in place of "using" that "found" would be reassuringly explicit too, I'll put it in and we can remove it later if you like.

xmake/modules/lib/detect/find_package.lua
```diff
@@ -23,6 +23,7 @@ import("core.base.option")
 import("core.project.config")
 import("lib.detect.cache")
 import("package.manager.find_package")
+import("utils.split_package_name")
```
waruqi, line 26: Please move to `private.utils.split_package_name`

xmake/modules/package/manager/find_package.lua
```diff
@@ -176,5 +168,5 @@ function main(name, opt)
     end
 
     -- ok?
-    return result
+    return result, manager_name, package_name
```
waruqi, line 171: We should return the package manager that was actually detected, because the passed name parameter may not contain `xxx::`.

https://github.com/xmake-io/xmake/blob/d79274d2e2d57a873e69463848f0ba4e1074dd7c/xmake/modules/package/manager/find_package.lua#L67-L76

xmake/modules/utils/split_package_name.lua
```diff
@@ -0,0 +1,52 @@
+--!A cross-platform build utility based on Lua
```
waruqi, line 1: Please move to private.utils.split_package_name

xmake/modules/lib/detect/find_package.lua
```diff
@@ -83,9 +85,15 @@ function main(name, opt)
...
     if opt.verbose or option.get("verbose") then
         if result then
-            cprint("checking for the %s ... ${color.success}%s", name, result.version and result.version or "${text.success}")
+            -- only display manager of found package if the package we searched for
+            -- did not specify a package manager
+            local original_manager_name, _, _ = split_package_name(name)
+            local display_name = original_manager_name and package_name
+                or (manager_name .. "::" .. package_name)
+
+            cprint("checking for %s ... ${color.success}found %s %s", name, display_name, result.version and result.version or "")
```
waruqi, line 94: I think there is no need to add `found` word.

xmake/modules/lib/detect/find_package.lua
```diff
@@ -83,9 +85,15 @@ function main(name, opt)
     -- trace
     if opt.verbose or option.get("verbose") then
         if result then
-            cprint("checking for the %s ... ${color.success}%s", name, result.version and result.version or "${text.success}")
+            -- only display manager of found package if the package we searched for
+            -- did not specify a package manager
+            local original_manager_name, _, _ = split_package_name(name)
+            local display_name = original_manager_name and package_name
+                or (manager_name .. "::" .. package_name)
```
waruqi, line 92: we need only the returned real manager name and package name. So we need not call split_package_name()

We need to ensure that the returned manager_name must be a non-empty string.

```lua
local display_name = (manager_name or "") .. "::" .. package_name
```

xmake l find_package luajit          -- manager_name is vcpkg/xmake/xx  or system (default)
xmake l find_package vcpkg::luajit  -- manager_name vcpkg

waruqi: Therefore, we don’t need to put split_package_name separately in the independent module file, just use it in package/manager/find_package.lua

xmake/modules/lib/detect/find_package.lua
paul-reilly: No problem, removed.

xmake/modules/package/manager/find_package.lua
paul-reilly: Good point, thanks. Getting it from there.

xmake/modules/lib/detect/find_package.lua
paul-reilly: Oh yeah, sorry about no nil guard there.

Re the second point, to satisfy https://github.com/xmake-io/xmake/pull/954#issuecomment-693410906 I believe we still need to check if the requested manager is the same as the found manager. We can return the requested manager from `find_package` as an additional return value though, so there's no code duplication or extra file.

xmake/modules/lib/detect/find_package.lua
```diff
@@ -85,13 +84,13 @@ function main(name, opt)
...
+
             -- only display manager of found package if the package we searched for
             -- did not specify a package manager
-            local original_manager_name, _, _ = split_package_name(name)
-            local display_name = original_manager_name and package_name
-                or (manager_name .. "::" .. package_name)
-
-            cprint("checking for %s ... ${color.success}found %s %s", name, display_name, result.version and result.version or "")
+            found_manager_name = found_manager_name or ""
+            local display_manager = orig_manager_name == found_manager_name and "" or found_manager_name .. "::"
```
waruqi, line 91: We need not orig_manager_name.

```lua
local display_manager = name:find("::", 1, true) and "" or (found_manager_name .. "::")
```

xmake/modules/package/manager/find_package.lua
```diff
@@ -65,15 +64,16 @@ function _find_package_with_builtin_rule(package_name, opt)
...
     -- find package from the given package manager
-    local result = nil
+    local result, manager_name = nil, nil
     for _, manager_name in ipairs(managers) do
         dprint("finding %s from %s ..", package_name, manager_name)
         result = import("package.manager." .. manager_name .. ".find_package", {anonymous = true})(package_name, opt)
         if result then
             break
         end
+        manager_name = nil
```
waruqi, line 74:
```lua
    local result = nil
    local manager_found = nil
    for _, manager_name in ipairs(managers) do
        dprint("finding %s from %s ..", package_name, manager_name)
        result = import("package.manager." .. manager_name .. ".find_package", {anonymous = true})(package_name, opt)
        if result then
            manager_found = manager_name
            break
        end
    end
    return result, manager_found
```

xmake/modules/lib/detect/find_package.lua
paul-reilly: Ok, changed.

xmake/modules/package/manager/find_package.lua
paul-reilly: Sorry about the variable shadowing on the for loop variable. Changed to your working/cleaner version!

waruqi:
Thanks!

paul-reilly:
You're welcome!

PR #1021, opened 2020-11-09
toolchain: remove includedirs and linkdirs of system path

jonathanpoelen:
`/usr/{local/,}include` and `/usr/{local/,}lib` are in the default configuration of the compilers and forcing them to be added (before any other includes) conflicts with the project files.


waruqi:
Thanks!

PR #1022, opened 2020-11-09
prefer to use c++ compiler for c++ files (gcc/clang -> g++/clang++)

jonathanpoelen:
Simple consistency :)


waruqi:
This is unnecessary, gcc/clang can compile cpp files well, we just need to use g++/clang++ when linking to ensure automatic linking of some C++ libraries.

PR #1023, opened 2020-11-10
ninja generator: support for CLI-specified tools

jonathanpoelen:
Before:

```bash
xmake f --cxx=g++-8 && xmake project -k ninja && ninja
```

```
create ok!
ninja: error: build.ninja:68: unknown build rule 'cxx'
build build/.objs/hello/linux/x86_64/release/src/main.cpp.o: cxx src/mai...
                                                             ^ near here
```

`cxx` rule is missing when `--cxx=...` is used. Same with `ld` and `--ld=...`, etc


waruqi:
Thanks!

PR #1025, opened 2020-11-10
Enable more warnings on "all" settings

SirLynix:
It seems that xmake warnings settings "all" and "more" are mostly equivalents, for MSVC and GCC/Clang. This PR enables more warnings on the "all" settings (using -Wextra and /W4).

### DIFF
--- a/xmake/modules/core/tools/cl.lua
+++ b/xmake/modules/core/tools/cl.lua
@@ -136,7 +136,7 @@ function nf_warning(self, level)
         none       = "-W0"
     ,   less       = "-W1"
     ,   more       = "-W3"
-    ,   all        = "-W4" -- = "-Wall" will enable too more warnings
+    ,   all        = "-W3" -- = "-Wall" will enable too more warnings
     ,   everything = "-Wall"
     ,   error      = "-WX"
     }
--- a/xmake/modules/core/tools/cl.lua
+++ b/xmake/modules/core/tools/cl.lua
@@ -136,7 +136,7 @@ function nf_warning(self, level)
         none       = "-W0"
     ,   less       = "-W1"
     ,   more       = "-W3"
-    ,   all        = "-W3" -- = "-Wall" will enable too more warnings
+    ,   all        = "-W4" -- = "-Wall" will enable too more warnings
     ,   everything = "-Wall"
     ,   error      = "-WX"
     }
--- a/xmake/modules/core/tools/cl.lua
+++ b/xmake/modules/core/tools/cl.lua
@@ -139,6 +139,7 @@ function nf_warning(self, level)
     ,   all        = "-W3" -- = "-Wall" will enable too more warnings
     ,   everything = "-Wall"
     ,   error      = "-WX"
+    ,   extra      = "-W4"
     }
 
     -- make it
--- a/xmake/modules/core/tools/cl.lua
+++ b/xmake/modules/core/tools/cl.lua
@@ -137,9 +137,9 @@ function nf_warning(self, level)
     ,   less       = "-W1"
     ,   more       = "-W3"
     ,   all        = "-W3" -- = "-Wall" will enable too more warnings
+    ,   allextra   = "-W4"
     ,   everything = "-Wall"
     ,   error      = "-WX"
-    ,   extra      = "-W4"
     }
 
     -- make it
--- a/xmake/modules/core/tools/gcc.lua
+++ b/xmake/modules/core/tools/gcc.lua
@@ -112,9 +112,9 @@ function nf_warning(self, level)
     ,   less       = "-Wall"
     ,   more       = "-Wall"
     ,   all        = "-Wall"
+    ,   allextra   = "-Wall -Wextra"
     ,   everything = "-Wall -Wextra -Weffc++"
     ,   error      = "-Werror"
-    ,   extra      = "-Wextra"
     }
     return maps[level]
 end
--- a/xmake/modules/core/tools/gcc.lua
+++ b/xmake/modules/core/tools/gcc.lua
@@ -111,7 +111,7 @@ function nf_warning(self, level)
         none       = "-w"
     ,   less       = "-Wall"
     ,   more       = "-Wall"
-    ,   all        = "-Wall"
+    ,   all        = "-Wall -Wextra"
     ,   everything = "-Wall -Wextra -Weffc++"
     ,   error      = "-Werror"
     }
--- a/xmake/modules/core/tools/gcc.lua
+++ b/xmake/modules/core/tools/gcc.lua
@@ -111,7 +111,7 @@ function nf_warning(self, level)
         none       = "-w"
     ,   less       = "-Wall"
     ,   more       = "-Wall"
-    ,   all        = "-Wall -Wextra"
+    ,   all        = "-Wall"
     ,   everything = "-Wall -Wextra -Weffc++"
     ,   error      = "-Werror"
     }
--- a/xmake/modules/core/tools/gcc.lua
+++ b/xmake/modules/core/tools/gcc.lua
@@ -114,6 +114,7 @@ function nf_warning(self, level)
     ,   all        = "-Wall"
     ,   everything = "-Wall -Wextra -Weffc++"
     ,   error      = "-Werror"
+    ,   extra      = "-Wextra"
     }
     return maps[level]
 end
--- a/xmake/modules/core/tools/nvcc.lua
+++ b/xmake/modules/core/tools/nvcc.lua
@@ -122,7 +122,7 @@ function nf_warning(self, level)
         none       = "-w"
     ,   less       = "-Wall"
     ,   more       = "-Wall"
-    ,   all        = "-Wall"
+    ,   all        = "-Wall -Wextra"
     ,   everything = "-Weverything -Wall -Wextra -Weffc++"
     ,   error      = "-Werror"
     }
--- a/xmake/modules/core/tools/nvcc.lua
+++ b/xmake/modules/core/tools/nvcc.lua
@@ -110,6 +110,7 @@ function nf_warning(self, level)
     ,   all        = "-W3" -- = "-Wall" will enable too more warnings
     ,   everything = "-Wall"
     ,   error      = "-WX"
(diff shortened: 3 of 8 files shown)


xmake/modules/core/tools/gcc.lua
```diff
@@ -111,7 +111,7 @@ function nf_warning(self, level)
         none       = "-w"
     ,   less       = "-Wall"
     ,   more       = "-Wall"
-    ,   all        = "-Wall"
+    ,   all        = "-Wall -Wextra"
```
waruqi, line 114: Currently `all` warning level corresponds to -Wall, if we add `-Wextra`, it will cause too many warnings. This will affect all existing projects with `set_warnings("all", "error")` set, causing compilation interruption.

waruqi: We should add `-Wextra` to ʻeverything` warning level.

SirLynix: -Wextra doesn't add too much warning in my opinion, here's the list:
```


-Wclobbered  
-Wcast-function-type  
-Wdeprecated-copy (C++ only) 
-Wempty-body  
-Wenum-conversion (C only) 
-Wignored-qualifiers 
-Wimplicit-fallthrough=3 
-Wmissing-field-initializers  
-Wmissing-parameter-type (C only)  
-Wold-style-declaration (C only)  
-Woverride-init  
-Wsign-compare (C only) 
-Wstring-compare 
-Wredundant-move (only for C++)  
-Wtype-limits  
-Wuninitialized  
-Wshift-negative-value (in C++03 and in C99 and newer)  
-Wunused-parameter (only with -Wunused or -Wall) 
-Wunused-but-set-parameter (only with -Wunused or -Wall)
```
(from https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html )

everything truly means "every possible warning" in my opinion, which Clangs supports with -Weverything, which is truly too much.

SirLynix: Also, "more" already corresponds to -Wall, it's a bit odd to have "more" and "all" warning levels causing the same flags.

xmake/modules/core/tools/cl.lua
```diff
@@ -136,7 +136,7 @@ function nf_warning(self, level)
         none       = "-W0"
     ,   less       = "-W1"
     ,   more       = "-W3"
-    ,   all        = "-W3" -- = "-Wall" will enable too more warnings
+    ,   all        = "-W4" -- = "-Wall" will enable too more warnings
```
waruqi, line 139: Here too, this will cause more warnings to be triggered, affecting existing projects

SirLynix: I agree this is a breaking change, I leave you the choice on whether it's too much of a change or not (I'll just argue that there's not so many additional warnings, and that simply upgrading a compiler also enables more warnings).

If you decide it's too much, maybe provide an "extra" warning level, enabling `/W4` and `-Wall -Wextra`?

xmake/modules/core/tools/gcc.lua
waruqi: > Also, "more" already corresponds to -Wall, it's a bit odd to have "more" and "all" warning levels causing the same flags.

Therefore, we should improve `more` to set warnings less than `-Wall`

xmake/modules/core/tools/cl.lua
waruqi: We can add a warning level `extra` to correspond to `-Wextra`

`set_warnings("all", "extra", "error")`

xmake/modules/core/tools/gcc.lua
SirLynix: That's a possibility, but it means xmake is responsible of choosing which warnings are enabled by default on GCC/Clang, since they don't have other warnings groups than Wall and Wextra afaik.

xmake/modules/core/tools/cl.lua
SirLynix: Sure, do you want me to try to do it on this PR?

xmake/modules/core/tools/gcc.lua
waruqi: It is for this reason, currently I just correspond `more` to `-Wall` for gcc/clang.

xmake/modules/core/tools/cl.lua
waruqi: Yes, Thanks.

SirLynix:
Here you are, I added the extra warning level when the compiler command line supported it.

Just a note however: enabling both /W3 and /W4 with MSVC will trigger a D9025 warning.
`cl : Command line warning D9025 : overriding '/W3' with '/W4'`

So if "extra" warning level is specified, only /W4 should be passed to the commandline. I don't know how to handle this.

waruqi:
> Here you are, I added the extra warning level when the compiler command line supported it.
> 
> Just a note however: enabling both /W3 and /W4 with MSVC will trigger a D9025 warning.
> `cl : Command line warning D9025 : overriding '/W3' with '/W4'`
> 
> So if "extra" warning level is specified, only /W4 should be passed to the commandline. I don't know how to handle this.

Oh, we should add another level(`allextra`) between `all` and `everything` to correspond to `-Wall -Wextra` or `-W4`.  

less, morr, all, allextra, everything + erroe

SirLynix:
I had the same issue when using
```lua
add_cxflags("-Wall", "-Wextra")
```
before this PR, as this added both /W3 and /W4.

As xmake fully controls the commandline, isn't it possible to let it know some flags conflicts? Like if both /W1 and /W2 are present, only /W2 is kept? (same for /W4)

I know it's a bit more complex, but I have the feeling this will be useful at some other places.

waruqi:
xmake will not deal with these conflicts, users should adjust the configuration according to the error message.

SirLynix:
Isn't xmake purpose to abstract the compiler/backend? I mean if I'm developing exclusively on Linux I won't see MSVC messages.

waruqi:
> Isn't xmake purpose to abstract the compiler/backend? I mean if I'm developing exclusively on Linux I won't see MSVC messages.

? If you are on linux, why do you see msvc’s W3 and W4 conflict information？

SirLynix:
I'm not on Linux, this was a "if" situation 😄.
Let's reverse the thing to be more accurate to my situation, as I'm developing exclusively on Windows I won't see it **if** GCC or Clang were to warn about the command-line. As this one is fully generated by xmake I though it would be nice to check thoses conflicts.

waruqi:
Are you referring to the gcc/clang warning conflict on linux?

xmake only abstracts set_warnings, but if users use add_cflags to set other flags and cause conflicts with set_warnings, xmake will not care about them. Users should consider how to avoid conflicts.

xmake can only guarantee that all flags in set_warnings will not conflict.

SirLynix:
I understand your point.

However, as xmake does translate command-line options, I still think there's a problem.

For example, before I made this PR I did this:
```lua
add_cxflags("-Wall", "-Wextra")
```

xmake abstract and translate flags, so if I'm using Clang/GCC I will have `-Wall -Wextra`, but if I'm using MSVC this will translate to `/W3 /W4`, causing the warning.

Fixing this as a user isn't easy, as I have to check what compiler is used and pass /W4 if using MSVC and -Wall -Wextra on Clang/GCC. It's a bit of a shame to have to do this.

This PR will fix this problem with warning levels, but I'm thinking about the case where other settings conflicts. I don't say we have to handle all of them, but handling at least some of them would be great imho.

waruqi:
Since extra mapping to -W4 will conflict, then for msvc, we can ignore the extra level.

only for gcc/clang: extra -> -Wextra   

and for msvc:  extra -> ignore

SirLynix:
Then "-Wall" would enable /W4 on MSVC?

waruqi:
At present, xmake does not support mapping multiple flags into one flag, `-Wall -Wextra` -> `-W4`, which will make the processing logic inside xmake too complicated, and we will not consider supporting this complex mapping later. relationship.

Therefore, what we can do is simply ignore `-Wextra` for msvc, because this only exists in gcc/clang.

As for the `-W4` abstract support for msvc, as I said before, we can consider adding a warning level: allextra, which will be mapped to `-W4` for msvc/`-Wall -Wextra` for gcc/clang

We only need a simple configuration: `set_warnings("allextra", "error")`

If the user has additionally configured ʻadd_cflags("-Wall", "-Wextra")`, please delete it and use `set_warnings("allextra", "error")`.

Or continue to use ʻadd_cflags("-Wall", "-Wextra")` -> `-W3`. Only one of these two configurations can be used. I think since users want to use set_warnings to abstract warning flags, they should not continue to use add_cflags("-Wall", "-Wextra")`

SirLynix:
> which will make the processing logic inside xmake too complicated, and we will not consider supporting this complex mapping relationship later..
This is why I think it's better to handle it as priorities, if /W4 is set then /W3 is removed (same for /W2, /W1 and /W0 of course), it fixes both problems and is way simpler to implement (as far as I know about xmake internals).

As they are many conflicting options for all compilers (-Wall and -w, /EHsc and /EHs, and more), detecting them in xmake could make it emit a warning when conflicting options are passed (like if I pass /EHs and xmake passes /EHsc) or silently ignore them (/W4 and /W3).

However, this is your project, and I can understand you don't want to complicate the code for what is for now just a tiny warning from one compiler. As I said all I had to say about this, the decision is yours.

I will replace the `extra` warning setting by `allextra` so you can merge this PR which will resolve my issue anyway. 🙂

xmake/modules/core/tools/nvcc.lua
```diff
@@ -125,6 +126,7 @@ function nf_warning(self, level)
     ,   all        = "-Wall"
     ,   everything = "-Weverything -Wall -Wextra -Weffc++"
     ,   error      = "-Werror"
+    ,   extra      = "-Wextra"
```
waruqi, line 129: and here

waruqi:
And clang.lua, can you also add it?

SirLynix:
Here you are, I also handled tiny c compiler.

waruqi:
Ok, thank you very much for your contribution. :+1:

PR #1033, opened 2020-11-14
fix default value of choice options displayed in the graphical menu

jonathanpoelen:
With `option:set_values()` the index of the value was displayed instead of the value.


waruqi:
Thanks! :+1:

PR #1176, opened 2020-12-30
fix debugging flags for ldc2

BarrOff:
Hello,

ldc2 does not accept a `-debug` flag, like dmd does. Instead it uses `--d-debug[=<level/idents>]`


waruqi:
Thanks. 👍

PR #1194, opened 2021-01-12
Add compile mode parameter for windeployqt.exe

knightjun:
When using the install command on a Qt project, the debug library is copied to the installation directory even if the project is set to release mode. This problem can occur on windows systems with multiple versions of Qt installed.
For more details: https://blog.csdn.net/qq_36170958/article/details/108721408

Therefore, when running windeployqt.exe, set the debug or release parameter according to the project mode

### DIFF
--- a/xmake/rules/qt/install/windows.lua
+++ b/xmake/rules/qt/install/windows.lua
@@ -71,7 +71,7 @@ function main(target, opt)
 
     if is_mode("debug") then
         table.insert(argv, "--debug")
-    elseif is_mode("release") then
+    else
         table.insert(argv, "--release")
     end
 
--- a/xmake/rules/qt/install/windows.lua
+++ b/xmake/rules/qt/install/windows.lua
@@ -68,6 +68,13 @@ function main(target, opt)
     else
         table.insert(argv, "--verbose=0")
     end
+
+    if is_mode("debug") then
+        table.insert(argv, "--debug")
+    elseif is_mode("release") then
+        table.insert(argv, "--release")
+    end
+
     if qmldir then
         table.insert(argv, "--qmldir=" .. qmldir)
     end


xmake/rules/qt/install/windows.lua
```diff
@@ -68,6 +68,13 @@ function main(target, opt)
     else
         table.insert(argv, "--verbose=0")
     end
+
+    if is_mode("debug") then
+        table.insert(argv, "--debug")
+    elseif is_mode("release") then
+        table.insert(argv, "--release")
```
waruqi, line 75: and minsizerel, profile, .. other modes.

we can pass `--release` by default

```lua
    if is_mode("debug") then
        table.insert(argv, "--debug")
    else
        table.insert(argv, "--release")
    end
```

SirLynix:
windeployqt should detect if an executable is linked with debug Qt dll or release Qt dll, I don't understand why this is needed.

waruqi:
> When using the install command on a Qt project, the debug library is copied to the installation directory even if the project is set to release mode. This problem can occur on windows systems with multiple versions of Qt installed.
> For more details: https://blog.csdn.net/qq_36170958/article/details/108721408
> 
> Therefore, when running windeployqt.exe, set the debug or release parameter according to the project mode

xmake uses the absolute path to select the correct windeployqt program. 

https://github.com/xmake-io/xmake/blob/0716e54e663f5e6a6400efd1576bdc89c86701d3/xmake/rules/qt/install/windows.lua#L37

knightjun:
> > When using the install command on a Qt project, the debug library is copied to the installation directory even if the project is set to release mode. This problem can occur on windows systems with multiple versions of Qt installed.
> > For more details: https://blog.csdn.net/qq_36170958/article/details/108721408
> > Therefore, when running windeployqt.exe, set the debug or release parameter according to the project mode
> 
> xmake should have used the absolute path to select the correct windeployqt program.
> 
> https://github.com/xmake-io/xmake/blob/0716e54e663f5e6a6400efd1576bdc89c86701d3/xmake/rules/qt/install/windows.lua#L37



> windeployqt should detect if an executable is linked with debug Qt dll or release Qt dll, I don't understand why this is needed.

xmake did find the correct path to windeployqt, and windeployqt provides a detection mechanism, but its detection on my system was actually incorrect. The release mode program was detected as debug mode (Qt5.14.2 msvc2017_64). Search results show that others have the same situation
This should be a bug in windeployqt, but it provides an option to specify the compile mode, so I think xmake should use that option

```
$ /f/Qt/Qt5.14.2/5.14.2/msvc2017_64/bin/windeployqt.exe /f/test/build/windows/x64/release/test.exe
F:\test\build\windows\x64\release\test.exe 64 bit, debug executable
Adding Qt5Svg for qsvgicond.dll
Direct dependencies: Qt5Core Qt5Gui Qt5Widgets
All dependencies   : Qt5Core Qt5Gui Qt5Widgets
To be deployed     : Qt5Core Qt5Gui Qt5Svg Qt5Widgets
```

waruqi:
ok, although I cannot reproduce this problem, it seems that this patch will also not produce other side effects. I can merge it first.

If this patch will affect your project, please let me know. @SirLynix

Issue #1239, opened 2021-02-11
Improve to find system/3rd package for add_requires/find_package.lua in package definition


waruqi:
We can improve the builin find_package script in `add_requires` to support to find more system/3rd package in other package managers.

### Example (libusb)

for apt and some linux systems, it is libusb-1.0

But, we always use libusb as package name to find and use it:

```lua
add_requires("libusb")
target("test")
    set_kind("binary")
    add_files("src/*.c")
    add_packages("libusb")
```

for find_package, it always be libusb.

```console
xmake l find_package libusb
```

### Solution

```lua
package("libusb")
    -- pacman::xxx, brew::xxx ...
    add_extsources("pkg_config::libusb-1.0", "conan::libusb")
```

related issue: https://github.com/xmake-io/xmake/issues/1140

waruqi:
I added `add_extsources()` to add external package sources. https://github.com/xmake-io/xmake/pull/1248

```lua
        add_extsources("pkgconfig::libusb-1.0")
```

And I have improved xmake-repo/libusb to support it. 

https://github.com/xmake-io/xmake-repo/blob/66f5ae98c44320a33275f2ac782effd5b87298f0/packages/l/libusb/xmake.lua#L25-L27

waruqi:
`on_fetch` has been supported.

```lua
package("libusb")
    on_fetch("linux", function(package, opt)
        if opt.system then
            return find_package("pkgconfig::libusb-1.0")
        end
    end)
```

PR #1248, opened 2021-02-21
Improve find_package from external package sources

waruqi:


waruqi:
https://github.com/xmake-io/xmake/issues/1239

PR #1283, opened 2021-03-11
Zig toolchain - Linker

kassane:
Reference:
https://github.com/ziglang/zig/blob/c760532be04667c8a8d49a7cf1b7582da3d1632d/src/main.zig#L189-L192

### DIFF
--- a/xmake/toolchains/zig/xmake.lua
+++ b/xmake/toolchains/zig/xmake.lua
@@ -57,7 +57,17 @@ toolchain("zig")
         local zig = toolchain:config("zig") or "zig"
         toolchain:set("toolset", "cc",    zig .. " cc")
         toolchain:set("toolset", "cxx",   zig .. " c++")
-        toolchain:set("toolset", "ld",    zig .. " c++")
+        if toolchain:is_plat("macosx") then
+            toolchain:set("toolset", "ld",    zig .. " ld64.lld")
+        elseif toolchain:is_plat("linux") then
+            toolchain:set("toolset", "ld",    zig .. " ld.lld")
+        elseif toolchain:is_plat("windows") then
+            toolchain:set("toolset", "ld",    zig .. " lld.link")
+        elseif toolchain:is_plat("wasm") then
+            toolchain:set("toolset", "ld",    zig .. " wasm-ld")
+        else
+            toolchain:set("toolset", "ld",    zig .. " zig c++")
+        end
         toolchain:set("toolset", "sh",    zig .. " c++")
         toolchain:set("toolset", "zc",   "$(env ZC)", zig)
         toolchain:set("toolset", "zcar", "$(env ZC)", zig)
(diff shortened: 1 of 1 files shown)


waruqi:
We should not use ld directly, `zig c++` will automatically choose which linker it is. 

Even for the clang/gcc tool chain, xmake will use clang++/g++ as the linker by default to automatically handle the automatic linking of the c++ library.

In addition, this patch has no effect, it cannot use ld correctly on macos, and linker detection will fail.

```console
ruki-2:console ruki$ xmake f --toolchain=zig -c
checking for SDK version of Xcode for macosx (x86_64) ... 11.0
checking for Minimal target version of Xcode for macosx (x86_64) ... 10.15
ruki-2:console ruki$ xmake -rv
checking for /Users/ruki/Downloads/zig-macos-x86_64-0.8.0-dev.1431+c760532be/zig ld64.lld ... no
checking for the linker (ld: /Users/ruki/Downloads/zig-macos-x86_64-0.8.0-dev.1431+c760532be/zig
ld64.lld) ... no
checking for xcrun -sdk macosx clang++ ... /usr/bin/xcrun -sdk macosx clang++
checking for the linker (ld) ... xcrun -sdk macosx clang++
checking for /Users/ruki/Downloads/zig-macos-x86_64-0.8.0-dev.1431+c760532be/zig c++ ... ok
checking for the c++ compiler (cxx) ... zig c++
checking for /Users/ruki/Downloads/zig-macos-x86_64-0.8.0-dev.1431+c760532be/zig c++ ... ok
checking for flags (-fvisibility-inlines-hidden) ... ok
checking for flags (-O3) ... ok
checking for flags (-DNDEBUG) ... ok
checking for ccache ... /usr/local/bin/ccache
[ 25%]: ccache compiling.release src/main.cpp
/usr/local/bin/ccache /Users/ruki/Downloads/zig-macos-x86_64-0.8.0-dev.1431+c760532be/zig c++ -c
-target x86_64-macos-gnu -fvisibility=hidden -fvisibility-inlines-hidden -O3 -DNDEBUG -o build/.o
bjs/test/macosx/x86_64/release/src/main.cpp.o src/main.cpp
checking for flags (-MMD -MF) ... ok
checking for flags (-fdiagnostics-color=always) ... ok
[ 50%]: linking.release test
"/usr/bin/xcrun -sdk macosx clang++" -o build/macosx/x86_64/release/test build/.objs/test/macosx/
x86_64/release/src/main.cpp.o -arch x86_64 -mmacosx-version-min=10.15 -isysroot /Applications/Xco
de.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk -stdlib=libc++
-Wl,-x -lz
[100%]: build ok!
````
> checking for the linker (ld: /Users/ruki/Downloads/zig-macos-x86_64-0.8.0-dev.1431+c760532be/zig
ld64.lld) ... no

This is the reason why I do not consider using the original ld directly, we need to deal with more things to support it. And the ld of each platform needs to be processed differently.

PR #1303, opened 2021-03-23
Add scons buildsystem to package.tools

PucklaMotzer09:
This adds the scons buildsystem so that it can be more easily used inside packages.

### DIFF
--- a/xmake/modules/package/tools/scons.lua
+++ b/xmake/modules/package/tools/scons.lua
@@ -1,0 +1,36 @@
+--!A cross-platform build utility based on Lua
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/scons.lua
```diff
@@ -0,0 +1,36 @@
...
+import("core.base.option")
+import("lib.detect.find_tool")
+
+-- build package
+function build(package, configs, opt)
+    opt = opt or {}
+    local buildir = opt.buildir or os.curdir()
+    local njob = opt.jobs or option.get("jobs") or tostring(math.ceil(os.cpuinfo().ncpu * 3 / 2))
+    local scons = assert(find_tool("scons"), "scons not found!")
+    local argv = {"-C", buildir, "-j", njob}
```
waruqi, line 31: Can you add verbose output?

```lua
if option.get("verbose") then
    table.insert("argv", ...) --  scons VERBOSE=1?
end
```

PucklaMotzer09: I looked at the output of ```scons -H``` and there seems to be no such thing as verbose output. ```-v``` is used for version displaying. At default scons outputs every invoked compiler call.

waruqi: try `scons VERBOSE=1`?

https://scons.org/doc/1.0.1/HTML/scons-user/x1884.html

PucklaMotzer09: This does not seem to change anything. I also see every compiler call. But we could leave it in anyway.

waruqi: ok, I merged it first. we can improve it later.

PR #1405, opened 2021-05-11
add portage support

Phate6660:
This PR will add support for Gentoo's Portage. 
I have tested with multiple packages that finding and installing works.
If I missed anything or anything needs to be changed, I'll change it right away.


waruqi:
I have merged it, thanks! :+1:

and I will fix appveyor ci fails.

PR #1407, opened 2021-05-12
remove gentoolkit dep, initial implementation of getting package files in pure lua

Phate6660:
This PR will remove the gentoolkit dep and is an initial 
implementation of getting the package file list in pure lua.


waruqi:
Thanks, and I have improved this patch, you can try it again.

Phate6660:
Just tested again with both an installed, and not installed package, and it appears to still work fine. :+1:

PR #1434, opened 2021-06-01
Move vsxmake.autoupdate dependfile to the persistent .xmake cache folder

SirLynix:
plugin.vsxmake.autoupdate relies on a temporary file to know if it should regenerate the vsxmake project or not. This can be annoying because Windows will regulary remove it (everyday at least once, or when you reboot your computer), regenerating the project (which may take some time).

This moves the dependfile to the persistent .xmake cache folder where it will not be automatically deleted by the system.

### DIFF
--- a/xmake/rules/plugin/vsxmake/xmake.lua
+++ b/xmake/rules/plugin/vsxmake/xmake.lua
@@ -37,7 +37,7 @@ rule("plugin.vsxmake.autoupdate")
         import("core.base.task")
 
         -- run only once for all xmake process in vs
-        local tmpfile = os.tmpfile(path.join(os.projectdir(), "plugin.vsxmake.autoupdate"))
+        local tmpfile = path.join(os.projectdir(), ".xmake", "plugin.vsxmake.autoupdate")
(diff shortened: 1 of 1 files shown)


xmake/rules/plugin/vsxmake/xmake.lua
```diff
@@ -37,7 +37,7 @@ rule("plugin.vsxmake.autoupdate")
         import("core.base.task")
 
         -- run only once for all xmake process in vs
-        local tmpfile = os.tmpfile(path.join(os.projectdir(), "plugin.vsxmake.autoupdate"))
+        local tmpfile = path.join(os.projectdir(), ".xmake", "plugin.vsxmake.autoupdate")
```
waruqi, line 40: please use local build/.gens direcrory

```lua
import("core.project.config")
local tmpfile = path.join(config.buildir(), ".gens", "rules", "plugin.vsxmake.autoupdate")
```

like this

https://github.com/xmake-io/xmake/blob/171780c824fdacef1173f6363c8c173b9906ab9d/xmake/core/project/target.lua#L890

Issue #1506, opened 2021-07-10
c/c++项目add_define()定义的内容为字符串时，compile_commands.json配置错误


Nats-ji:
> Note: If you use the fuzzy expressions such as 'can't use/no effect/problem/error', but don't give any relevant auxiliary information according to the template below, this issue will be not replied.

### Describe the bug
在xmake.lua中添加如下define
```lua
add_defines("Hello=\"hello\"")
```

xmake会在 compile_commands.json 中会生成 `"/DHELLO=\"hello\""`，导致intellisense报错：

![image](https://user-images.githubusercontent.com/13299626/125176048-06b16380-e203-11eb-8248-994f0eca7ef6.png)

### Expected behavior

正确的 compile_commands.json 配置应该为 `"/DHELLO=\\\"hello\\\""`：

![image](https://user-images.githubusercontent.com/13299626/125176193-29904780-e204-11eb-89c0-ff4495677930.png)

### Error output
If applicable, please add verbose the errors output to help us analyze your problem.

> If it is a xmake related compilation problem, please add the `-v -D` parameter and give the detailed compilation output information.

### Related Environment

Please provide compiling and running environment information:

- xmake version: 2.5.5
- os: windows
- target platform: windows

waruqi:
最近一直没时间看，非常感谢。 👍

PR #1508, opened 2021-07-12
fix double quote escape for defines in compile_commands.json

Nats-ji:
Fixes the issue reported at: https://github.com/xmake-io/xmake/issues/1506

Result:

`add_defines("HELLO=\"hello\"")`

image


waruqi:
thanks

PR #1510, opened 2021-07-12
Makefile generation: fix circular dependency when targetdir is "."

### LINKED ISSUE #1509: Makefile generation: circular dependency when targetdir is "."
> Note: If you use the fuzzy expressions such as 'can't use/no effect/problem/error', but don't give any relevant auxiliary information according to the template below, this issue will be not replied.

### Describe the bug
The issue happen when a makefile is generated by `xmake project -k make`.

When `set_targetdir(".")` is called, the generated makefile is incorrect and make will warn about a "circular dependency" issue.

### Expected behavior
Make should not show an error.

### Error output
When calling `make`
```
make: Circular sampletarget <- sampletarget dependency dropped.
```

### Related Environment
Please provide compiling and running environment information:

- xmake version: 2.5.5
- os: Fedora Linux
- target platform: -

### Additional context
The faulty `xmake.lua`:

```
target("sampletarget")
    set_kind("binary")
    set_targetdir(".")  -- targetdir is .
    add_files("*.c")
```

norech:
Fixes #1509.

### DIFF
--- a/xmake/plugins/project/make/makefile.lua
+++ b/xmake/plugins/project/make/makefile.lua
@@ -220,8 +220,17 @@ function _make_target(makefile, target, targetflags)
 
     -- make head
     local targetfile = target:targetfile()
-    makefile:print("%s: %s", target:name(), targetfile)
-    makefile:printf("%s:", targetfile)
+    local targetname = target:name()
+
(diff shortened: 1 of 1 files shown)


waruqi:
thanks!

PR #1532, opened 2021-07-26
Move config check before package installation

SirLynix:


xmake/actions/config/main.lua
```diff
@@ -395,6 +395,9 @@ force to build in current directory via run `xmake -P .`]], os.projectdir())
         config.set("buildir", path.relative(buildir, project.directory()), {readonly = true, force = true})
     end
 
+    -- check configs
+    _check_configs()
+    
     -- only config for building project using third-party buildsystem
     if not trybuild then
```
waruqi, line 403: please move into here.

waruqi: trybuild mode need not it

SirLynix: Done

Issue #1567, opened 2021-08-09
Package "lock file" support to freeze dependencies


jorgenpt:
### Is your feature request related to a problem? Please describe.

When I check in requirement for my project, I want to make sure that everyone else (including the build server) use the exact same version until we intentionally upgrade. Using `add_requirement("foo ~1.2")` is helpful to constrain the version, but I don't want xmake to upgrade the package unless I explicitly e.g. run `xmake upgrade-requirements`. If xrepo adds support for transitive dependencies (i.e. my package can depend on a different package), I'd like to be able to stick with a specific version of those dependencies as well.

I also want to make sure that the dependencies haven't been tampered with, so I'd like some guarantee that the checksums in the package lua file haven't been modified without bumping the version.

### Describe the solution you'd like

Many package managers ([npm w/ package.lock](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json), [cargo w/ cargo.lock](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html), [pip freeze w/ requirements.txt](https://pip.pypa.io/en/stable/cli/pip_freeze/), etc) use a `.lock` file that specifies more specific information about the dependencies I've installed. 

For xrepo, this could be a file that contains URLs to all my packages' `xmake.lua` files, the version of the packages, and the specific commit hash of the `xmake.lua` file that should be used. By default, installing requirements via e.g. `xmake` should default to using the specific package listed in the lock file if it is present, otherwise it should use the current rules to determine the version and record it in a .lock file. Then there should be an explicit command to update my packages that would re-evaluate any requirements listed in the xmake.lua and update xmake.lock to the latest version of the package that meets the requirements and the latest commit of the package's `xmake.lua` file.

waruqi:
I have supported it in https://github.com/xmake-io/xmake/pull/1587

We need only add `set_policy("package.requires_lock", true)` to enable `xmake-requires.lock` file

```lua
add_requires("zlib")
set_policy("package.requires_lock", true)

target("test1")
    set_kind("binary")
    add_files("src/*.c")
    add_packages("zlib")
```

We can use `xmake-requires.lock` to lock packages for each platform/archs, it will also lock their repositories with the special commit/branch and their install scripts.

```console
$ cat ./xmake-requires.lock
{
    __meta__ = {
        version = "1.0"
    },
    ["zlib#ddbdd8b8f76d455a8c0ef45fffe37160"] = {
[...]

PR #1587, opened 2021-08-19
Support xmake-requires.lock to lock all required packages

waruqi:


waruqi:
https://github.com/xmake-io/xmake/issues/1567

PR #1589, opened 2021-08-19
Change implib name on mingw (.lib -> .dll.a)

ImperatorS79:
On mingw, implib have generally the .dll.a extension instead of .lib. You can see for reference any msys2 package, for exemple: https://packages.msys2.org/package/mingw-w64-x86_64-chipmunk?repo=mingw64

### DIFF
--- a/xmake/platforms/mingw/xmake.lua
+++ b/xmake/platforms/mingw/xmake.lua
@@ -31,7 +31,7 @@ platform("mingw")
     set_archs("i386", "x86_64", "arm", "arm64")
 
     -- set formats
-    set_formats("static", "$(name).lib")
+    set_formats("static", "$(name).a")
     set_formats("object", "$(name).obj")
     set_formats("shared", "$(name).dll")
     set_formats("binary", "$(name).exe")
--- a/xmake/modules/core/tools/gcc.lua
+++ b/xmake/modules/core/tools/gcc.lua
@@ -334,7 +334,7 @@ function linkargv(self, objectfiles, targetkind, targetfile, flags, opt)
 
     -- add `-Wl,--out-implib,outputdir/libxxx.a` for xxx.dll on mingw/gcc
     if targetkind == "shared" and is_plat("mingw") then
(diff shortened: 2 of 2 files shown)


waruqi:
We don't need to modify this, here it only takes effect for the shared target deps defined by xmake, and xmake generates .lib by default, this change will break the compilation.

Link dependent packages will not execute this line of code

SirLynix:
Since they can be installed with `xmake install`, I think too that this is a bit confusing, people might get `.lib` and think they come from MSVC instead of MinGW.

waruqi:
> Since they can be installed with `xmake install`, I think too that this is a bit confusing, people might get `.lib` and think they come from MSVC instead of MinGW.

That should be modified here at the same time, https://github.com/xmake-io/xmake/blob/9c096fc45a4fb8ac817eb48ab8e762e2b41ab259/xmake/platforms/mingw/xmake.lua#L34

But I am not sure if there are other effects, because I remember that I used .a a long time ago, and then changed to .lib

xmake/platforms/mingw/xmake.lua
```diff
@@ -31,7 +31,7 @@ platform("mingw")
     set_archs("i386", "x86_64", "arm", "arm64")
 
     -- set formats
-    set_formats("static", "$(name).lib")
+    set_formats("static", "$(name).a")
```
waruqi, line 34: `lib$(name).a`

waruqi:
If you want to modify the suffix to `.a`, it may break in many places. At present, some of the things I can think of are: `xmake package` and `xmake install` also need to be adapted and supported, otherwise there will be some problems.

`.dll.a` files need be packaged and installed for these actions.

ImperatorS79:
Okay, noted

waruqi:
see https://github.com/xmake-io/xmake/pull/1675

Many changes have been made to this patch, which may break the compilation of the mingw platform. Everyone can test more first.

@SirLynix @x1244 @ImperatorS79

PR #1594, opened 2021-08-20
Fix simple support for c++20 modules on VS

TheVeryDarkness:
Update build_modulefiles.lua to fix the support on Microsoft Visual Studio 2022 Preview for simple projects using c++20 modules.

It also works with the latest VS 2019.

But I've met problems when a module imports another one. Let me investigate on it for some time then.


xmake/rules/c++/modules/build_modulefiles.lua
```diff
@@ -103,11 +103,11 @@ function _build_modulefiles_msvc(target, sourcebatch, opt)
     for _, sourcefile in ipairs(sourcebatch.sourcefiles) do
         local objectfile = target:objectfile(sourcefile)
         local dependfile = target:dependfile(objectfile)
-        local modulefile = objectfile .. ".pcm"
+        local modulefile = objectfile .. ".ifc"
 
         -- compile module file to *.pcm
         local singlebatch = {sourcekind = "cxx", sourcefiles = {sourcefile}, objectfiles = {objectfile}, dependfiles = {dependfile}}
-        opt.configs.cxxflags = {"/experimental:module /module:interface /module:output " .. os.args(modulefile), "/TP"}
+        opt.configs.cxxflags = {"/experimental:module /interface /ifcOutput " .. os.args(modulefile), "/TP"}
```
waruqi, line 110: Is this change compatible with lower versions vs?

TheVeryDarkness: I guess VS2017 may not support c++ modules, and I don't have one to test.
And although I have a VS2019 build tools, xmake doesn't use that even I have configured xmake to it. And it keeps changing my config from --vs=2019 to --vs=2022. May I know how you treat with vswhere.exe?

TheVeryDarkness: Well, You're right, in vs 2015, 2017, 2019, it's still "/experimental:module" but not "/module". Then, how to judge the version of vs?

waruqi: use has_flags to detect it. it should have been checked. 

https://github.com/xmake-io/xmake/blob/bc582fd4bdaee3add88a4e15f1764c7e11eec41c/xmake/rules/c%2B%2B/modules/build_modulefiles.lua#L137

xmake/rules/c++/modules/build_modulefiles.lua
```diff
@@ -19,7 +19,10 @@
 --
 
 -- imports
-import("core.tool.compiler")
+local flag    = import("get_module_flags")
+local json    = import("core.base.json")
+local builder = import("private.action.build.object")
```
waruqi, line 24: import("private.action.build.object", {alias = "builder"})

xmake/rules/c++/modules/build_modulefiles.lua
```diff
@@ -19,7 +19,10 @@
 --
 
 -- imports
-import("core.tool.compiler")
+local flag    = import("get_module_flags")
+local json    = import("core.base.json")
```
waruqi, line 23: only import("core.base.json")

xmake/rules/c++/modules/build_modulefiles.lua
```diff
@@ -19,7 +19,10 @@
 --
 
 -- imports
-import("core.tool.compiler")
+local flag    = import("get_module_flags")
```
waruqi, line 22: import("get_module_flags", {alias = "flag"}) 

In addition, do not use too short names to avoid conflicts with variable names

waruqi: you can rename to `module.lua` and `import("module")`, call `module.getflags()`

xmake/rules/c++/modules/build_modulefiles.lua
```diff
@@ -100,29 +102,109 @@ function _build_modulefiles_msvc(target, sourcebatch, opt)
...
+        moduleMap[name] = modulefile
+        table.insert(uncompiled_modules, sourcefile)
+    end
+
+    while #uncompiled_modules > 0 do
+        local compiled_one_in_this_loop = false
+
+        for _, to_be_compiled in ipairs(uncompiled_modules) do
+
+            local dependTable = moduleDependenciesMap[to_be_compiled]
```
waruqi: lower case please

TheVeryDarkness:
I'm fixing them

xmake/rules/c++/modules/get_module_flags.lua
```diff
@@ -0,0 +1,52 @@
+function get_module_flags(compinst, toolname, opt)
```
waruqi, line 1: we can define main entry `function main(compinst, toolname, opt)` if there is only one interface.

then

```lua
import("get_module_flags")

get_module_flags(compinst, toolname, opt)
```

waruqi:
Is this completely done?

TheVeryDarkness:
I forgot to push the last commit about function name. And I think I've done what I can do at this time.
Well, it mat take a lot of codes if we are going to scan the dependencies by ourselves, but I'm not so used to Lua language and the architecture of this project, maybe I can do more if I would know more about this project.
I'm working on a dependency scanner in Python these days, but I'm not sure about its reliability. And it's too slow, compared with xmake, as my scanner takes seconds to scan a project that contains just not more than ten lines of codes. By the way, maybe clang-scan-deps is a good choice if xmake could call that?

waruqi:
I think this patch cannot be merged yet, and there may be a lot of work to be done.

If I have time later, I will try to improve c++ modules. At the same time, thank you very much for your contribution to xmake, and I will refer to your patch later.

TheVeryDarkness:
@waruqi hello, my scanning tool in Python has been open source on [scan.py](https://github.com/TheVeryDarkness/umake/blob/main/scan.py). And all codes in that project are under MIT License. You may want to reference or modify it if you are going to handle module dependencies inside xmake.

waruqi:
ok, thanks

waruqi:
I will improve it in v2.5.9. https://github.com/xmake-io/xmake/issues/1729

waruqi:
I initially fixed some module flags of msvc and clang in https://github.com/xmake-io/xmake/pull/1740

In addition, I also added initial support for gcc.

Before improving the support for inter-module dependency graphs, I need some examples of modules with complex dependencies for testing.

Can you help me add some project examples with dependencies between modules in https://github.com/xmake-io/xmake/tree/master/tests/projects/c%2B%2B/modules? Thanks!

TheVeryDarkness:
Sorry, I was busy doing some other work, and I didn't check my emails for github notifications. Maybe I'll check if it works with my projects later.

PR #1659, opened 2021-09-08
Improve cmake

SirLynix:
Previous fix has some issues in some cases, this is a better fix (but not the best we can make)


waruqi:
It should work, I merged it first, because I am goting to sleep. you can open a new pr again if it does not work.

PR #1847, opened 2021-11-21
Install `bzip2` with `yum`

### LINKED ISSUE #1846: `bzip2` missing when installing xmake on Fedora
### Describe the bug

When I install xmake using the installation script, and then try to compile a project, I get an error saying that `tar` is missing `lbzip2`.

### Expected behavior

Installing xmake with the installation script installs everything necessary to build project and install packages from xrepo.

### Error output

```
  => download https://www.x.org/archive/individual/util/util-macros-1.19.2.tar.bz2 .. failed

we can also download these packages manually:
  - https://www.x.org/archive/individual/util/util-macros-1.19.2.tar.bz2
to the local search directories:
  - util-macros-1.19.2.tar.bz2
and we can run `xmake g --pkg_searchdirs=/xxx` to set the search directories.
error: tar (child): lbzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
/usr/bin/tar: Child returned status 2
/usr/bin/tar: Error is not recoverable: exiting now
```

### Related Environment

- xmake version: master
- os: Fedora 34
- target platform: Fedora 34

### Additional context

Manually installing `bzip2` (with `yum` or `dnf`) fixes the problem.

nasso:
We might want to install it with other package managers if necessary, but I've only ever encountered this problem on Fedora with `yum`/`dnf`.

Fixes #1846.

### DIFF
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -150,7 +150,7 @@ test_tools()
 install_tools()
 {
     { apt --version >/dev/null 2>&1 && $sudoprefix apt install -y git build-essential libreadline-dev ccache; } ||
-    { yum --version >/dev/null 2>&1 && $sudoprefix yum install -y git readline-devel ccache && $sudoprefix yum groupinstall -y 'Development Tools'; } ||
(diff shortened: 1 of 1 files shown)


waruqi:
thanks

PR #1876, opened 2021-12-02
Add nondepflags (fix #1824)

### LINKED ISSUE #1824: When building a Windows project in commandline and then with VS, everything is rebuilt
### Describe the bug
I noticed this behavior for the last year, when you build a project using xmake in commandline and then switch to a visual studio project, every file built in commandline has to be rebuilt in visual studio.

I did look at xmake internals to understand why xmake though files needed to be rebuilt and finally understood it, it comes from the additionnal `-FC` flag added by xmake when using VS.

From modules/core/tools/cl.ua:
```lua
    -- we need show full file path to goto error position if xmake is called in vstudio
    -- https://github.com/xmake-io/xmake/issues/1049
    if os.getenv("XMAKE_IN_VSTUDIO") then
        self:add("cxflags", "-FC")
    end
```
https://github.com/xmake-io/xmake/blob/f9af5b38ecf35c9e5741542e7a0f02f9dd0409f8/xmake/modules/core/tools/cl.lua#L37-L41

The problem is that -FC gets into depends depvalues, and as depvalues change between VS and commandline, xmake thinks file needs to be recompiled.

Maybe that flag should be added after checking depends? Or that depends should ignore some flags that won't affect object/errors/warnings generation (such as -FC).

### Expected behavior
That files don't get rebuild without change between commandline and VS.

### Related Environment

- xmake version: 2.5.9
- os: Windows 10
- target platform: Windows 10

SirLynix:


xmake/core/tool/compiler.lua
```diff
@@ -335,5 +335,27 @@ function compiler:compflags(opt)
...
+    end
+
+    local filteredflags = {}
+    for _, flag in ipairs(flags) do
+        if not nondepflags[flag] then
+            table.insert(filteredflags, flag)
+        end
+    end
+
+    return filteredflags
```
waruqi, line 357: This will affect compilation performance, because every time we compile a source file, we need to traverse all the flags additionally.

So I haven't tried this before.

waruqi:
I think that we just add -FC here

https://github.com/xmake-io/xmake/blob/690d8ae73adb4e568d62c3b082f4b2d079a20bb6/xmake/modules/core/tools/cl.lua#L415

SirLynix:
It seems to be a better fix, I'll open a PR with that change 👍

PR #1878, opened 2021-12-02
Add -FC after depends check (fix #1824)

### LINKED ISSUE #1824: When building a Windows project in commandline and then with VS, everything is rebuilt

SirLynix:


xmake/modules/core/tools/cl.lua
```diff
@@ -413,12 +407,19 @@ function compile(self, sourcefile, objectfile, dependinfo, flags, opt)
...
             if dependinfo then
                 if _has_source_dependencies(self) then
                     depfile = os.tmpfile()
-                    compflags = table.join(flags, "/sourceDependencies", depfile)
+                    compflags = table.join(compflags, "/sourceDependencies", depfile)
                 else
-                    compflags = table.join(flags, "-showIncludes")
+                    compflags = table.join(compflags, "-showIncludes")
                 end
             end
```
waruqi, line 425: we should add it here

```lua
   if os.getenv("XMAKE_IN_VSTUDIO") then
          if compflags == flags then
              compflags = table.join(compflags, "-FC")
          else
               table.join2(compflags, "-FC")
          end
   end
```

xmake/modules/core/tools/cl.lua
```diff
@@ -413,12 +407,19 @@ function compile(self, sourcefile, objectfile, dependinfo, flags, opt)
 
             -- generate includes file
             local compflags = flags
+
+            -- we need show full file path to goto error position if xmake is called in vstudio
+            -- https://github.com/xmake-io/xmake/issues/1049
+            if os.getenv("XMAKE_IN_VSTUDIO") then
```
waruqi, line 413: we need cache the env result.

PR #1892, opened 2021-12-07
Fix detecting of Intel Fortran Compiler 2021 on Windows

PucklaMotzer09:
Previously the compiler could not be detected, because xmake was just looking for the environment variable IFORT_COMPILER20, but now it is also looking for IFORT_COMPILER21.

### DIFF
--- a/xmake/modules/detect/sdks/find_ifortenv.lua
+++ b/xmake/modules/detect/sdks/find_ifortenv.lua
@@ -100,11 +100,11 @@ function _find_intel_on_windows(opt)
     opt = opt or {}
 
     -- find ifortvars_bat.bat
-    local paths = {os.getenv("IFORT_COMPILER20")}
+    local paths = {"$(env IFORT_COMPILER20)"}
     local ifortvars_bat = find_file("bin/ifortvars.bat", paths)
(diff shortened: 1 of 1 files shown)


xmake/modules/detect/sdks/find_ifortenv.lua
```diff
@@ -100,8 +100,14 @@ function _find_intel_on_windows(opt)
     opt = opt or {}
 
     -- find ifortvars_bat.bat
-    local paths = {"$(env IFORT_COMPILER20)"}
+    local paths = {os.getenv("IFORT_COMPILER20")}
```
waruqi, line 103: Why use os.getenv instead of $(env)? This will cause environment variables to be calculated in advance. If the searched list is very large, they will be obtained in advance, if one of them returns nil, the table will be interrupted

PucklaMotzer09: My bad, I thought this would be executed as a system command and there is no `env` command on windows. But this does indeed work.

waruqi: $(env) will also call os.getenv

PR #1936, opened 2021-12-22
Vstudio runenv

SirLynix:
I took some code from the vsxmake project to make debug environment work, I have tested it and it work great.


waruqi:
Thanks. 👍

PR #1946, opened 2021-12-27
add CMakeLists headeronly support

Q2h1Cg:
before:

```
$ xmake project -k cmake
error: unknown target kind headeronly
```

after:

```
$ xmake project -k cmake
create ok!
```


waruqi:
thanks

PR #1959, opened 2022-01-06
Improve vstudio projects

SirLynix:
Add supports for the following vcxproj xml nodes:
1. `<ExternalIncludePath>`
1. `<DisableSpecificWarnings>`
1. `<AdditionalLibraryDirectories>`
1. `<AdditionalDependencies>`
1. `<LanguageStandard>`
1. `<LanguageStandard_C>`

I had to move common flags building in a separate function because I needed them in `_make_configurations` (for external include path)

Using `<ExternalIncludePath>` instead of `-external:I` fixes the Intellisense issue with "go to definition" or "open file" erroring on external include paths.

The next big step I would like to implement now is rule support, using phony project inside of VS.


waruqi:
thanks! 👍

PR #2052, opened 2022-02-10
Fix vs generator config

SirLynix:
This way it works right, it needed a recheck to ensure states

### DIFF
--- a/xmake/plugins/project/vstudio/impl/vs201x.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x.lua
@@ -96,6 +96,10 @@ function _clear_cache()
     localcache.clear("option")
     localcache.clear("package")
     localcache.clear("toolchain")
+
+    -- force recheck
+    localcache.set("config", "recheck", true)
+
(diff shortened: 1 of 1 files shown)


xmake/plugins/project/vstudio/impl/vs201x.lua
```diff
@@ -592,11 +593,6 @@ function make(outputdir, vsinfo)
     -- clear local cache
     _clear_cache()
 
-    -- restore previous config data
-    if configdata then
-        io.writefile(config.filepath(), configdata)
-    end
```
waruqi, line 598: Why? xmake.conf maybe has been changed, it's not localcache("config")

SirLynix: xmake.conf is no longer cleared since those lines got removed:
```lua
    config.clear()
    config.save()
```

so its content is not overwrited

waruqi: ok

PR #2112, opened 2022-02-28
fix vsxmake generator

xq114:


xq114:
https://github.com/xmake-io/xmake/blob/a76408793027b1aca6b8938a51e5d121bb42084b/xmake/plugins/project/vsxmake/vsxmake.lua#L191-L192

These two lines could cause all configs in the project to be lost, including the `buildir` and other settings. What are they doing for? How can I preserve the configs in the project while generating vsxmake files?

SirLynix:
> https://github.com/xmake-io/xmake/blob/a76408793027b1aca6b8938a51e5d121bb42084b/xmake/plugins/project/vsxmake/vsxmake.lua#L191-L192
> 
> These two lines could cause all configs in the project to be lost, including the `buildir` and other settings. What are they doing for? How can I preserve the configs in the project while generating vsxmake files?

I will fix them in a separate PR, but you're right they can be safely removed

PR #2157, opened 2022-03-13
Support mesh / raytracing shaders and customizable target environment

Shan-Zhou:
Now the utils.glsl2spv rule supports mesh and ray tracing shaders.
The target environment can be set by `add_rules("utils.glsl2spv", {outputdir = "build", targetenv = "vulkan1.2"})`. If not set explicitly, the default target environment would be `vulkan1.0`.

### DIFF
--- a/xmake/rules/utils/glsl2spv/xmake.lua
+++ b/xmake/rules/utils/glsl2spv/xmake.lua
@@ -57,7 +57,7 @@ rule("utils.glsl2spv")
         assert(glslangValidator or glslc, "glslangValidator or glslc not found!")
 
         -- glsl to spv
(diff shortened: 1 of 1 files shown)


xmake/rules/utils/glsl2spv/xmake.lua
```diff
@@ -57,14 +57,15 @@ rule("utils.glsl2spv")
         assert(glslangValidator or glslc, "glslangValidator or glslc not found!")
 
         -- glsl to spv
+        local targetenv = target:extraconf("rules", "utils.glsl2spv", "env") or "vulkan1.0"
```
waruqi, line 60: I think it would be better to use `targetenv` instead of `env`

PR #2183, opened 2022-03-20
Add resource files include and pre-processor defines for the VS gener…

yamashi:


xmake/plugins/project/vstudio/impl/vs201x_vcxproj.lua
```diff
@@ -877,6 +910,12 @@ function _make_common_item(vcxprojfile, vsinfo, target, targetinfo)
 
     vcxprojfile:leave("</ClCompile>")
     
+    vcxprojfile:enter("<ResourceCompile>")
+        -- make resource options
+        _make_resource_options_cl(vcxprojfile, targetinfo.commonflags.cl)
```
waruqi, line 915: two args? condition arg is nil

xmake/plugins/project/vstudio/impl/vs201x_vcxproj.lua
```diff
@@ -493,6 +493,39 @@ function _make_source_options_cl(vcxprojfile, flags, condition)
     end
 end
 
+-- make source options for cl
+function _make_resource_options_cl(vcxprojfile, flags, condition)
+
+    -- exists condition?
+    condition = condition or ""
```
waruqi, line 500: condition is unused

yamashi: Is is like 524

PR #2244, opened 2022-04-06
vstudio generator: fix batchcmd dir

SirLynix:
I don't think this is the best way to fix this. I think we should add a `batchcmd:path` function which will translate path accordingly

### DIFF
--- a/xmake/plugins/project/vstudio/impl/vs201x.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x.lua
@@ -111,7 +111,7 @@ function _get_command_string(cmd, vcxprojdir)
 end
 
 -- add target custom commands for target
-function _make_custom_commands_for_target(commands, target, suffix, vcxprojdir)
+function _make_custom_commands_for_target(commands, target, vcxprojdir, suffix)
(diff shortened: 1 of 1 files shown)


xmake/plugins/project/vstudio/impl/vs201x.lua
```diff
@@ -192,19 +199,19 @@ function _make_custom_commands_for_objectrules(commands, target, sourcebatch, su
 end
 
 -- make custom commands
-function _make_custom_commands(target)
+function _make_custom_commands(target, vcxprojdir)
     local commands = {}
-    _make_custom_commands_for_target(commands, target, "before")
-    _make_custom_commands_for_target(commands, target)
+    _make_custom_commands_for_target(commands, target, "before", vcxprojdir)
+    _make_custom_commands_for_target(commands, target, nil, vcxprojdir)
```
waruqi, line 205: you can adjust vcxprojdir position to avoid nil argument.

```
_make_custom_commands_for_target(commands, target, vcxprojdir)
    _make_custom_commands_for_target(commands, target, vcxprojdir, "before")

```

PR #2264, opened 2022-04-13
remove extra_cflags from gn as it isn't standard

yamashi:


waruqi:
so, how do I add extra cflags for gn? we need set vs runtime

yamashi:
It's dependant of the project, some use a list as extra_clfags, others use a string, so it has to be set per project depending on which one they expect :(

PR #2269, opened 2022-04-14
update(rules/platform/linux/bpf): enforce -O2 when compiling BPF

leodido:
To avoid compiler optimizations messing up with generated IR and BPF instructions is paramount to enforce the optimization level to 2 disregarding the mode (whether it is set to debug or release etc.)

### DIFF
--- a/xmake/rules/platform/linux/bpf/xmake.lua
+++ b/xmake/rules/platform/linux/bpf/xmake.lua
@@ -46,6 +46,7 @@ rule("platform.linux.bpf")
             targetarch = "__TARGET_ARCH_powerpc"
         end
         target:add("includedirs", path.directory(headerfile))
+        target:set("optimize", "faster")
         batchcmds:show_progress(opt.progress, "${color.build.object}compiling.bpf %s", sourcefile)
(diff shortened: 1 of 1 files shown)


xmake/rules/platform/linux/bpf/xmake.lua
```diff
@@ -46,6 +46,7 @@ rule("platform.linux.bpf")
             targetarch = "__TARGET_ARCH_powerpc"
         end
         target:add("includedirs", path.directory(headerfile))
+        target:set("optimize", "faster")
         batchcmds:show_progress(opt.progress, "${color.build.object}compiling.bpf %s", sourcefile)
         batchcmds:mkdir(path.directory(objectfile))
         batchcmds:compile(sourcefile, objectfile, {configs = {force = {cxflags = {"-target bpf", "-g"}}, defines = targetarch}})
```
waruqi, line 52: Why not add `-O2` to here? it only for compiling bpf code `cxflags = {"-target bpf", "-g", "-O2"}`

leodido: If you prefer that way I can do it that way :)

PR #2311, opened 2022-04-25
fix generated cmakelists when using `set_languages("c++latest")`

Latias94:
According to [CMAKE_CXX_KNOWN_FEATURES](https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html), the corresponding value for `set_languages("c++latest")` should be `cxx_std_23`. This value is new in version 3.20.

### DIFF
--- a/xmake/plugins/project/cmake/cmakelists.lua
+++ b/xmake/plugins/project/cmake/cmakelists.lua
@@ -450,6 +450,7 @@ function _add_target_languages(cmakelists, target)
     ,   cxx14 = "cxx_std_14"
     ,   cxx17 = "cxx_std_17"
     ,   cxx20 = "cxx_std_20"
+    ,   cxxlatest = "cxx_std_23"
     }
     local languages = target:get("languages")
     if languages then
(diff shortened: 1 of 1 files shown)


xmake/plugins/project/cmake/cmakelists.lua
```diff
@@ -450,6 +450,7 @@ function _add_target_languages(cmakelists, target)
     ,   cxx14 = "cxx_std_14"
     ,   cxx17 = "cxx_std_17"
     ,   cxx20 = "cxx_std_20"
+    ,   cxxlatest = "cxx_std_23"
```
waruqi, line 453: We can't simply set cxx_std_23. This is because different compiler versions support different versions of the latest std.

If we encounter a low version of the compiler, cmake will encounter an error. Even if the corresponding std version is automatically generated by native compiler detection, you may still encounter problems if you distribute cmakelists.txt to other users.

Latias94: For g++.exe (GCC) 11.2.0, the flag will be `-std=gnu++23`. 
For Clang-cl 14 and vs2022, the flag will be `-std:c++latest`. 
You are right when I test lower version of the compiler. However, find out the latest support c++ version for each compilers may be beyond my ability. Or I just simply add `cxx23` support?

waruqi: Since cmake doesn't natively support c++lastest, maybe we shouldn't add c++latest to cmakelists. If you want to use std23, I suggest you should just add `cxx23 = "cxx_std_23"`

waruqi:
thanks

PR #2440, opened 2022-06-08
Improve package installation error messages

SirLynix:
```lua
add_requires("libcurle") -- mispelled libcurl
add_requires("util-linux") -- isn't available on Windows
add_requires("openssl", { system = true }) -- isn't present on the system

target("xmake-nonexistantpkg")
    set_kind("binary")
    add_files("src/*.cpp")
```

=>
```
C:\Projets\Perso\Tests\xmake-nonexistantpkg>xmake
checking for platform ... windows
checking for architecture ... x64
checking for Microsoft Visual Studio (x64) version ... 2022
note: the following packages were not found in any repository (check if they are spelled correctly):
  -> libcurle
note: the following packages are unsupported on windows/x64:
  -> util-linux 2.36.2
note: the following packages were not found on your system, try again after installing them:
  -> openssl 1.1.1o
```

~~One issue I have is that I can't detect this case:~~ (fixed)
```lua
package("test")
package_end()

add_requires("test") -- local package that xmake can't install
add_requires("libcurle") -- mispelled libcurl
```

```
note: the following packages were not found in any repository (check if they are spelled correctly):
  -> test
  -> libcurle
```

I don't know how to detect `test`, a locally defined package vs a non-existing package

### DIFF
--- a/xmake/modules/private/action/require/impl/install_packages.lua
+++ b/xmake/modules/private/action/require/impl/install_packages.lua
@@ -669,6 +669,7 @@ function main(requires, opt)
     local packages_download = {}
     local packages_unsupported = {}
     local packages_not_found = {}
+    local packages_unknown = {}
     for _, instance in ipairs(packages) do
         if package.should_install(instance) then
             if instance:is_supported() then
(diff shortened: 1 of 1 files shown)


waruqi:
try this? https://github.com/xmake-io/xmake/blob/a3f81d902ec232527f7b21e08b33d97936fc98d2/xmake/core/package/package.lua#L533-L543

Issue #2482, opened 2022-06-22
Pacman包的 include 路径错误


sunxfancy:
### Xmake 版本

xmake v2.6.8+202206221218

### 操作系统版本和架构

5.13.10-1-MANJARO

### 描述问题

在引用Pacman的包LLVM时，include path中会引入大量不必要且可能引发文件名冲突的子目录，如：
-isystem /usr/include/llvm/DebugInfo/GSYM
-isystem /usr/include/llvm/Transforms/Coroutines
...

我不确定其他pacman的包是否有类似问题，但看起来更像是pacman包引入机制的问题。

### 期待的结果

正常情况下，应该仅仅引入 /usr/include/ 即可，用户会在include时指定 "llvm/DebugInfo/GSYM/..."

### 工程配置

add_rules("mode.debug", "mode.release")
set_languages("c17", "c++20")
add_requires("pacman::llvm")

target("Pass")
    set_kind("shared")
    add_files("compiler/*.cpp", "compiler/*.ixx")
    add_packages("pacman::llvm")

### 附加信息和错误日志

xmake b -r -vD                                                                                                                                     

checking for gcc ... /usr/bin/gcc
checkinfo: cannot runv(dmd --version), No such file or directory
checking for dmd ... no
checkinfo: cannot runv(ldc2 --version), No such file or directory
checking for ldc2 ... no
checkinfo: cannot runv(gdc --version), No such file or directory
checking for gdc ... no
checkinfo: cannot runv(zig version), No such file or directory
checking for zig ... no
checkinfo: cannot runv(zig version), No such file or directory
checking for zig ... no
checking for unzip ... /usr/bin/unzip
checking for git ... /usr/bin/git
checking for gzip ... /usr/bin/gzip
checking for tar ... /usr/bin/tar
finding llvm from pacman ..
checking for pacman ... /usr/bin/pacman
checking for pacman::llvm ... llvm 13.0.1
checking for gcc ... /usr/bin/gcc
checking for the c++ compiler (cxx) ... gcc
checking for /usr/bin/gcc ... ok
checking for flags (-fPIC) ... ok
> gcc "-fPIC"
checking for flags (-fmodules-ts) ... ok
> gcc "-fmodules-ts"
generating.moduledeps runtime/map.ixx
checking for flags (-O3) ... ok
> gcc "-O3"
checking for flags (-std=c++20) ... ok
> gcc "-std=c++20"
checking for flags (-fmodules-ts) ... ok
[...]

PR #2487, opened 2022-06-24
Fix include paths of pacman packages

sunxfancy:
It's a quick fix of `find_package.lua` for pacman packages (issue #2482). The bug is related to the incorrect include paths when the pacman package didn't contain pkg-config files (`*.pc`) indicating the include paths. 

Changes:
Add a check after the scanning of C++ header files in pacman packages. If the path is starting with `/usr/include/`, then we will ignore this header since it's inside the compiler's default include path.


waruqi:
thanks

PR #2497, opened 2022-06-25
Support of custom QML2_IMPORT_PATH and auto-add output path to import custom qmlplugins

Arthapz:


xmake/rules/qt/env/xmake.lua
```diff
@@ -30,13 +30,18 @@ rule("qt.env")
             qt = assert(find_qt(nil, {verbose = true}), "Qt SDK not found!")
             target:data_set("qt", qt)
         end
+
+        local qmlimportpath = target:get("qt.env.qmlimportpath") or {}
```
waruqi, line 34: use target:values

PR #2498, opened 2022-06-25
add icx toolchain

jonathanpoelen:
Support of icx / icpx on linux. I don't know if I should do something special for windows.

I'm looking tomorrow for dpcpp.


waruqi:
Thanks!

waruqi:
> Support of icx / icpx on linux. I don't know if I should do something special for windows.

I don't have icx for windows environment either, so we can improve it later.

PR #2499, opened 2022-06-26
add dpcpp toolchain

jonathanpoelen:


xmake/toolchains/dpcpp/xmake.lua
```diff
@@ -0,0 +1,35 @@
...
+-- @author      ruki
+-- @file        xmake.lua
+--
+
+-- define toolchain
+toolchain("dpcpp")
+
+    -- set homepage
+    set_homepage("https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html")
+    set_description("Intel LLVM C/C++ Compiler")
```
waruqi, line 26: It is also the `Intel LLVM C/C++ Compiler` ? What is the difference between it and icx?

waruqi: Could you use a different description to distinguish between icx and dpcc?

jonathanpoelen: yes, it's another intel compiler that uses SYCL.

But I just realized that it only allows to compile C sources. Is removing `toolchain:set("toolset", "cc", "dpcpp")` enough ? Should I remove `set_kind("standalone")` ?

waruqi: need not remove it

PR #2433, opened 2022-06-05
Non-default target can be built in vs

Domain:
Add conditional template in vsxmake. Fix #2432

Issue #2432, opened 2022-06-03
Can't compile a set_default(false) target with vsxmake2022 project


WubiCookie:
### Xmake Version

2.6.7

### Operating System Version and Architecture

Windows 10 version 1909

### Describe Bug

With this script,
```lua
target("main")
    set_default(false)
    set_kind("binary")
    add_files("main.cpp")
```
the target is never built, even when specifically rebuilding it from Visual Studio.

My use case is a test target for a library. I don't want the tests to automatically be built and/or installed and/or packaged, but I still want to be able to execute them from vs when I'm working on the library.

### Expected Behavior

The target should be built

### Additional Information and Error Logs

Visual Studio output:
```
Rebuild started...
1>------ Rebuild All started: Project: main, Configuration: release x64 ------
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
```

waruqi:
I have merged this patch https://github.com/xmake-io/xmake/pull/2433

PR #2517, opened 2022-06-30
Add project dependency in solution file. Fix #2506

### LINKED ISSUE #2506: vsxmake生成的工程没有依赖关系
### Xmake 版本

2.6.8

### 操作系统版本和架构

Windows 11

### 描述问题

之前讨论过的依赖关系和编译顺序在命令行直接执行xmake的时候运行得很好，但是通过vsxmake生成得工程没有依赖关系，导致生成顺序不能保证

### 期待的结果

vsxmake生成的工程需要有依赖关系。应该是sln文件的Project字段里面需要包含ProjectSection(ProjectDependencies)

### 工程配置

_No response_

### 附加信息和错误日志

_No response_

Domain:

------

### DIFF
--- a/xmake/plugins/project/vsxmake/render.lua
+++ b/xmake/plugins/project/vsxmake/render.lua
@@ -34,29 +34,104 @@ end
 
 function _cfill(opt, params)
     return function(match)
-        local tmp = match:split(";", {strict = true})
+        local tmp = match:split(";", {strict = true, plain = true})
         assert(#tmp == 2 or #tmp == 3)
 
         local cond = tmp[1]
         local value1 = tmp[2]
         local value2 = ""
-        
+        local len = false
+   
         if #tmp == 3 then
             value2 = tmp[3]
         end
 
-        tmp = cond:split("=")
-        assert(#tmp == 2)
+        local ops = { "==", "~=", "<=", ">=", "<", ">", "=" }
+        local op = nil
+        local k = nil
+        local v = nil
 
-        local k = tmp[1]:trim()
-        local v = tmp[2]:trim()
+        for _, o in ipairs(ops) do
+            local i, j = cond:find(o)
+            if i ~= nil then
+                op = o
+                k = cond:sub(1, i - 1):trim()
+                v = cond:sub(j + 1):trim()
+                break
+            end
+        end
 
-        if opt.paramsprovider(k, params) == v then
+        assert(op)
+        assert(k)
+        assert(v)
+        
+        if k:startswith("#") then
+            k = k:sub(2)
+            len = true
+        end
+
+        local m = k:split(".", { plain = true })
+        local p
+
+        local old_target = opt.paramsprovider() or ""
+
+        for i,item in ipairs(m) do
+            if i == 1 then
+                p = opt.paramsprovider(item, params)
+                if item == "target" then
+                    opt.paramsprovider(p)
+                end
+            else
+                p = _expand(opt.paramsprovider({item}))
+            end
+        end
+
+        if _cmp(p, v, op, len) then
             return value1
+        end
+
+        if old_target ~= opt.paramsprovider() then
+            opt.paramsprovider(old_target)
         end
 
         return value2
     end
+end
+
+function _cmp(lhs, rhs, op, len)
+    if len then
+        lhs = #lhs
+    end
+
+    if type(lhs) == "number" then
+        rhs = tonumber(rhs)
+    end
+
+    if op == "==" or op == "=" then
+        return lhs == rhs
+    end
+
+    if op == "~=" then
+        return lhs ~= rhs
+    end
+
+    if op == "<=" then
+        return lhs <= rhs
+    end
+
+    if op == ">=" then
+        return lhs >= rhs
+    end
+
+    if op == "<" then
+        return lhs < rhs
+    end
+
+    if op == ">" then
+        return lhs > rhs
+    end
+
+    raise("unknown operator '" .. op .. "'")
 end
 
 function _expand(params)
@@ -84,6 +159,8 @@ function _expand(params)
     return r
 end
 
+local save_target
+
 function _render(templatepath, opt, args)
     local template = io.readfile(templatepath)
     local params = _expand(opt.paramsprovider(args))
@@ -92,10 +169,16 @@ function _render(templatepath, opt, args)
         local tmpl = template:gsub(opt.cpattern, _cfill(opt, v))
         replaced = replaced .. tmpl:gsub(opt.pattern, _fill(opt, v))
     end
+
+    if save_target ~= opt.paramsprovider() then
+        opt.paramsprovider(save_target)
+    end
+
     return replaced
 end
 
 function main(templatepath, pattern, cpattern, paramsprovider)
     local opt = { pattern = pattern, cpattern = cpattern, paramsprovider = paramsprovider, templatedir = path.directory(templatepath) }
+    save_target = paramsprovider() or ""
     return _render(templatepath, opt, {})
 end
--- a/xmake/plugins/project/vsxmake/render.lua
+++ b/xmake/plugins/project/vsxmake/render.lua
@@ -41,18 +41,16 @@ function _cfill(opt, params)
         local value1 = tmp[2]
         local value2 = ""
         local len = false
-   
+
         if #tmp == 3 then
             value2 = tmp[3]
         end
 
         local ops = { "==", "~=", "<=", ">=", "<", ">", "=" }
-        local op = nil
-        local k = nil
-        local v = nil
+        local op, k, v
 
         for _, o in ipairs(ops) do
-            local i, j = cond:find(o)
+            local i, j = cond:find(o, 1, true)
             if i ~= nil then
                 op = o
                 k = cond:sub(1, i - 1):trim()
@@ -61,9 +59,7 @@ function _cfill(opt, params)
             end
         end
 
-        assert(op)
-        assert(k)
-        assert(v)
+        assert(op and k and v)
         
         if k:startswith("#") then
             k = k:sub(2)
@@ -72,10 +68,8 @@ function _cfill(opt, params)
 
         local m = k:split(".", { plain = true })
         local p
-
         local old_target = opt.paramsprovider() or ""
-
-        for i,item in ipairs(m) do
+        for i, item in ipairs(m) do
             if i == 1 then
                 p = opt.paramsprovider(item, params)
                 if item == "target" then
--- a/xmake/plugins/project/vsxmake/vsproj/templates/sln/ProjConfig(target,mode,arch)
+++ b/xmake/plugins/project/vsxmake/vsproj/templates/sln/ProjConfig(target,mode,arch)
@@ -1,2 +1,2 @@
-		{#target_id#}.#mode#|#arch#.ActiveCfg = #mode#|#vsarch#@default=true;
+		{#target_id#}.#mode#|#arch#.ActiveCfg = #mode#|#vsarch#@default==true;
 		{#target_id#}.#mode#|#arch#.Build.0 = #mode#|#vsarch#@
--- a/xmake/plugins/project/vsxmake/vsproj/templates/sln/Project(target)
+++ b/xmake/plugins/project/vsxmake/vsproj/templates/sln/Project(target)
@@ -1,2 +1,4 @@
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "#target#", "#target#\#target#.vcxproj", "{#target_id#}"
-EndProject
+@#target.dep > 0;	ProjectSection(ProjectDependencies) = postProject
+#Import(ProjectDep)#	EndProjectSection
+@EndProject
--- a/xmake/plugins/project/vsxmake/vsproj/templates/sln/ProjectDep(dep)
+++ b/xmake/plugins/project/vsxmake/vsproj/templates/sln/ProjectDep(dep)
@@ -1,0 +1,1 @@
+		{#target_id#} = {#target_id#}
--- a/xmake/plugins/project/vsxmake/vsxmake.lua
+++ b/xmake/plugins/project/vsxmake/vsxmake.lua
@@ -57,7 +57,7 @@ function _buildparams(info, target, default)
         local i = info
         local r = info[match]
         if target then
-            opt = table.join(target, opt)
+            opt = table.unique(table.join(target, opt))
         end
         for _, k in ipairs(opt) do
             local v = (i._targets or {})[k]
@@ -149,11 +149,25 @@ function _buildparams(info, target, default)
         return r
     end
 
-    return function(match, opt)
-        if type(match) == "table" then
-            return listconfig(match)
+    return function(...)
+        local args = {...}
+        if #args == 0 then
+            return target
         end
-        return getprop(match, opt)
+
+        if type(args[1]) == "string" and #args == 1 then
+            target = args[1]
+            if target == "" then
+                target = nil
+            end
+            return
+        end
+
+        if type(args[1]) == "table" then
+            return listconfig(args[1])
+        end
+
+        return getprop(args[1], args[2])
     end
 end
 


xmake/plugins/project/vsxmake/render.lua
```diff
@@ -34,31 +34,106 @@ end
...
 
-        tmp = cond:split("=")
-        assert(#tmp == 2)
+        local ops = { "==", "~=", "<=", ">=", "<", ">", "=" }
+        local op = nil
+        local k = nil
+        local v = nil
+
+        for _, o in ipairs(ops) do
+            local i, j = cond:find(o)
```
waruqi, line 55:
```lua
cond:find(o, 1, true)
```

xmake/plugins/project/vsxmake/render.lua
```diff
@@ -34,31 +34,106 @@ end
...
         if #tmp == 3 then
             value2 = tmp[3]
         end
 
-        tmp = cond:split("=")
-        assert(#tmp == 2)
+        local ops = { "==", "~=", "<=", ">=", "<", ">", "=" }
+        local op = nil
+        local k = nil
+        local v = nil
```
waruqi, line 52: local op, k, v

xmake/plugins/project/vsxmake/render.lua
```diff
@@ -34,31 +34,106 @@ end
...
+                op = o
+                k = cond:sub(1, i - 1):trim()
+                v = cond:sub(j + 1):trim()
+                break
+            end
+        end
+
+        assert(op)
+        assert(k)
+        assert(v)
```
waruqi, line 66: assert(op and k and v)

xmake/plugins/project/vsxmake/render.lua
```diff
@@ -34,31 +34,106 @@ end
...
+        
+        if k:startswith("#") then
+            k = k:sub(2)
+            len = true
+        end
+
+        local m = k:split(".", { plain = true })
+        local p
+
```
waruqi, line 75: 去掉冗余空行

xmake/plugins/project/vsxmake/render.lua
```diff
@@ -34,31 +34,106 @@ end
...
+        local old_target = opt.paramsprovider() or ""
```
waruqi, line 77: 还有这

xmake/plugins/project/vsxmake/render.lua
```diff
@@ -34,31 +34,106 @@ end
...
 
-        local k = tmp[1]:trim()
-        local v = tmp[2]:trim()
+        for i,item in ipairs(m) do
```
waruqi, line 78: `i, item` 空格

waruqi:
@SirLynix Can you help me review and test this patch for your projects?

SirLynix:
So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.

I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?

vsxmake generator doesn't need dependencies as they are handled by xmake already.

waruqi:
> So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> 
> I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> 
> vsxmake generator doesn't need dependencies as they are handled by xmake already.

You are right, xmake will ensure the order of dependencies

I will reconsider whether I need to merge it or not.

Domain:
> 



> So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> 
> I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> 
> vsxmake generator doesn't need dependencies as they are handled by xmake already.



> So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> 
> I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> 
> vsxmake generator doesn't need dependencies as they are handled by xmake already.

Are you sure the slowdown is caused by this patch?
My test slows down less than 1 second! The original vsxmake generated solution already slows down 15%. And this is not because of this patch. 
My solution has 28 projects, 10000+ cpp/h files. The total time used from different build methods:
Build from cli xmake: 10:03.28
Build from original vsxmake solution: 11:35.76
Build from this patch: 11:36.21

SirLynix:
Yes I am sure, I have a big project with 15-20 targets which benefit a lot
from xmake compilation of all .cpp and then parallel linking.

I went from 40s (xmake dev) to 1m30 (your branch)

SirLynix:
My question still stands though, why do you need this?

waruqi:
This patch is equivalent to the global configuration of `set_policy("build.cross_targets_in_parallel", false)`, which disables xmake's built-in cross-target parallel compilation optimizations.

Even further, all targets are forced to be linked serially.

waruqi:
> > 
> 
> > So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> > I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> > vsxmake generator doesn't need dependencies as they are handled by xmake already.
> 
> > So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> > I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> > vsxmake generator doesn't need dependencies as they are handled by xmake already.
> 
> Are you sure the slowdown is caused by this patch? My test slows down less than 1 second! The original vsxmake generated solution already slows down 15%. And this is not because of this patch. My solution has 28 projects, 10000+ cpp/h files. The total time used from different build methods: Build from cli xmake: 10:03.28 Build from original vsxmake solution: 11:35.76 Build from this patch: 11:36.21

If there are a lot of targets in the project and they are all dynamic library programs that need to be fully linked, then this performance difference will be very noticeable. However, if your current project is mostly static libraries, the impact is not significant as link archiving is very fast. Also some of the targets in your project are already configured with `set_policy("build.cross_targets_in_parallel", false)`

Domain:
> Yes I am sure, I have a big project with 15-20 targets which benefit a lot
> from xmake compilation of all .cpp and then parallel linking.
> 
> I went from 40s (xmake dev) to 1m30 (your branch)

40s is built from original vsxmake solution or from command line? In my test, this patch has no big slowdown than the original solution. 

I just want to make sure the projects has a right dependency in vs solution. If no significant slowdown than the original vsxmake solution,why not? Yes, you can make a dummy project to build the entire solution. That is what I do in fastbuild. But that is because fastbuild cannot control this dependency. Xmake can generate much more complicated project files. If someone wants fastest built time, always use command line. Use vscode open the folder, no need generated solution.

Domain:
> > > 
> > 
> > > So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> > > I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> > > vsxmake generator doesn't need dependencies as they are handled by xmake already.
> > 
> > > So yeah, it works but it slows down the compilation as hell as it tries to link before going to the next project, it's even slower than native Visual Studio because it can't build projects in parallel.
> > > I don't understand what's the point of this since it slows down building a lot and removes one advantage of using vsxmake generator. What does it aim to fix?
> > > vsxmake generator doesn't need dependencies as they are handled by xmake already.
> > 
> > Are you sure the slowdown is caused by this patch? My test slows down less than 1 second! The original vsxmake generated solution already slows down 15%. And this is not because of this patch. My solution has 28 projects, 10000+ cpp/h files. The total time used from different build methods: Build from cli xmake: 10:03.28 Build from original vsxmake solution: 11:35.76 Build from this patch: 11:36.21
> 
> If there are a lot of targets in the project and they are all dynamic library/binary programs that need to be fully linked, then this performance difference will be very noticeable. However, if your current project is mostly static libraries, the impact is not significant as link archiving is very fast. Also some of the targets in your project are already configured with `set_policy("build.cross_targets_in_parallel", false)`

I have 5 static libs and 23 binary. Only 6 targets have this policy (5 libs and one helper tools).

waruqi:
> I just want to make sure the projects has a right dependency in vs solution. 

vsxmake will call xmake to build, and xmake will ensure that the correct dependencies are used to compile. This is no different to compiling directly with cli.

In vs it just looks like the logs are messed up, in fact xmake will still ensure that all targets are compiled in the correct order.

In theory, this patch is better suited to the vs gererator than to the vsxmake generator

> If someone wants fastest built time, always use command line. Use vscode open the folder, no need generated solution.

The patch itself should not break compilation performance, even in vs projects, which can lead to other more user feedback problems. Not all users will accept that compilation in vs is several times slower than before.

waruqi:
> Build from cli xmake: 10:03.28
Build from original vsxmake solution: 11:35.76
Build from this patch: 11:36.21

If the compilation error is caused by calling xmake in the vs project and not compiling in the correct order, then it should be a bug in xmake itself, and you can raise the bug issue separately and I will investigate the cause.

But from what I can see here, it looks like you've successfully compiled in the vs project, and haven't encountered any compilation errors due to incorrect compilation order.

Domain:
> But from what I can see here, it looks like your project has been successfully compiled in the vs project, and haven't encountered any compilation errors due to incorrect compilation order.

That's because I run the helper tools manually before the build.

waruqi:
> > But from what I can see here, it looks like your project has been successfully compiled in the vs project, and haven't encountered any compilation errors due to incorrect compilation order.
> 
> That's because I run the helper tools manually before the build.

So, if the compilation error is due to a compilation order problem, you should open a new issue and provide a reproducible demo project. I'll have a look at it, but for now this patch will break a lot of extra stuff.

Domain:
> > > But from what I can see here, it looks like your project has been successfully compiled in the vs project, and haven't encountered any compilation errors due to incorrect compilation order.
> > 
> > That's because I run the helper tools manually before the build.
> 
> So, if the compilation error is due to a compilation order problem, you should open a new issue and provide a reproducible demo project and errors. I'll have a look at it, but for now this patch will break a lot of extra stuff.

I just want to make sure the slowdown is caused by this patch or vsxmake itself. I don't have significant slowdown in my test. 
vsxmake is slow, that's because vs has to spawn many xmake to build each project (at least 2 xmakes for each one. One for config, one for build. I want single command to config and build!) 
And the dummy project mentions above is not a solution now. If you want to build that dummy project when you build the solution, you have to set it's default to true, and all other projects to false. That means you cannot build all targets in cli too! You cannot specify multiple targets in xmake now.

waruqi:
> vsxmake is slow, that's because vs has to spawn many xmake to build each project

vsxmake is not slow. @SirLynix's tests prove this.

> I went from 40s (xmake dev) 

although vs will spawn many xmake build processes, xmake will ensure that only one process completes the compilation of the target and its deps at the same time by project file lock.

After that, the other xmake processes will not recompile them, but simply exit immediately, unless they encounter targets that have not been compiled yet.

for example:

```
- root target 
  - dep1
  - dep2 
    - dep3 
    - dep4 
```

case1:

vs will open 5 build processes at once, and if the root target gets the lock first, the others will simply exit quickly afterwards and do nothing.

```console
xmake build root --> get project lock, xmake will build all targets, only one build process
xmake build dep1  --> wait lock and fast exit later, do nothing
xmake build dep2  --> wait lock and fast exit later, do nothing
xmake build dep3  --> wait lock and fast exit later, do nothing
xmake build dep4  --> wait lock and fast exit later, do nothing
```

case2:

If one of its dependencies has priority to get a lock, then after xmake has compiled all of its dependencies in parallel, no other process will compile them either.

```console
xmake build dep2  --> get project lock, and build dep2, dep3, dep4 by parallel
xmake build dep1  --> wait lock and build dep1 later
xmake build dep3  --> wait lock and fast exit later, do nothing
xmake build root --> get project lock, and only build root target later
xmake build dep4  --> wait lock and fast exit later, do nothing
```

Of course, if you really want to optimize it with this patch, you should reverse the target dependency graph to ensure that the root target, as a child node, is called by vs first. That is, always leave vsxmake in the optimal case of case1.

However, case2 doesn't really cause any performance or sequencing issues at the moment.


> (at least 2 xmakes for each one. One for config, one for build. 

The xmake config process is not involved in compiling, it usually ends very quickly and its time consumption can be ignored compared to xmake build.

> I want single command to config and build!)

Even this patch does not guarantee that there will be only one build process at a time; vs will spawn multiple build processes for targets without any dependencies.

Domain:
From msbuild diagnostics logs, each xmake spawn takes seconds! So a solution with 30 projects will takes at least 30 seconds to build ( all up-to-date ). From cli, it takes less than 10 seconds. For comparing, fastbuild takes less than 2 seconds.

waruqi:
> From msbuild diagnostics logs, each xmake spawn takes seconds! So a solution with 30 projects will takes at least 30 seconds to build ( all up-to-date ).

This is due to the mechanics of the vsxmake generator, you can't avoid multiple xmake processes running, even with this patch.

The windows process is inherently heavy, and the lua script takes time to load, even if it does nothing.

If you want to use a single process, you should use vs generator instead of vsxmake generator

> From cli, it takes less than 10 seconds. For comparing, fastbuild takes less than 2 seconds.

This has nothing to do with this pr, if you feel that xmake cli is slow you can open separate issues and suggest optimisations that can be made.

Also, lua scripts have some inherent performance loss compared to c++.

SirLynix:
Sorry for the lack of reply, different timezones.

## Test

So, how did I test your PR, here's what I did:

**Your PR**
- `xmake update -s github:Domain/xmake#sln_dep`
- `xmake project -k vsxmake`
- `xmake clean`
- Select one of my example target which basically depends on every other lib, right click => build.
- Measure time
- Project A builds => xmake config => xmake build => build ok
- Project B builds => xmake config => xmake build => build ok
- Project C builds => xmake config => xmake build => build ok
- ...
- build finished in 70s (it was 20s faster than yesterday because I built another target)

CPU usage:
image

(we can see it pauses a lot)

**vsxmake**

- `xmake update -s dev`
- `xmake project -k vsxmake`
- `xmake clean`
- Select the same project as before, right click => build.
- Measure time
- Project builds => xmake config => xmake build => build ok
- build finished in 40s (lost a bit time doing config)

CPU usage:
image

(only one pause due to one of my rule)

**xmake**

- Still on xmake dev
- `xmake clean`
- `time xmake build TargetName`
```
real    0m35.620s
user    0m0.000s
sys     0m0.054s
```

CPU usage:
image

(same)

## Conclusions

vsxmake is basically a front-end to xmake, there should be no difference between building from cli and building from VS.

Adding project dependencies in vsxmake solutions has no purpose since xmake handles dependencies, and slow downs building a lot due to performing like `"build.cross_targets_in_parallel"` policy to false, basically (which is VS default behavior too) + the two commands vsxmake starts (xmake config and xmake build) for every dependency.

## vsxmake is not perfect

However there are issues which should be fixed.

Here, I only built one projets which has many dependencies, but I didn't build the whole solution (which is very slow to do with vsxmake).

One solution would be to add a fake project "AllBuild" and remove every other project from the configuration manager. This way when executing the "build solution" command, only this project would get build (which would run "xmake build" for all targets). Building each project with right click or "ctrl + B" still works.

I tested it manually with success.

See https://stackoverflow.com/questions/1433657/visual-studio-how-to-build-only-some-projects
(I'm not sure how this translates to .sln)

> vsxmake is slow, that's because vs has to spawn many xmake to build each project (at least 2 xmakes for each one. One for config, one for build. I want single command to config and build!)

This would fix this issue.
However I agree xmake would benefit from having a "config, build and forget" command which would configure the project, built it and then revert to the previous configuration. It would be very useful here, and would be useful for testing in cli too.

> And the dummy project mentions above is not a solution now. If you want to build that dummy project when you build the solution, you have to set it's default to true, and all other projects to false. That means you cannot build all targets in cli either! You cannot specify multiple targets in xmake now.

I'm not sure to understand, simply run `xmake build`?

Domain:
Each time, I do a fresh clone, and build the entire solution. There is no significant difference build time between my patch(11:36) and the original vsxmake (11:35). 
And I try to use `xmake clean` and build the solution again in then same repos:
1. My patch: 2:33
2. Original vsxmake: 2:36
3. cli xmake: 11:32
The first 2 still have no significant difference. What surprise me is the last one. I am not sure why. It looks like the first 2 use local cache, but not the last one. So I test it again.
1. My patch: 2:32
2. Original vsxmake: 2:45
3. cli xmake: 2:10
My patch still faster than the original one. This time all use local cache.

> I'm not sure to understand, simply run xmake build?

I mean you cannot specify one dummy project to build all other targets now. See #2432 #2433

SirLynix:
How many subdependencies do you have?

With your patch my project needs to wait until six libraries are individually built before compiling, there's no way it won't slow down your build.

(Please note that I'm not using the 
"build.cross_targets_in_parallel" policy which may be the difference here).

About the BuildAll dummy target, I wasn't talking about adding it yourself but to improve vsxmake to add it automatically

Domain:
> How many subdependencies do you have?
> 
> With your patch my project needs to wait until six libraries are individually built before compiling, there's no way it won't slow down your build.
> 
> (Please note that I'm not using the "build.cross_targets_in_parallel" policy which may be the difference here).
> 
> About the BuildAll dummy target, I wasn't talking about adding it yourself but to improve vsxmake to add it automatically

I have 5 libs, 1 builder tool, 1 dummy project to run the builder, and 21 apps. All these apps are depended on the builder and 3 ~ 5 libs. All libs and builder have  "build.cross_targets_in_parallel" policy set to false. And VS already support parallel build. I can see all the libs are built at the same time, then the builder, the run the builder to generate source files, and finally 20 apps are built together.

SirLynix:
> And VS already support parallel build. I can see all the libs are built at the same time, then the builder, the run the builder to generate source files, and finally 20 apps are built together. That is exactly what I wanted.

I'm sorry but are you sure to be using vsxmake generator? vs generator can built targets in parallel but vsxmake can't because xmake has a lock on the project, only one xmake build can run at the same time.

So while VS will indeed start `xmake build TargetA` and `xmake build TargetB` in parallel, the first xmake process will block the second, which will have to wait. This is what I observe.

> All libs and builder have "build.cross_targets_in_parallel" policy set to false.

This is why you don't have a lot of difference with this PR, it does massively increase the compilation speed when you have a lot of libraries.

Also, could you please explain what issue does this PR fixes? Because xmake handles dependencies order so there should be no difference between running vsxmake and running xmake from a cli. And if there's an issue with dependencies order that can't be solved with a policy, then it's a xmake issue, not a vsxmake one.

I repeat but so far this PR doesn't fix anything, it prevents xmake to compile as fast as possible.

Domain:
> So while VS will indeed start `xmake build TargetA` and `xmake build TargetB` in parallel, the first xmake process will block the second, which will have to wait. This is what I observe.

I don't know the detail. If that was true, this PR change nothing! VS always spawn xmake for each project, then they are all blocked. All projects build in linear sequence! If you want real parallel build, you must use cli.

> Also, could you please explain what issue does this PR fixes? Because xmake handles dependencies order so there should be no difference between running vsxmake and running xmake from a cli. And if there's an issue with dependencies order that can't be solved with a policy, then it's a xmake issue, not a vsxmake one.
> 
> I repeat but so far this PR doesn't fix anything, it prevents xmake to compile as fast as possible.

In a word, this PR can generate a right solution file to build my projects.
I am tired to explain, feel free to close this.

SirLynix:
Here's the build log with this PR: https://justpaste.it/9m23w

I can confirm that while VS does trigger multiple `xmake build`, only one library is built at the same time, which isn't the case with xmake dev.

For example:
```
5>------ Début de la génération : Projet : NazaraPlatform, Configuration : debug x64 ------
6>------ Début de la génération : Projet : NazaraPhysics3D, Configuration : debug x64 ------
7>------ Début de la génération : Projet : NazaraPhysics2D, Configuration : debug x64 ------
7>$xmake config -y  -P . -p windows -m debug -a x64 -o "build"
5>$xmake config -y  -P . -p windows -m debug -a x64 -o "build"
6>$xmake config -y  -P . -p windows -m debug -a x64 -o "build"
5>$xmake build -y  -P . -w "NazaraPlatform"
7>$xmake build -y  -P . -w "NazaraPhysics2D"
6>$xmake build -y  -P . -w "NazaraPhysics3D"

...
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Arbiter2D.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Collider2D.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Components\RigidBody2DComponent.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Constraint2D.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Debug\NewOverload.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Physics2D.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\PhysWorld2D.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\RigidBody2D.cpp
7>[ 87%]: compiling.debug src\Nazara\Physics2D\Systems\Physics2DSystem.cpp
7>C:\Projets\Perso\NazaraNext\NazaraEngine\src\Nazara\Physics2D\RigidBody2D.cpp(161): warning C4701: variable locale 'closest' potentiellement non initialisée utilisée
7>[ 98%]: linking.debug NazaraPhysics2D-d.dll
7>[100%]: build ok!
7>Génération du projet "NazaraPhysics2D.vcxproj" terminée.
6>[ 89%]: compiling.debug src\Nazara\Physics3D\Collider3D.cpp
6>[ 89%]: compiling.debug src\Nazara\Physics3D\Components\RigidBody3DComponent.cpp
6>[ 89%]: compiling.debug src\Nazara\Physics3D\Debug\NewOverload.cpp
6>[ 89%]: compiling.debug src\Nazara\Physics3D\Physics3D.cpp
6>[ 89%]: compiling.debug src\Nazara\Physics3D\PhysWorld3D.cpp
6>[ 89%]: compiling.debug src\Nazara\Physics3D\RigidBody3D.cpp
6>[ 89%]: compiling.debug src\Nazara\Physics3D\Systems\Physics3DSystem.cpp
6>[ 98%]: linking.debug NazaraPhysics3D-d.dll
6>[100%]: build ok!
```

Same thing with xmake dev:

Full log: https://justpaste.it/2xmlw

```
1>$xmake config -y  -P . -p windows -m debug -a x64 -o "build"
1>$xmake build -y  -P . -w "Tut00_EmptyProject"
1>[  3%]: compiling.debug src\Nazara\Audio\Audio.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\Debug\NewOverload.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\VulkanTexture.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\VulkanDevice.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\OpenALDevice.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\Sound.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\VulkanRenderPipeline.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\AudioSource.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\SoundBuffer.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\DummyAudioSource.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\VulkanRenderWindow.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\Debug\NewOverload.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\Wrapper\CommandPool.cpp
1>[  3%]: compiling.debug src\Nazara\VulkanRenderer\VulkanUploadPool.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\Music.cpp
1>[  3%]: compiling.debug src\Nazara\Audio\OpenALSource.cpp
```

only one xmake build issued, and all .cpp are compiled in parallel.

> I don't know the detail. If that was true, this PR change nothing! VS always spawn xmake for each project, then they are all blocked. All projects build in linear sequence! If you want real parallel build, you must use cli.

It does change the behavior because you added dependencies, which force VS to spawn one xmake build **per project**. Without this PR, VS is not aware of target dependencies and will issue only one process which will be able to greatly improve the compilation speed.

Do you understand the difference now?

> In a word, this PR can generate a right solution file to build my projects.

vsxmake is only a front-end to xmake, the issue should be no different in xmake CLI, and if it is we should try to understand why it's different.

waruqi:
First of all, thank you very much for your contribution, but this pr really doesn't solve anything at the moment and there's no point in continuing the discussion.

I'm sorry I'll have to close it for now, maybe vs generator is a better fit for it.

Thanks to @SirLynix  for helping me to better explain the mechanics of vsxmake generator.

SirLynix:
Maybe I'm missing something and this PR does fix something beside adding project dependencies? If we can remove the project dependencies part and have a fix, everyone would be happy.

PR #2553, opened 2022-07-10
Add support for cxxlatest (msvc) with CMakeLists generation

fifty-six:
When using `set_languages("cxxlatest")` for MSVC on windows, this gives us the `/std:c++latest` flag. However, this wasn't put in the CMakeLists file so it would fail to use newer C++ features and be unable to build. This just adds support for it.


waruqi:
Thanks

PR #2582, opened 2022-07-19
cmake: add support of include_external_headers policy

SirLynix:
See https://cmake.org/cmake/help/latest/variable/CMAKE_NO_SYSTEM_FROM_IMPORTED.html#variable:CMAKE_NO_SYSTEM_FROM_IMPORTED


xmake/modules/package/tools/cmake.lua
```diff
@@ -275,6 +275,9 @@ function _get_configs_for_generic(package, configs, opt)
     if package:config("pic") ~= false then
         table.insert(configs, "-DCMAKE_POSITION_INDEPENDENT_CODE=ON")
     end
+    if package:policy("package.include_external_headers") == false then
```
waruqi, line 278: please rebase to dev and use `if not package:use_external_includes() then`

PR #2627, opened 2022-07-31
Improve Qt detection on Linux

Arthapz:
When Qt6 and Qt5 are installed on archlinux, Qt6 is installed in /usr/{lib,include}/qt6 (with qmake6 executable instead of qmake)
wich prevent XMake to detect it


Arthapz:
that pull request wasn't usefull on second through

PR #2644, opened 2022-08-05
Improve cxx20 modules support

Arthapz:


xmake/rules/c++/modules/modules_support/gcc.lua
```diff
@@ -149,77 +149,91 @@ function generate_dependencies(target, sourcebatch, opt)
...
-            end
+        local args = { "-c" }
+        local headerunit_path
+        if headerunit.type == ":quote" then
+            table.join2(args, { "-I", path.directory(path.relative(headerunit.path, projectdir)), "-x", "c++-user-header", headerunit.name })
+            headerunit_path = path.join(".", path.relative(headerunit.path, projectdir))
+        elseif headerunit.type == ":angle" then
+            table.join2(args, { "-x", "c++-system-header", headerunit.name })
+            -- if path is relative then its a subtarget path
+            headerunit_path = path.is_absolute(headerunit.path) and headerunit.path or path.join(".", headerunit.path)
```
waruqi, line 214: > path.join(".", headerunit.path)

why? I think it is exactly equivalent to `headerunit_path  = headerunit_path `

xmake/rules/c++/modules/modules_support/gcc.lua
```diff
@@ -149,77 +149,91 @@ function generate_dependencies(target, sourcebatch, opt)
...
+            batchcmds:show_progress(opt.progress, "${color.build.object}generating.cxx.headerunit.bmi %s", headerunit.name)
+            batchcmds:vrunv(compinst:program(), table.join(compinst:compflags({target = target}), args))
 
-            local bmifilename = path.basename(objectfile) .. get_bmi_extension()
-            local bmifile = (outdir and path.join(outdir, bmifilename) or bmifilename)
-            if not os.isdir(path.directory(objectfile)) then
-                os.mkdir(path.directory(objectfile))
-            end
+            batchcmds:set_depmtime(os.mtime(bmifile))
+            batchcmds:set_depcache(target:dependfile(bmifile))
```
waruqi, line 171: We should not set it multiple times in the loop

xmake/rules/c++/modules/modules_support/gcc.lua
```diff
@@ -149,77 +149,91 @@ function generate_dependencies(target, sourcebatch, opt)
...
+            headerunit_path = path.is_absolute(headerunit.path) and headerunit.path or path.join(".", headerunit.path)
+        end
+
+        if _add_module_to_mapper(mapper_file, headerunit_path, path.absolute(bmifile, projectdir)) then
+            batchcmds:show_progress(opt.progress, "${color.build.object}generating.cxx.headerunit.bmi %s", headerunit.name)
+            batchcmds:vrunv(compinst:program(), table.join(compinst:compflags({target = target}), args))
+
+            batchcmds:add_depfiles(headerunit.path)
+            batchcmds:set_depmtime(os.mtime(bmifile))
+            batchcmds:set_depcache(target:dependfile(bmifile))
```
waruqi, line 223: We should not set it multiple times in the loop

xmake/rules/c++/modules/modules_support/gcc.lua
```diff
@@ -149,77 +149,91 @@ function generate_dependencies(target, sourcebatch, opt)
...
+            os.mkdir(outdir)
+        end
 
-                    batchcmds:show_progress(opt.progress, "${color.build.object}generating.cxx.headerunit.bmi %s", headerunit.name)
-                    batchcmds:vrunv(compinst:program(), table.join(compinst:compflags({target = target}), args))
+        local bmifilename = path.basename(objectfile) .. get_bmi_extension()
+        local bmifile = (outdir and path.join(outdir, bmifilename) or bmifilename)
+        if not os.isdir(path.directory(objectfile)) then
+            os.mkdir(path.directory(objectfile))
+        end
```
waruqi, line 204: we can only use `batchcmds:mkdir(path.directory(objectfile))`

xmake/rules/c++/modules/modules_support/msvc.lua
```diff
@@ -123,62 +123,84 @@ function generate_headerunits(target, batchcmds, headerunits, opt)
...
+            outputdir = path.join(cachedir, path.directory(headerunit.path):sub(3))
+        end
+        if not os.isdir(outputdir) then
+            os.mkdir(outputdir)
         end
+
+        local bmifilename = path.basename(objectfile) .. get_bmi_extension()
+        local bmifile = (outputdir and path.join(outputdir, bmifilename) or bmifilename)
+        if not os.isdir(path.directory(objectfile)) then
+            os.mkdir(path.directory(objectfile))
```
waruqi, line 186: use `batchcmds:mkdir`

xmake/rules/c++/modules/modules_support/gcc.lua
Arthapz: > 

because gcc prepend paths with . when it's relative, and without it it doesn't work on my tests, i think GCC do a strict string comparison and not a resolved path comparison

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -185,13 +185,13 @@ function generate_user_headerunits(target, batchcmds, headerunits, opt)
             outdir = path.join(cachedir, path.directory(headerunit.path))
         end
         if not os.isdir(outdir) then
-            os.mkdir(outdir)
+            batchcmds:mkdir(outdir)
```
waruqi, line 188: we can remove `if not os.isdir(outdir) then`, need only batchcmds:mkdir, it will handle it.

```lua
function _runcmd_mkdir(cmd, opt)
    local dir = cmd.dir
    if not opt.dryrun and not os.isdir(dir) then
        os.mkdir(dir)
    end
end
```

xmake/rules/c++/modules/modules_support/gcc.lua
Arthapz: so i can't use set_depmtime/set_depcache because there is a lot of bmifile ?

waruqi: yes, they should be set once at the end

```lua
    before_buildcmd_files(function(target, batchcmds, sourcebatch, opt)
         for loop
               batchcmds:vrunv()
               batchcmds:vrunv()
         end
         batchcmds:set_depmtime(math.max(all_bmi_files_mtimes))
         batchcmds:set_depcache(target:dependfile("all_bmi_files ..."))
    end)
```

Unless we use `before_buildcmd_file` and build the bmi file separately for each sourcefile, then we set mtime/depcache for each bmifile, but there will be parallel compilation problems.

This is because each file job is executed in parallel.

waruqi: So, while before_buildcmd_files are usually used for project generators and simple tasks, before_build_files can handle tasks in a more flexible way.

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -149,11 +151,12 @@ function generate_stl_headerunits(target, batchcmds, headerunits, opt)
             batchcmds:vrunv(compinst:program(), table.join(compinst:compflags({target = target}), args))
         end
 
-        batchcmds:set_depmtime(os.mtime(bmifile))
-        batchcmds:set_depcache(target:dependfile(bmifile))
-
         table.insert(flags, modulefileflag .. bmifile)
+        table.append(bmifiles, bmifile)
```
waruqi, line 155: please use table.insert

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -149,11 +151,12 @@ function generate_stl_headerunits(target, batchcmds, headerunits, opt)
...
+        depmtime = math.max(depmtime, os.mtime(bmifile))
     end
+    batchcmds:set_depmtime(depmtime)
+    batchcmds:set_depcache(target:dependfile(bmifiles))
```
waruqi, line 159: target:dependfile can only accept string

Arthapz: how should i format this string ? juste append all paths ?

waruqi: or you can remove `batchcmds:set_depcache`, it will use tmpfile as depcache automatically

waruqi:
```lua
function batchcmds:set_depcache(cachefile)
    local deps = self._DEPS or {}
    deps.dependfile = cachefile
    self._DEPS = deps
end
```

```lua
    -- get dependfile
    local dependfile = opt.dependfile
    if not dependfile then
        dependfile = project.tmpfile(table.concat(table.wrap(opt.files), ""))
    end
```

It will use `batchcmds:add_depfiles` to calculate a tmpfile as depcache file path.

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -215,44 +236,44 @@ function build_modules(target, batchcmds, objectfiles, modules, opt)
...
 
             local args = { emitmoduleinterfaceflag }
             local bmiflags = {}
-            local bmifiles = {}
+            local bmifiles_ = {}
             for name, provide in pairs(m.provides) do
                 batchcmds:show_progress(opt.progress, "${color.build.object}generating.cxx.module.bmi %s", name)
 
                 local bmifile = provide.bmi
                 table.join2(args, { "-c", "-x", "c++-module", "--precompile", provide.sourcefile, "-o", bmifile })
```
waruqi, line 255: If there are multiple provides, there seems to be some problems here.

`batchcmds:vrunv` will run `-c sourcefile1 -o bmifile1 -c sourcefile2 -o bmifile2 ...`?

Arthapz: I didn't explored this part so much, and i don't know how we should handle multiple provided modules, i didn't encounter a case where provides had more than one module

waruqi: Let's assume for a moment that provides is only one? If we encounter more than one later, we can improve it.

we can add `assert(#provides == 1, "multiple provides are not supported now!")`

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -235,33 +235,30 @@ function build_modules(target, batchcmds, objectfiles, modules, opt)
...
-
-                batchcmds:add_depfiles(provide.sourcefile)
-
-                table.join2(bmiflags, {modulefileflag .. bmifile})
-                depmtime = math.max(depmtime, os.mtime(bmifile))
+        if m and m.provides then
+            -- assume there that provides is only one, until we encounter the case
+            local length = 0
+            for _, _2 in pairs(m.provides) do
+                length = length + 1
```
waruqi, line 242:
```
if length > 1 then
    break
end
```

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -235,33 +235,30 @@ function build_modules(target, batchcmds, objectfiles, modules, opt)
...
             end
+            assert(length == 1, "multiple provides are not supported now!")
+            local name, provide = pairs(m.provides)(m.provides)
```
waruqi, line 245:
```lua
local name, provide
local length = 0
for k, v in pairs(m.provides) do
      length = length + 1
      if length == 1 then
           name = k
           provide = v
           break
      elif length > 1 then
           raise("multiple provides are not supported now!")
      end
end
```

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -215,44 +230,42 @@ function build_modules(target, batchcmds, objectfiles, modules, opt)
...
-            batchcmds:set_depcache(target:dependfile(objectfile))
+            batchcmds:add_depfiles(provide.sourcefile)
 
+            local bmiflags = modulefileflag .. bmifile
             target:add("cxxflags", bmiflags, {public = true, force = true})
             target:add("objectfiles", objectfile)
-            for _, f in ipairs(bmiflags) do
-                target:data_add("cxx.modules.flags", f)
-            end
+            target:data_add("cxx.modules.flags", f)
```
waruqi, line 264: where is the `f` var?

xmake/rules/c++/modules/modules_support/gcc.lua
waruqi: ok, got it.

waruqi:
Is it ready? I am going to continue to improve it based on this patch.

PR #2668, opened 2022-08-11
In project.clang, replace -isystem by -I only if it starts with.

glcraft:
I wasn't able to use the clang flag `-stdlib++-isystem` and make a compile_commands.json with xmake because during the flags translation, it replaces all `-isystem` by `-I` (so i had `-stdlib++-I` in the compile_commands.json).

Now, it replaces the flags when it starts with `-isystem`

I also prevent code if `-isystem-after` flag is used or if `-isystem` if present in the path following `-isystem` flag.

### DIFF
--- a/xmake/plugins/project/clang/compile_commands.lua
+++ b/xmake/plugins/project/clang/compile_commands.lua
@@ -45,9 +45,9 @@ function _translate_arguments(arguments)
             arg = arg .. ".exe"
         end
         if arg:startswith("-isystem-after", 1, true) then
-            arg = "-I" .. arg:sub(15, -1)
+            arg = "-I" .. arg:sub(15)
         elseif arg:startswith("-isystem", 1, true) then
-            arg = "-I" .. arg:sub(9, -1)
+            arg = "-I" .. arg:sub(9)
         elseif arg:find("[%-/]external:I") then
             arg = arg:gsub("[%-/]external:I", "-I")
         elseif arg:find("[%-/]external:W") or arg:find("[%-/]experimental:external") then
--- a/xmake/plugins/project/clang/compile_commands.lua
+++ b/xmake/plugins/project/clang/compile_commands.lua
@@ -44,7 +44,7 @@ function _translate_arguments(arguments)
         if idx == 1 and is_host("windows") and path.extension(arg) == "" then
             arg = arg .. ".exe"
         end
-        if arg:find("-isystem", 1, true) then
+        if arg:startswith("-isystem", 1, true) then
             arg = arg:replace("-isystem", "-I")
         elseif arg:find("[%-/]external:I") then
             arg = arg:gsub("[%-/]external:I", "-I")
--- a/xmake/plugins/project/clang/compile_commands.lua
+++ b/xmake/plugins/project/clang/compile_commands.lua
@@ -45,7 +45,7 @@ function _translate_arguments(arguments)
             arg = arg .. ".exe"
         end
         if arg:startswith("-isystem-after", 1, true) then
-            arg = "-I" .. arg:sub(13, -1)
+            arg = "-I" .. arg:sub(15, -1)
         elseif arg:startswith("-isystem", 1, true) then
             arg = "-I" .. arg:sub(9, -1)
         elseif arg:find("[%-/]external:I") then
--- a/xmake/plugins/project/clang/compile_commands.lua
+++ b/xmake/plugins/project/clang/compile_commands.lua
@@ -44,8 +44,10 @@ function _translate_arguments(arguments)
         if idx == 1 and is_host("windows") and path.extension(arg) == "" then
             arg = arg .. ".exe"
         end
-        if arg:startswith("-isystem", 1, true) then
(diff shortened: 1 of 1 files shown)


xmake/plugins/project/clang/compile_commands.lua
```diff
@@ -44,8 +44,10 @@ function _translate_arguments(arguments)
         if idx == 1 and is_host("windows") and path.extension(arg) == "" then
             arg = arg .. ".exe"
         end
-        if arg:find("-isystem", 1, true) then
-            arg = arg:replace("-isystem", "-I")
+        if arg:startswith("-isystem-after", 1, true) then
+            arg = "-I" .. arg:sub(13, -1)
+        elseif arg:startswith("-isystem", 1, true) then
+            arg = "-I" .. arg:sub(9, -1)
```
waruqi, line 50: It will break `-isystem /xxxx` => `-I /xxxx`

xmake/plugins/project/clang/compile_commands.lua
```diff
@@ -44,8 +44,10 @@ function _translate_arguments(arguments)
         if idx == 1 and is_host("windows") and path.extension(arg) == "" then
             arg = arg .. ".exe"
         end
-        if arg:find("-isystem", 1, true) then
-            arg = arg:replace("-isystem", "-I")
+        if arg:startswith("-isystem-after", 1, true) then
+            arg = "-I" .. arg:sub(13, -1)
```
waruqi, line 48: why? startswith? len is 12?  `#("-isystem-after")` is 14

xmake/plugins/project/clang/compile_commands.lua
```diff
@@ -44,8 +44,10 @@ function _translate_arguments(arguments)
         if idx == 1 and is_host("windows") and path.extension(arg) == "" then
             arg = arg .. ".exe"
         end
-        if arg:find("-isystem", 1, true) then
```
waruqi, line 47: To fix the problem you describe, simply change this

```lua
        if arg == "-isystem" then
```

xmake/plugins/project/clang/compile_commands.lua
glcraft: It should be okay : 
if the argument is `-isystem`, it start with `-isystem` indeed, and arg:sub will return an empty string.

We can prevent that by adding 
```lua
if arg == "-isystem" or arg == "-isystem-after" then
    arg = "-I"
elseif ...
```

xmake/plugins/project/clang/compile_commands.lua
glcraft: Oh, my mistake you right !

xmake/plugins/project/clang/compile_commands.lua
glcraft: Clang accept arguments like `-isystem<directory>`. Same for  `-isystem-after<directory>`. arg:startswith handles those cases

xmake/plugins/project/clang/compile_commands.lua
glcraft: Resolved in [91a84f7](https://github.com/xmake-io/xmake/pull/2668/commits/91a84f7536e1563b448e85e2c4dacc70bcdd7a82)

xmake/plugins/project/clang/compile_commands.lua
```diff
@@ -44,8 +44,10 @@ function _translate_arguments(arguments)
...
-            arg = arg:replace("-isystem", "-I")
+        if arg:startswith("-isystem-after", 1, true) then
+            arg = "-I" .. arg:sub(15, -1)
```
waruqi, line 48: we just write `arg:sub(15)`

xmake/plugins/project/clang/compile_commands.lua
```diff
@@ -44,8 +44,10 @@ function _translate_arguments(arguments)
...
+        elseif arg:startswith("-isystem", 1, true) then
+            arg = "-I" .. arg:sub(9, -1)
```
waruqi, line 50: `arg:sub(9)`

xmake/plugins/project/clang/compile_commands.lua
glcraft: Oh... Yes it is.
I tried that before, I had an error. I think I wrote `arg.sub` instead of `arg:sub` when I got the error and then use `arg:sub` when I commit. Let me correct that 👍

glcraft: Resolved at [c7d932e](https://github.com/xmake-io/xmake/pull/2668/commits/c7d932ea12a0fa61951c1fd9f6d87e2674aa0569)

xmake/plugins/project/clang/compile_commands.lua
glcraft: Resolved at Resolved at [c7d932e](https://github.com/xmake-io/xmake/pull/2668/commits/c7d932ea12a0fa61951c1fd9f6d87e2674aa0569)

waruqi:
thanks

PR #2674, opened 2022-08-13
Call cygpath once at the beginning for msys2 pacman packages

ImperatorS79:
In the case of a pacman package under msys2 without .pc files, `cygpath` is called multiple times to obtain a good path. Calling `cygpath` is slow, and for a big package without .pc files like `boost`, it takes too much time to parse the package content.

With this change, `cygpath` is called once to obtain the correct msys2 base path, and then `path.join` is used. Reading a package like `boost` is now very fast.

### DIFF
--- a/xmake/modules/package/manager/pacman/find_package.lua
+++ b/xmake/modules/package/manager/pacman/find_package.lua
@@ -29,16 +29,14 @@ function _find_package_from_list(list, name, pacman, opt)
 
     -- mingw + pacman = cygpath available
     local cygpath = nil
-    local pathtomsys2 = ""
+    local pathtomsys2 = nil
     if is_subhost("msys") and opt.plat == "mingw" then
         cygpath = find_tool("cygpath")
         if not cygpath then
             return
         end
         pathtomsys = os.iorunv(cygpath.program, {"--windows", "/"})
-        if pathtomsys:endswith("\n") then
-            pathtomsys = pathtomsys:sub(1, -2)
-        end
+        pathtomsys = pathtomsys:trim()
     end
 
     -- iterate over each file path inside the pacman package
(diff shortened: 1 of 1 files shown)


xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -29,11 +29,16 @@ function _find_package_from_list(list, name, pacman, opt)
 
     -- mingw + pacman = cygpath available
     local cygpath = nil
+    local pathtomsys2 = ""
     if is_subhost("msys") and opt.plat == "mingw" then
         cygpath = find_tool("cygpath")
         if not cygpath then
             return
         end
+        pathtomsys = os.iorunv(cygpath.program, {"--windows", "/"})
```
waruqi, line 38: pathtomsys = pathtomsys:trim()

xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -29,11 +29,16 @@ function _find_package_from_list(list, name, pacman, opt)
 
     -- mingw + pacman = cygpath available
     local cygpath = nil
+    local pathtomsys2 = ""
```
waruqi, line 32: pathtomsys2 = nil

xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -43,22 +48,21 @@ function _find_package_from_list(list, name, pacman, opt)
...
             if not line:startswith("/usr/include/") then
                 local hpath = line
-                if is_subhost("msys") and opt.plat == "mingw" then
-                    hpath = os.iorunv(cygpath.program, {"--windows", line})
-
+                if is_subhost("msys") and opt.plat == "mingw" then 
+                    hpath = path.join(pathtomsys, line)
                     if opt.arch == "x86_64" then
-                        local basehpath = os.iorunv(cygpath.program, {"--windows", "/mingw64/include"})
+                        local basehpath = path.join(pathtomsys, "/mingw64/include")
```
waruqi, line 54: local basehpath = path.join(pathtomsys,  "mingw64/include")

xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -91,7 +95,7 @@ function _find_package_from_list(list, name, pacman, opt)
         result.version = version:split('-')[1]
     else
         result = nil
-    end
+    end
```
waruqi, line 98: remove tailing spaces

xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -29,16 +29,14 @@ function _find_package_from_list(list, name, pacman, opt)
 
     -- mingw + pacman = cygpath available
     local cygpath = nil
-    local pathtomsys2 = ""
+    local pathtomsys2 = nil
```
waruqi, line 32: pathtomsys, not pathtomsys2

PR #2680, opened 2022-08-16
Improve MSVC module code

Arthapz:


xmake/rules/c++/modules/modules_support/msvc.lua
```diff
@@ -151,11 +151,38 @@ function generate_dependencies(target, sourcebatch, opt)
...
+
+        common.memcache():set2(name, "generating", true)
+        progress.show((index * 100) / total, "${color.build.object}generating.cxx.headerunit.bmi %s", name)
+        os.vrunv(compinst:program(), table.join(compinst:compflags({target = target}), common_flags, flags), {envs = vcvars})
+    end
+end
+
+-- generate header unit module bmi for batchcmds
+function generate_headerunit_for_batchcmds(target, name, flags, batchcmds)
+    -- don't generate same header unit bmi at the same time across targets
```
waruqi, line 171: We don't need to deal with batchcmds, they are just a list of commands, we haven't actually run them yet. It is not accurate to determine the status here, and we don't need to do any additional processing, as batchcmds are usually used only for the project generator.

Arthapz: it's juste to factorize a little (avoid repetition between generate_stl_header_unit and generate_user_header_unit

waruqi: only for batchjob, we need only generate_headerunit_for_batchjob

because batchcmds:vrunv is not actually executed yet, it just adds commands to the list

waruqi: For batchcmds, there is no point in maintaining the state of generating, which is always set to true before all batchcmds are executed.

For batchjobs only, it is only valid to use the generating state to avoid headerunits being generated repeatedly.

Arthapz: If i don't write it, i'm getting double generation errors

```
[ 37%]: generating.cxx.headerunit.bmi iostream
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64\cl.exe /EHsc -nologo -std:c++20 -experimental:module -TP -c -exportHeader -headerName:angle iostream -ifcOutput build\stlmodules\cache -Fobuild\stlmodules\cache\iostream.ifc.obj
iostream
[ 37%]: generating.cxx.headerunit.bmi iostream
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64\cl.exe /EHsc -nologo -std:c++20 -experimental:module -TP -c -exportHeader -headerName:angle iostream -ifcOutput build\stlmodules\cache -Fobuild\stlmodules\cache\iostream.ifc.obj
iostream
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\include\iostream(76): error C3474: could not open output file 'build\stlmodules\cache\iostream.ifc'
error: @programdir\modules\private\async\runjobs.lua:256: @programdir\core\sandbox\modules\os.lua:372: execv(C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.33.31629\bin\HostX64\x64\cl.exe /EHsc -nologo -std:c++20 -experimental:module -TP -c -exportHeader -headerName:angle iostream -ifcOutput build\stlmodules\cache -Fobuild\stlmodules\cache\iostream.ifc.obj) failed(2)
```

Issue #2719, opened 2022-08-25
Packages hash should inherit their parent package hashes


SirLynix:
### Is your feature request related to a problem? Please describe.

As of today, xmake handles packages by computing a hash for them, this hash is built based on package version, configs options, platform, arch, etc.

This means that if a newer version of a package is available and require lock isn't used, xmake will propose to upgrade a package, which is great.

However, if package `bar` depends on package `foo`, and package `foo` gets updated, package `bar` key won't change, this can lead to issues.

```lua
-- bar depends on foo
add_requires("foo", "bar")

-- app uses both foo and bar
target("app")
    add_packages("foo", "bar")
```

1) If `foo` and `bar` aren't found on the computer, no problem, xmake will install `foo` and then install `bar` using `foo`.
2) Sometimes later, `foo` receives an update (security update for instance)
3) xmake will ask to update `foo`, but not `bar`
4) We now have two differents versions of `foo` and `bar` for `app`

This can be problematic for a lot of reasons:

- Security updates won't be fully applied
- ABI issues (if foo update changes the ABI)
- Two different versions of .dll/.so/.dynlib for package foo is shared, which one will be used is uncertain
- And more

So until `bar` is updated, this will be an issue, and of course package don't receive an update every time their dependencies is updated (especially for minor/security fixes)

### Describe the solution you'd like

A simple solution would be that packages inherit their parent configs when building the hash, so when the dependencies hash changes, the package hash changes too.

Of course, this means that nearly all packages hash will change if this update is applied, which was also the case with vs_runtime. I don't think it would be a real issue to do it, especially for xmake 2.7.1

### Describe alternatives you've considered

The only alternative I can think of is to freeze the package version to be the same on both sides, or force uninstall both packages when an update occurs.

waruqi:
It does have some problems, but modifying the hash directly also introduces a lot of problems.

For example, libcurl, openssl and so on, whose version updates basically do not modify abi

2. it will break the package hash for existing users, resulting in a lot of package re-installations

3. affects all existing pre-compiled packages in build-artifacts, and reduces the pre-compile hit rate of new packages, with child package updates failing to trigger pre-compile builds of parent packages. All pre-built artifacts will gradually fail as package versions are updated.

While the current hash logic does have some problems, calculating the dependency on hash will lead to many more problems.

Perhaps we need a better solution that can avoid breaking the current compatibility as much as possible.

[...]

PR #2781, opened 2022-09-05
Strict compatibility for package linkdeps

waruqi:
https://github.com/xmake-io/xmake/issues/2719

we can use `set_policy("package.librarydeps.strict_compatibility, true)` to limit package compatibility in package().

```lua
package("libpng")
    add_deps("zlib")
    set_policy("package.librarydeps.strict_compatibility, true)
```

Or we can also use `set_policy("package.librarydeps.strict_compatibility, true)` to limit all packages compatibility in project xmake.lua

```lua
set_policy("package.librarydeps.strict_compatibility, true)
add_requires("libpng")
```

for example:

I updated zlib 1.2.11 => 1.2.12 in package()

### Dependency compatibility is not strictly tested by default

```console
$ xmake f -c
do nothing
```

### Enable strict compatibility for librarydeps

```console
$ xmake f -c --policies=package.librarydeps.strict_compatibility
in local-repo:
  -> libpng v1.6.37 [deps:*zlib]
please input: y (y/n/m)
```

### DIFF
--- a/xmake/actions/package/local/main.lua
+++ b/xmake/actions/package/local/main.lua
@@ -26,16 +26,16 @@ import("core.project.config")
 import("core.project.project")
 import("core.base.bit")
 
--- get link deps
-function _get_linkdeps(target)
-    local linkdeps = {}
+-- get library deps
+function _get_librarydeps(target)
+    local librarydeps = {}
     for _, depname in ipairs(target:get("deps")) do
         local dep = project.target(depname)
         if not ((target:is_binary() or target:is_shared()) and dep:is_static()) then
-            table.insert(linkdeps, dep:name())
+            table.insert(librarydeps, dep:name())
         end
     end
-    return linkdeps
+    return librarydeps
 end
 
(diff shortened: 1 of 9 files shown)


SirLynix:
This sounds great! However I don't think only linked packages should be considered, since you can have a lot of code coming from header-only libs as well.

waruqi:
I will continue to improve it tomorrow, right now it is not fully working

waruqi:
> This sounds great! However I don't think only linked packages should be considered, since you can have a lot of code coming from header-only libs as well.

It is only the name that is ambiguous, it actually already contains the headeronly library.

https://github.com/xmake-io/xmake/blob/ad0112c31cf46fa7e42ba79efdc60c14dafd6416/xmake/modules/private/action/require/impl/package.lua#L293

waruqi:
I have rename linkdeps to librarydeps and it works fine for me.

you can try it. @SirLynix

PR #2787, opened 2022-09-06
Improve cjson to support empty array

waruqi:
https://github.com/mpx/lua-cjson/issues/11


Wyfy0107:
Hello, when would this be released? I'm using cjson and this feature would be awesome

waruqi:
Maybe after a month, you can start with the master version

PR #2807, opened 2022-09-12
Rename 'runtime_version' to 'version'

WubiCookie:


xmake/modules/detect/sdks/find_matlab.lua
```diff
@@ -33,7 +33,7 @@ import("detect.sdks.matlab")
 --
 function main(opt)
     opt = opt or {}
-    local runtime_version = opt.runtime_version and tostring(opt.runtime_version) or nil
+    local version = opt.version and tostring(opt.version) or nil
```
waruqi, line 36: please use require_version, all find_package and find_tool used it.

Issue #2814, opened 2022-09-13
Qt: always compile .ui before running moc.


paul-reilly:
### Is your feature request related to a problem? Please describe.

As discussed on Discord, when doing a clean build of a Qt project with .ui form files, sometimes the build fails because `rule(qt.ui).before_buildcmd_file` and `rule(qt.moc). before_buildcmd_file` are run in parallel and ui_ files are not generated when needed.

### Describe the solution you'd like

For the ui_ files always to be generated when needed.

_(originally posted in xmake-repo, I closed that one)_

waruqi:
try `xmake update -s github:xmake-io/xmake#rule`

paul-reilly:
Wow, that was quick, thanks!

It's working with a clean build here on Manjaro and in CI on Ubuntu: https://github.com/paul-reilly/gideros/actions/runs/3059134920

PR #2817, opened 2022-09-14
Improve rule to support dependence order

waruqi:
https://github.com/xmake-io/xmake/issues/2814


waruqi:
@xq114 @Arthapz @SirLynix

I have adjusted the script execution logic for the internal rules and I am not sure if it will break the existing project. But in theory it shouldn't affect existing projects, it should be consistent with the previous behaviour.

But for the better stability, I would also like your help to test together if it affects your own projects.

`xmake update github:xmake-io/xmake#rule`

SirLynix:
Seems to works fine on my projects!

waruqi:
ok, thanks.

Arthapz:
Work fine too on my side

waruqi:
> Work fine too on my side

thanks

xq114:
Good on my side too

waruqi:
thanks

PR #2833, opened 2022-09-17
remove `-fbuiltin-module-map` in the (windows)mingw environment make the compilation work

shithappens2016:
There is no builtin module for clang in the mingw environment, so remove `-fbuiltin-module-map` in the mingw environment will make the compilation work.

### DIFF
--- a/xmake/rules/c++/modules/modules_support/clang.lua
+++ b/xmake/rules/c++/modules/modules_support/clang.lua
@@ -489,7 +489,11 @@ function get_builtinmodulemapflag(target)
     if builtinmodulemapflag == nil then
         local compinst = target:compiler("cxx")
         if compinst:has_flags("-fbuiltin-module-map", "cxxflags", {flagskey = "clang_builtin_module_map"}) then
-            builtinmodulemapflag = "-fbuiltin-module-map"
+            if is_plat("windows", "mingw") then
+                builtinmodulemapflag = ""
+            else
+                builtinmodulemapflag = "-fbuiltin-module-map"
+            end
         end
         assert(builtinmodulemapflag, "compiler(clang): does not support c++ module!")
         _g.builtinmodulemapflag = builtinmodulemapflag or false
--- a/xmake/rules/c++/modules/modules_support/clang.lua
+++ b/xmake/rules/c++/modules/modules_support/clang.lua
@@ -487,15 +487,15 @@ end
 function get_builtinmodulemapflag(target)
     local builtinmodulemapflag = _g.builtinmodulemapflag
     if builtinmodulemapflag == nil then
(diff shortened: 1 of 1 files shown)


xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -489,7 +489,11 @@ function get_builtinmodulemapflag(target)
     if builtinmodulemapflag == nil then
         local compinst = target:compiler("cxx")
         if compinst:has_flags("-fbuiltin-module-map", "cxxflags", {flagskey = "clang_builtin_module_map"}) then
-            builtinmodulemapflag = "-fbuiltin-module-map"
+            if is_plat("windows", "mingw") then
```
waruqi, line 492: target:is_plat

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -489,7 +489,11 @@ function get_builtinmodulemapflag(target)
     if builtinmodulemapflag == nil then
         local compinst = target:compiler("cxx")
         if compinst:has_flags("-fbuiltin-module-map", "cxxflags", {flagskey = "clang_builtin_module_map"}) then
-            builtinmodulemapflag = "-fbuiltin-module-map"
+            if target:is_plat("windows", "mingw") then
```
waruqi, line 492: windows plat will use msvc.

waruqi:
@Arthapz any idea?

xmake/rules/c++/modules/modules_support/clang.lua
waruqi: for clang-cl? did you test it?

shithappens2016: Tested in windows msys2(mingw-w64) environment and it works fine. msvc and clang-cl are not part of the mingw environment.

waruqi: so we should just use ` if target:is_plat("mingw") then`

shithappens2016: No. linux also has a mingw backend.

waruqi: but you should not add the windows platform, which is always used for msvc.

SirLynix: I don't understand how we get inside that `if`, shouldn't `compinst:has_flags("-fbuiltin-module-map", "cxxflags")` return false for mingw?

Arthapz: The flag can be supported but but the modulemap file can be missing, and it seems clang on mingw doesn't distribute it

Issue #2866, opened 2022-09-26
Better control over the order of execution of rules


waruqi:
### Is your feature request related to a problem? Please describe.

Although `add_deps("xxx", {order = true})` can also control the order of rules, we can't get it to be executed before some of the built-in rules if it's a user-defined rule.

### Describe the solution you'd like

I haven't figured out how to do it yet.

waruqi:
like this?

```lua
rule("cppfront")
    add_orders("c++.modules", {before = true})
    add_orders("xxxx", {after = true})
```

waruqi:
I think it would probably be nice to offer the possibility of rewriting the built-in rules, and we could do other things like adding c++ file extensions and so on, and be able to control the order of dependencies using a consistent approach.

like this.

```lua
rule("c++.build.modules2")
    set_base("c++.build.modules", {override = true})

rule("cppfront")
    add_deps("c++.build.modules2", {order = true})
```

Arthapz:
I like the clone method, it will provide a lot better way to handle generated C++ as we don't need anymore to build the files by hand (and it enable c++ module generation)

PR #2874, opened 2022-09-27
Support for dynamic creation and injection of rules and targets

waruqi:
https://github.com/xmake-io/xmake/issues/2866


waruqi:
@Arthapz @SirLynix @xq114  Can you help me test your project, like, project generator, ..

I have changed a lot of the logic related to target/rule.

Arthapz:
@waruqi  seem like Ok, doesn't break my projects

waruqi:
> @waruqi seem like Ok, doesn't break my projects, but didn't found a way to make my cppfront rule work

Even though I was able to get the cppfront rules executed first by injecting dependencies, the c++modules dependency scan analysis is not executed in the batchjob, it is still executed early because it is executed during the build batchjob dependency graph phase.

```lua
rule("cppfront")
    set_extensions(".cpp2")
    on_load(function (target)
        -- modify and replace the builtin rule
        local rule = target:rule("c++.build.modules.builder"):clone()
        rule:add("deps", "cppfront", {order = true})
        target:rule_add(rule)
    end)
    on_build_file(function (target, sourcefile, opt)
        print("build cppfront file")
    end)

target("test")
    set_kind("binary")
    add_rules("cppfront")
    add_files("src/*.cpp")
    add_files("src/*.cpp2")
```

https://github.com/xmake-io/xmake/blob/76bced915dfb870a34528270a02236a3f5f6ff4e/xmake/rules/c%2B%2B/modules/xmake.lua#L60

PR #2959, opened 2022-10-22
Stabilize vs and vsxmake project generation

SirLynix:
vs and vsxmake file generation isn't stable because of the hashtables used internally by xmake, this can be a problem for people saving those files on git, and also because vsxmake saves files only if changes are made (_writefileifneeded).

Changing a .sln/.vcxproj file triggers visual studio to ask if a reload is needed, stabilizing generation prevents this (at least for vsxmake as vs generator currently always write those files).

I made generation stable by changing the generators but had to change `project.ordertargets()` and `target:orderpkgs()` to make them stable between xmake runs as well.

### DIFF
--- a/xmake/core/project/project.lua
+++ b/xmake/core/project/project.lua
@@ -896,7 +896,8 @@ end
 function project.ordertargets()
     local ordertargets = project._memcache():get("ordertargets")
     if not ordertargets then
-        local targets = project.targets()
+        local targets = table.values(project.targets())
+        table.sort(targets, function (a,b) return a:name() < b:name() end)
         ordertargets = {}
         local targetrefs = {}
         for _, t in pairs(targets) do
--- a/xmake/core/project/project.lua
+++ b/xmake/core/project/project.lua
@@ -896,11 +896,12 @@ end
 function project.ordertargets()
     local ordertargets = project._memcache():get("ordertargets")
     if not ordertargets then
-        local targets = table.values(project.targets())
-        table.sort(targets, function (a,b) return a:name() < b:name() end)
+        local targets = project.targets()
+        local sorted_targets = table.values(targets)
+        table.sort(sorted_targets, function (a,b) return a:name() < b:name() end)
         ordertargets = {}
         local targetrefs = {}
-        for _, t in pairs(targets) do
+        for _, t in ipairs(sorted_targets) do
             instance_deps.sort_deps(targets, ordertargets, targetrefs, t)
         end
         project._memcache():set("ordertargets", ordertargets)
--- a/xmake/core/project/target.lua
+++ b/xmake/core/project/target.lua
@@ -980,7 +980,9 @@ function _instance:pkgenvs()
                 end
             end
         end
-        for _, pkg in pkgs:keys() do
+        local orderedpkgs = pkgs:to_array()
+        table.sort(orderedpkgs, function (a, b) return a:name() < b:name() end)
+        for _, pkg in ipairs(orderedpkgs) do
             local envs = pkg:get("envs")
             if envs then
                 for name, values in pairs(envs) do
--- a/xmake/core/project/target.lua
+++ b/xmake/core/project/target.lua
@@ -953,6 +953,7 @@ function _instance:orderpkgs(opt)
                 end
             end
         end
+        table.sort(packages, function (a, b) return a:name() < b:name() end)
         self:_memcache():set(cachekey, packages)
     end
     return packages
--- a/xmake/plugins/project/vstudio/impl/vs201x.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x.lua
@@ -335,8 +335,9 @@ function _make_targetinfo(mode, arch, target, vcxprojdir)
         end
     end
     for k, v in pairs(addrunenvs) do
+        table.sort(v)
         if k:upper() == "PATH" then
-            runenvs[k] = _translate_path(v, vcxprojdir) .. ";$([System.Environment]::GetEnvironmentVariable('" .. k .. "'))"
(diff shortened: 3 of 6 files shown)


xmake/core/project/project.lua
```diff
@@ -896,7 +896,8 @@ end
 function project.ordertargets()
     local ordertargets = project._memcache():get("ordertargets")
     if not ordertargets then
-        local targets = project.targets()
+        local targets = table.values(project.targets())
+        table.sort(targets, function (a,b) return a:name() < b:name() end)
```
waruqi, line 900: targets is map, it will break

SirLynix: I used `table.values` on it to turn it into an array

xmake/core/project/target.lua
```diff
@@ -953,6 +953,7 @@ function _instance:orderpkgs(opt)
                 end
             end
         end
+        table.sort(packages, function (a, b) return a:name() < b:name() end)
```
waruqi, line 956: we should not sort it, it will break package links order

SirLynix: from my understanding, this table comes from `project.required_packages()` which calls `project._load_requires()` whichs doesn't seem to sort packages?

Those are xmake internals so I'm not sure at all, the only things I'm sure is that this table is not stable.

xmake/plugins/project/vstudio/impl/vs201x.lua
```diff
@@ -335,6 +335,7 @@ function _make_targetinfo(mode, arch, target, vcxprojdir)
         end
     end
     for k, v in pairs(addrunenvs) do
+        table.sort(v)
```
waruqi, line 338: We should not sort runenvs, It may affect the order in which packages so/dll are loaded, and if there are multiple so libraries with the same name, they behave differently.

SirLynix: I agree this could be an issue and same as above: I sorted it because order changed between runs, a better fix may be required here.

xmake/plugins/project/vstudio/impl/vs201x_solution.lua
```diff
@@ -35,23 +35,26 @@ function _make_projects(slnfile, vsinfo)
     -- make all targets
     local groups = {}
     local targets = {}
+    local default_targets = {}
     local vctool = "8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942"
-    for targetname, target in pairs(project.targets()) do
+    for targetname, target in pairs(project.ordertargets()) do
```
waruqi, line 40: it is array, not table

xmake/core/project/project.lua
waruqi: but instance_deps.sort_deps need map, not array.

xmake/core/project/target.lua
waruqi: Even if it is unstable, it is at least guaranteed to be in the correct linking order. Sorting it would completely break the linking order of the packages.

SirLynix: maybe package deps should be sorted by name then?

waruqi: no, we cannot sort it.

SirLynix: In fact I'm not sure where the order is lost, since package deps are already given in a specific order.

xmake/plugins/project/vstudio/impl/vs201x.lua
SirLynix: I think I fixed that in target:pkgenvs(), packages order wasn't stable (because it came from a hashmap) so I made it into an array and sorted it.

xmake/core/project/target.lua
SirLynix: I removed it, it wasn't necessary

xmake/core/project/project.lua
SirLynix: Fixed

xmake/core/project/target.lua
```diff
@@ -979,7 +980,9 @@ function _instance:pkgenvs()
                 end
             end
         end
-        for _, pkg in pkgs:keys() do
```
waruqi, line 982: we should add `hashset:orderkeys` like table.orderpairs()

SirLynix: I'm not sure how to do that in a better way than making a table and sorting it

waruqi:
we should use hashset:orderkeys and table.orderpairs

https://github.com/xmake-io/xmake/pull/2971

xmake/core/project/project.lua
```diff
@@ -897,9 +897,11 @@ function project.ordertargets()
     local ordertargets = project._memcache():get("ordertargets")
     if not ordertargets then
         local targets = project.targets()
+        local sorted_targets = table.values(targets)
+        table.sort(sorted_targets, function (a,b) return a:name() < b:name() end)
         ordertargets = {}
         local targetrefs = {}
-        for _, t in pairs(targets) do
+        for _, t in ipairs(sorted_targets) do
             instance_deps.sort_deps(targets, ordertargets, targetrefs, t)
```
waruqi, line 905: t:get("deps") is not fixed order, but it is just an issue your project.

https://github.com/NazaraEngine/NazaraEngine/blob/0a8048809c014bf41fa4c50f0cff4fc689832635/xmake.lua#L117

```lua
for name, module in pairs(rendererBackends) do
	ModuleTarget(name, module)
end
```

you should use `table.orderpairs`

SirLynix: you're right, however I'm testing project generation on this project: https://github.com/NazaraEngine/ShaderLang (which should be stable)

PR #2967, opened 2022-10-24
Fix issue#2954.

CCCptH:
Character `:` can not exist in path in windows system. So when using clang to build project that contains module partitions, the filename of bmifile it generates should not contains `:`
Change bmifile filename according to [https://clang.llvm.org/docs/StandardCPlusPlusModules.html#global-module-fragment.](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#global-module-fragment.).

> The file name of BMIs should end with .pcm. The file name of the BMI of a primary module interface unit should be module_name.pcm. The file name of BMIs of module partition unit should be module_name-partition_name.pcm.


waruqi:
see https://github.com/xmake-io/xmake/pull/2970

Issue #3000, opened 2022-10-30
Incremental compilation support for modules


waruqi:
### Xmake Version

2.7.2

### Operating System Version and Architecture

ubuntu

### Describe Bug

Modifying headers doesn't trigger recompiling; https://github.com/xmake-io/xmake/issues/2961

we can test it in user_headerunits test.

- [x] xxx.mpp cannot generate deps with header
  - [x] gcc
  - [x] clang
  - [x] msvc 
- [x] we need improve to parse header file from xxx.gcm.d and xxx.cpp.o.d if it contains modules
  - [x] gcc
  - [x] clang
  - [x] msvc 
- [x] ccache will also affect incremental compilation (disabled)
  - [x] object 
  - [x] fix dependfile 

### Expected Behavior

..

waruqi:
we need improve to parse include deps with modules for gcc.

hello.gcm.d

```
build/.objs/user_headerunit/linux/x86_64/release/src/hello.mpp.o  /mnt/xmake/tests/projects/c++/modules/user_headerunit/build/.gens/user_headerunit/linux/x86_64/release/rules/modules/cache/hello.gcm:  src/hello.mpp\
build/.objs/user_headerunit/linux/x86_64/release/src/hello.mpp.o  /mnt/xmake/tests/projects/c++/modules/user_headerunit/build/.gens/user_headerunit/linux/x86_64/release/rules/modules/cache/hello.gcm:  ./src/header.hpp.c++m\
hello.c++m:  /mnt/xmake/tests/projects/c++/modules/user_headerunit/build/.gens/user_headerunit/linux/x86_64/release/rules/modules/cache/hello.gcm\
.PHONY: hello.c++m\
/mnt/xmake/tests/projects/c++/modules/user_headerunit/build/.gens/user_headerunit/linux/x86_64/release/rules/modules/cache/hello.gcm:|  build/.objs/user_headerunit/linux/x86_64/release/src/hello.mpp.o\
CXX_IMPORTS += ./src/header.hpp.c++m
```

barcharcraz:
> we need improve to parse include deps with modules.
> ### gcc
> 
> hello.gcm.d
> 
> ```
> build/.objs/src/hello.mpp.o  /modules/cache/hello.gcm:  src/hello.mpp\
> build/.objs/src/hello.mpp.o  /modules/cache/hello.gcm:  ./src/header.hpp.c++m\
> hello.c++m:  /modules/cache/hello.gcm\
[...]

Issue #2961, opened 2022-10-22
Compiling Modules Inconsistant


cgbsu:
### Xmake Version

2.7.2

### Operating System Version and Architecture

Ubuntu (Container, Podman, amd x64)

### Describe Bug

Having some trouble using C++ modules with xmake, it seems inconsistent

If I compile a header into a module (Header.hpp + Header.mpp) it works... once, but does not propagate changes made to headers included by Header.hpp into the Header module (though I can, say, still call a new function that I add to an include of Header.hpp), then I have to delete .xmake and build

If I just use Header.mpp with Partition0.mpp Partition1.mpp etc.. it also works... once, but then doesn't seem to import anything from the module (in particular after an xmake clean) and it only works if I have a parallel set of headers Partition1.hpp Partition0.hpp that I have already put into a module as described above (even if I have removed them), though changes from PartitionX.mpp will be propagated and recompiled at this point.

Deleting .xmake and build will NOT work at this point, if I build again, it wont import anything from the modules, thought it appears to compile the modules themselves.

Its all really confusing

I have to use a gcc snapshot because gcc 12.1 or 12.2 crashes when compiling modules
gcc-13.0.0 seems to work OK (but not when including some headers, see the comment I put [here](https://gcc.gnu.org/bugzilla//show_bug.cgi?id=99241#c5)

You can see what I did hear 
["Master Header" version](https://github.com/cgbsu/WarpBootstrapCompiler2/commit/ae62c4db6b0e9b103c0c1ab9bfb88f8d044fb67f) (where Utilities.hpp includes several files and is imported by Utilities.mpp)
[Module Partition Version](https://github.com/cgbsu/WarpBootstrapCompiler2/commit/9e186e1060e1cff4871f1721cdf962273ca7a5a7)

These 3 commits may be of use: 
a3412b1bd2b4d902db102f3f39165bf0754b3ce1
ae62c4db6b0e9b103c0c1ab9bfb88f8d044fb67f
9e186e1060e1cff4871f1721cdf962273ca7a5a7

For the sake for being forthcoming I should mention that I am running as root with `XMAKE_ROOT=y` in a container

### Expected Behavior

~: xmake // Works
// Change file here 
~: xmake // Works, if the code itself does not have a problem, changes propagated to module files/caches in both cases ("Master Header" vs "Partitioned")
~: xmake clean
[...]

PR #3001, opened 2022-10-30
Incremental compilation support for modules #3000

waruqi:
https://github.com/xmake-io/xmake/issues/3000
https://github.com/xmake-io/xmake/issues/2961


waruqi:
@Arthapz I have improved modules to support incremental compilation. you can try it.

Arthapz:
@waruqi Okay i'll try it when i find time

PR #3007, opened 2022-11-01
change user_name to $USER

xqyjlj:
I often work in different dockers, and these dockers all have different usernames. So I changed the username of the ~/.xmake/profile file to $USER, in this condition, I can copy the xmake files to docker without making any changes：

```
old: export XMAKE_ROOTDIR = "/home/username/.local/bin"
new: export XMAKE_ROOTDIR = "/home/$USER/.local/bin"
```

### DIFF
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -230,6 +230,7 @@ write_profile()
 install_profile()
 {
     if [ ! -d ~/.xmake ]; then mkdir ~/.xmake; fi
+    prefix=${prefix//$USER/\$USER}
     echo "export XMAKE_ROOTDIR=\"$prefix/bin\"" > ~/.xmake/profile
     echo 'export PATH="$XMAKE_ROOTDIR:$PATH"' >> ~/.xmake/profile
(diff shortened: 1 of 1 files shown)


waruqi:
? it use `~/.local` not `$USER` 

I don't think this replacement operation should be valid, because it does not contains `$USER`

https://github.com/xmake-io/xmake/blob/e934dbed6f30a7d2f42edaf2f74120efdd041201/scripts/get.sh#L219

and `$USER` is empty on my docker image

xqyjlj:
`prefix=~/.local` seems to expand to `/home/username/.local` on ubuntu22.04.
Should I write like this?
```
${prefix//"/home/$USER"/\~}
```

waruqi:
> and if in root, $USER is empty

Even if it's not root, it's still empty on my docker

waruqi:
You can test it locally to verify it. But I still don't understand why you would want to modify it. Even if you switch between dockers, you just need the `~/.xmake/profile` of the current user, which always has a path to the current user's directory.

xqyjlj:
Well, this may be a bad habit of my using docker. I will close this pr

Issue #3009, opened 2022-11-01
Two mutually exclusive clang flags for C++20 modules compilation are included


aacirino:
### Xmake Version

2.7.1

### Operating System Version and Architecture

Fedora 36, clang 15

### Describe Bug

Since some time I've had problems building C++20 projects with modules. I had to resort to adding some compile flags but got tired of it and noticed that the compiler flags generation is including two mutually exclusive flags. In `xmake/rules/c++/modules/modules_support/clang.lua` we have

`local implicitmodulesflag = get_implicitmodulesflag(target)`
`local noimplicitmodulemapsflag = get_noimplicitmodulemapsflag(target)`

and then

`target:add("cxxflags", implicitmodulesflag, {force = true})`
`target:add("cxxflags", noimplicitmodulemapsflag, {force = true})`

Xmake should include only the implicit modules flag, allowing us to use clang's module maps transparently. Unfortunately xmake will not build in my environment due to an error something like `No rule to build the target 'lua/lapi.c'` (my system is in Brazilian Portuguese so I have to guess this is how it is phrased in English). I would change the modules flag generation myself and submit a pull request if I could build xmake.
In the meantime I have to keep adding the following flags to xmake.lua:

`"-fimplicit-module-maps",`
`"-fmodule-map-file=$(shell brew --prefix)/include/c++/v1/module.modulemap",`

### Expected Behavior

xmake should include only the `-fimplicit-modules` flag.

### Project Configuration

[trouble.zip](https://github.com/xmake-io/xmake/files/9914119/trouble.zip)

### Additional Information and Error Logs

[~/git/trouble]$ xmake -bv
[  0%]: generating.cxx.module.deps src/my_module.cpp
[ 23%]: generating.cxx.module.bmi my_module
/home/linuxbrew/.linuxbrew/opt/llvm/bin/clang -Qunused-arguments -m64 -fPIC -O3 -std=c++20 -fmodules -fbuiltin-module-map -fimplicit-modules -fno-implicit-module-maps -DNDEBUG -fmodules-cache-path=build/.gens/trouble/linux/x86_64/release/rules/modules/cache -c -x c++-module --precompile src/my_module.mpp -o build/.gens/trouble/linux/x86_64/release/rules/modules/cache/my_module.pcm
src/my_module.mpp:3:8: fatal error: module 'std' not found
import std;
~~~~~~~^~~
1 error generated.

waruqi:
[...]

PR #3016, opened 2022-11-03
Improve clang/gcc/msvc to better support std modules

waruqi:
https://github.com/xmake-io/xmake/issues/3009
https://github.com/xmake-io/xmake/issues/3022


waruqi:
We also need to better address the configuration of libc++ and libstdc++, and the handling of symbols

PR #3133, opened 2022-12-04
add support of msvc specific /internalPartition for modules

Arthapz:


Arthapz:
the 
```
{values = {["internalpartition"] = true}}
```

is for fallback generator, compiler implementing p1689r5 should handle it (msvc does)

waruqi:
> the
> 
> ```
> {values = {["internalpartition"] = true}}
> ```
> 
> is for fallback generator, compiler implementing p1689r5 should handle it (msvc does)

Is there a better way to detect it automatically? For example by analysing the module source code.

Arthapz:
it's possible i think, but i think it will complexifies a lot the code as for internalpartition detection we need to know the other modules files and see if an exported partition with the same name is already present

Arthapz:
it's possible i think, but i think it will complexifies a lot the code as for internalpartition detection we need to know the other modules files and see if an exported partition with the same name is already present

xmake/rules/c++/modules/modules_support/common.lua
```diff
@@ -453,9 +452,10 @@ function fallback_generate_dependencies(target, jsonfile, sourcefile, preprocess
         end
         if not module_name_private then
             module_name_private = line:match("module%s+(.+)%s*;") or line:match("__preprocessed_module%s+(.+)%s*;")
-        end
-        if internalpartition then
-            module_name_export = module_name_private
+            if module_name_private then
+                internal = module_name_private:find(":")
+                print(module_name_private, internal)
```
waruqi, line 457: remove print

PR #3134, opened 2022-12-04
Fix missing module include dirs on fallback deps generation step

Arthapz:


xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -259,7 +259,7 @@ function generate_dependencies(target, sourcebatch, opt)
                 for _, define in pairs(target:get("defines")) do
                     table.insert(defines, "-D" .. define)
                 end
-                local includedirs = {}
+                local includedirs = table.join2({}, target:get("includedirs"))
```
waruqi, line 262: please use table.join or table.join2(includedirs, ..)

PR #3143, opened 2022-12-07
Add emsdk and emscripten find scripts and config

SirLynix:
```
lynix@SirDesktop:/mnt/c/Projets/Perso/NazaraNext/NazaraEngine$ xmake.exe config --import=wasm.config -cv
checking for emsdk directory ... E:\GitHub\emsdk
checking for Emscripten toolchain directory ... E:\GitHub\emsdk\upstream\emscripten
checking for emcc.bat ... E:\GitHub\emsdk\upstream\emscripten\emcc.bat
```


waruqi:
I think it just need `find_emsdk` and it will return `{sdkdir, emscriptendir}` at same time.

we can find emscripten directory in find_emsdk. emsdk is a complete sdk with emscripten.

xmake/platforms/wasm/xmake.lua
```diff
@@ -40,3 +40,20 @@ platform("wasm")
     -- set toolchains
     set_toolchains("emcc")
 
+    -- set menu
+    set_menu {
+        config =
+        {
+            {category = "Emscripten Configuration"                    }
+        ,   {nil, "emsdk", "kv", nil, "emsdk directory" }
+        ,   {nil, "emscripten", "kv", nil, "emscripten directory" }
```
waruqi, line 49: we only need emsdk argument.

xmake/modules/detect/tools/find_emxx.lua
```diff
@@ -40,6 +41,17 @@ function main(opt)
...
+    -- init the search directories
+    local paths = {}
+    local emscriptendir = find_emscripten()
+    if emscriptendir then
+        table.insert(paths, emscriptendir.sdkdir)
+    end
+
+    -- find program
+    opt.paths = paths
+    opt.envs  = {PATH = os.getenv("PATH")}
```
waruqi, line 53: why? find_program will find it from $PATH by default.

xmake/modules/detect/sdks/find_emscripten.lua
```diff
@@ -0,0 +1,120 @@
...
+        if opt.verbose or option.get("verbose") then
+            cprint("checking for emscripten toolchain directory ... ${color.nothing}${text.nothing}")
+        end
+    end
+
+    -- save to cache
+    cacheinfo.sdk = sdk or false
+    detectcache:set(key, cacheinfo)
+    detectcache:save()
+    return sdk
```
waruqi, line 119: please find it in find_emsdk.lua

xmake/platforms/wasm/xmake.lua
SirLynix: from my understanding, you can have the emscripten toolchain without emsdk, that's why I separated them.
emsdk may be useful to install/update the emscripten toolchain where only the emscripten toolchain is required to build

xmake/modules/detect/sdks/find_emscripten.lua
SirLynix: from my understanding, you can have the emscripten toolchain without emsdk, that's why I separated them.
emsdk may be useful to install/update the emscripten toolchain where only the emscripten toolchain is required to build

xmake/modules/detect/tools/find_emxx.lua
SirLynix: didn't know that, I took it from find_vcpkg

xmake/platforms/wasm/xmake.lua
waruqi: > from my understanding, you can have the emscripten toolchain without emsdk, that's why I separated them.

But I have never seen a stand-alone emscripten toolchain.

I only know about installing and using emscripten in emsdk. Are there other installers and tarballs that only include emscripten without emsdk? I don't seem to have seen any.

SirLynix: I didn't see any either but they seem to be mostly isolated, emscripten folder doesn't seem to depend on emsdk folder. One can download emscripten without using emsdk which could be useful to add it as a toolchain.
Separating them seemed like the safest and most flexible option.

waruqi: This is similar to the relationship between vs/msvc, qtsdk/clang, but only emscripten is included in emsdk, and only emsdk is officially provided to install emscripten.

I don't think any other separate emscripten sdk will be distributed.

We could start by just find_emsdk to maintain the paths to emsdk and emscripten, which would be easier to maintain. Even if we come across a separate emscripten sdk later, it would be fine to go back and add it.

Also, we now have an emcc toolchain。

SirLynix: As you wish, I will update it

SirLynix:
Here you are, I tested it and it seems to work
image

xmake/modules/detect/tools/find_emar.lua
```diff
@@ -21,6 +21,7 @@
 -- imports
 import("core.tool.compiler")
 import("lib.detect.find_program")
+import("detect.sdks.find_emscripten")
```
waruqi, line 24: and here

xmake/modules/detect/tools/find_emxx.lua
```diff
@@ -21,6 +21,7 @@
 -- imports
 import("lib.detect.find_program")
 import("lib.detect.find_programver")
+import("detect.sdks.find_emscripten")
```
waruqi, line 24: here

SirLynix:
oops, I didn't test it enough.

I fixed it and fixed the cmake build, I tested it properly this time. 😅

Issue #3164, opened 2022-12-12
Add default components for packages


SirLynix:
### Is your feature request related to a problem? Please describe.

Hi.

Now that package components and rules are implemented, I plan to make new packages for Qt5 and Qt6, which would be based on them.

I plan to make two packages for Qt, `qt5` and `qt6` which would make use of components to allow to enable other Qt library. Components are perfect for the Qt library as the Qt installer (aqtinstaller) installs all Qt libraries.

This way, it would be easy to use Qt6:
```lua
add_requires("qt6")

target("MyWidgetApp")
    add_files("src/*.cpp")
    add_rules("@qt6/widgetapp")
    add_packages("qt6", { components = "widgets" }) -- widgets deps on gui and core
```

however most people will probably write this:
```lua
add_requires("qt6")

target("MyWidgetApp")
    add_files("src/*.cpp")
    add_rules("@qt6/widgetapp")
    add_packages("qt6") -- ??
```

with xmake 2.7.3, this will enable all qt6 components and libs, and there are dozens of them. it makes it too easy to enable too many libs.

### Describe the solution you'd like

I was thinking we could easily add a way to specify default components
```lua
package("qt6")
    add_components("core", "gui", "widgets", "network", ...)
    set_defaultcomponents("core", "gui", "widgets")
```

or maybe
```lua
    on_component("core", function (package, component)
        -- ...
        component:set_default(true)
    end)
```

these components would be the selected components if none if specified.

### Describe alternatives you've considered

adding an "error" component which would try to trigger a compilation error telling the user to select which components it wants to use.

waruqi:
It has been supported. you can use `add_configs()` to set default components.

```lua
package("test")
    add_configs("widgets",       {description = "", default = true, type = "boolean"})
    on_load(function (package)
        if package:config("widgets") then
            package:add("components", "main")
        end
    end)
```

waruqi:
try it https://github.com/xmake-io/xmake/pull/3167

```lua
add_components("graphics", {default = true})
```

or 

```lua
on_component(function (package, component))
    component:set("default", true)
end)
```

PR #3167, opened 2022-12-12
add default componments

waruqi:
https://github.com/xmake-io/xmake/issues/3164#issuecomment-1346782590

### DIFF
--- a/xmake/core/project/target.lua
+++ b/xmake/core/project/target.lua
@@ -419,7 +419,7 @@ function _instance:get_from_pkgs(name, opt)
         local components_default = pkg:components_default()
         if configinfo and (configinfo.components or components_default) and pkg:components() then
             local components_enabled = hashset.new()
-            for _, comp in ipairs(table.join(configinfo.components or {}, components_default)) do
+            for _, comp in ipairs(table.wrap(configinfo.components or components_default)) do
                 components_enabled:insert(comp)
                 for _, dep in ipairs(table.wrap(pkg:component_orderdeps(comp))) do
                     components_enabled:insert(dep)
--- a/xmake/core/project/target.lua
+++ b/xmake/core/project/target.lua
@@ -416,10 +416,10 @@ function _instance:get_from_pkgs(name, opt)
         local configinfo = self:pkgconfig(pkg:name())
         -- get values from package components
         -- e.g. `add_packages("sfml", {components = {"graphics", "window"}})`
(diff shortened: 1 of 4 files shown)


xmake/core/project/target.lua
```diff
@@ -416,9 +416,10 @@ function _instance:get_from_pkgs(name, opt)
         local configinfo = self:pkgconfig(pkg:name())
         -- get values from package components
         -- e.g. `add_packages("sfml", {components = {"graphics", "window"}})`
-        if configinfo and configinfo.components and pkg:components() then
+        local components_default = pkg:components_default()
+        if configinfo and (configinfo.components or components_default) and pkg:components() then
             local components_enabled = hashset.new()
-            for _, comp in ipairs(table.wrap(configinfo.components)) do
+            for _, comp in ipairs(table.join(configinfo.components or {}, components_default)) do
```
SirLynix, line 422: doesn't that mean that those components can't be disabled? that's not what I meant sorry for the confusion.

`for _, comp in ipairs(table.wrap(configinfo.components or components_default)) do` is more what I suggest

waruqi: ? Components that are not set with {compoments=""} and default=true are disabled by default.

you can do not set them.

waruqi: I have improved it, try it again.

xmake/core/project/target.lua
```diff
@@ -416,9 +416,10 @@ function _instance:get_from_pkgs(name, opt)
         local configinfo = self:pkgconfig(pkg:name())
         -- get values from package components
         -- e.g. `add_packages("sfml", {components = {"graphics", "window"}})`
-        if configinfo and configinfo.components and pkg:components() then
+        local components_default = pkg:components_default()
+        if configinfo and (configinfo.components or components_default) and pkg:components() then
```
SirLynix, line 420: this won't work if configinfo is nil (which is the case with `add_packages("xx")`

this works:
```lua
        local selected_components = configinfo and configinfo.components or pkg:components_default()
        if selected_components and pkg:components() then
            local components_enabled = hashset.new()
            for _, comp in ipairs(table.wrap(selected_components)) do
```

waruqi: try it again.

SirLynix: works fine, thanks! 👍

PR #3168, opened 2022-12-12
add support of C++23 standard modules on msvc

Arthapz:


Qudix:
Is stdifc even needed anymore? It's basically dead in the water now that MSVC is following the standard naming conventions and you don't need to install the "modules support" package as they come included with the STL itself. c++20 modules were essentially just an experiment.

Arthapz:
it's true, if ruki want it removed, i'll remove it

waruqi:
> it's true, if ruki want it removed, i'll remove it

you can remove it.

xmake/rules/c++/modules/modules_support/msvc.lua
```diff
@@ -127,27 +128,30 @@ function load(target)
...
-                    local stdifcdir = path.join(vcvars.VCInstallDir, "Tools", "MSVC", vcvars.VCToolsVersion, "ifc", arch)
-                    if os.isdir(stdifcdir) then
-                        target:add("cxxflags", {stdifcdirflag, winos.short_path(stdifcdir)}, {force = true, expand = false})
-                    end
-                end
+            if vcvars.VCInstallDir and vcvars.VCToolsVersion and semver.compare(vcvars.VCToolsVersion, "14.35") then
+                stdmodulesdir = path.join(vcvars.VCInstallDir, "Tools", "MSVC", vcvars.VCToolsVersion, "modules")
             end
         end
+        target:set("values", "c++.msvc.enable_std_import", iscpp23 and os.isdir(stdmodulesdir))
```
waruqi, line 154: if it's only private data in rule, you should use `target:data_set()` and `target:data()`

xmake/rules/c++/modules/modules_support/msvc.lua
```diff
@@ -127,27 +128,30 @@ function load(target)
     local modulesflag = get_modulesflag(target)
     target:add("cxxflags", modulesflag)
 
-    -- add stdifcdir in case of if the user ask for it
-    local stdifcdirflag = get_stdifcdirflag(target)
-    if stdifcdirflag then
+    -- enable std modules if c++23 by defaults
+    if target:values("c++.msvc.enable_std_import") == nil then
+        local languages = table.join({}, target:get("languages"))
```
waruqi, line 133: you just need:

```lua
for _, language in ipairs(languages) do
```

xmake/rules/c++/modules/modules_support/msvc.lua
```diff
@@ -127,27 +128,30 @@ function load(target)
...
+        local iscpp23 = false
+        for _, language in pairs(languages) do
+            if language:find("c++") or language:find("cxx") then
```
waruqi, line 136: language:startswith("c++"), it will be faster.

find will handle lua pattern.

xmake/rules/c++/modules/modules_support/msvc.lua
```diff
@@ -127,27 +128,30 @@ function load(target)
...
+                iscpp23 = true
+                for _, version in pairs({"11", "14", "17", "20"}) do
```
waruqi, line 138: ipairs

waruqi: two loop, it may be slower.

waruqi: we can use `local version = language:match("%d+")` to get version

xmake/rules/c++/modules/modules_support/msvc.lua
Arthapz: it's not private, i think some users may need to disable std modules

Issue #3192, opened 2022-12-22
JSON output for auto complete 


glcraft:
### Is your feature request related to a problem? Please describe.

The auto completion support for xmake can be improved/extended for other shells like [Nu shell](https://www.nushell.sh/) or Powershell by outputting result in JSON.

### Describe the solution you'd like

Add on option for `xmake lua private.utils.complete` to output in JSON.

### Additional context

*Hold on for my PR in a bit :)*

waruqi:
Thanks, I have merged it.

PR #3193, opened 2022-12-22
Json output in auto complete module

glcraft:
Feature request : #3192 

Output auto complete in JSON for shells that support it.

Here is an example for Nu shell : 
```
# in config.nu

# Xmake completer
let xmake_completer = {|spans| 
  XMAKE_SKIP_HISTORY=1 XMAKE_ROOT=y xmake lua 'private.utils.complete' 0 'nospace,json' $spans | from json | sort-by value
}

let-env config = {
# ...
completions: {
    case_sensitive: false # set to true to enable case-sensitive completions
    quick: true  # set this to false to prevent auto-selecting completions when only one remains
    partial: true  # set this to false to prevent partial filling of the prompt
    algorithm: "prefix"  # prefix or fuzzy
    external: {
      enable: true # set to false to prevent nushell looking into $env.PATH to find more suggestions, `false` recommended for WSL users as this look up my be very slow
      max_results: 100 # setting it lower can improve completion performance at the cost of omitting some options
      completer: $xmake_completer
    }
  }
# ...
}
```
image

### DIFF
--- a/xmake/modules/private/utils/complete.lua
+++ b/xmake/modules/private/utils/complete.lua
@@ -22,10 +22,12 @@
 import("core.base.option")
 import("core.base.task")
 import("core.base.cli")
+import("core.base.json")
 
 local raw_words = {}
 local raw_config
 local position = 0
+local no_json = false
 local use_spaces = true
 local no_key = false
 local reenter = false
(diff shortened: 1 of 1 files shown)


glcraft:
Note that by default, the output is still the same. To output in JSON, you have to add "json" option in the 2nd parameter like "nospace" for example

waruqi:
thanks

PR #3201, opened 2022-12-24
Add completer and xrepo complete

glcraft:
Transform private.utils.complete to a module completer which can be use for xmake and xrepo with the same code.

### DIFF
--- a/xmake/modules/private/utils/complete.lua
+++ b/xmake/modules/private/utils/complete.lua
@@ -14,7 +14,7 @@
 --
 -- Copyright (C) 2015-present, TBOOX Open Source Group.
 --
--- @author      OpportunityLiu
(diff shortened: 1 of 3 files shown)


waruqi:
thanks!

xmake/modules/private/utils/complete.lua
```diff
@@ -14,301 +14,20 @@
...
-    end
-
-    if optvs and #found_candidates == 0 then
-        -- show candidates of values
-        local values = optvs.values
-        if type(values) == "function" then
-            values = values(completing)
-        end
-        for _, v in ipairs(values) do
-            if tostring(v):startswith(completing) then
```
waruqi, line 218: it's broken now.

PR #3246, opened 2023-01-05
improve module cachedir

waruqi:
We should avoid creating directories in advance if possible. https://github.com/xmake-io/xmake/discussions/3245#discussioncomment-4601796


waruqi:
@Arthapz Can you help review this, I'm not sure if it will break anything as I haven't tested it in detail.

Arthapz:
seems to work on my side (tested on my personal toolkit and xmake tests with VS 15.2 preview 2)

waruqi:
thanks

PR #3317, opened 2023-01-27
Fix detection of target language for qt projects

xaizek:
Stray change in 02fb7ef69e78a5d6a8e0066f8866684533836ac3 broke it such that language set via `set_languages()` wasn't detected and `-std=c++11` was appended after `-std` flag that came from `set_languages()` in `xmake.lua`.

To reproduce:
 1. `xmake create -tqt.quickapp`
 2. `vim xmake.lua` and add `set_languages("c++17")`
 3. `xmake --verbose` and look at `-std=` options

### DIFF
--- a/xmake/rules/qt/load.lua
+++ b/xmake/rules/qt/load.lua
@@ -141,7 +141,8 @@ function main(target, opt)
     local languages = target:get("languages")
     local cxxlang = false
     for _, lang in ipairs(languages) do
-        if lang:startswith("cxx") or lang:startswith("c++") then
+        -- c++* or gnuc++*
+        if lang:find("cxx", 1, true) or lang:find("c++", 1, true) then
             cxxlang = true
             break
         end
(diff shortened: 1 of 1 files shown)


xmake/rules/qt/load.lua
```diff
@@ -141,7 +141,7 @@ function main(target, opt)
     local languages = target:get("languages")
     local cxxlang = false
     for _, lang in ipairs(languages) do
-        if lang:startswith("xx") or lang:startswith("++") then
+        if lang:startswith("cxx") or lang:startswith("c++") then
```
waruqi, line 144: This is to implement lookup support for gnuc++/gnucxx. maybe we should use find instead of startswith.

```lua
if lang:find("cxx", 1, true) or lang:find("c++", 1, true) then
```

xaizek: In that case I would do
```lua
if lang:startswith("cxx") or lang:startswith("c++") or lang:startswith("gnucxx") or lang:startswith("gnuc++") then
```
instead, but find should work as well at least until some weird `-std` value is introduced.

waruqi: This is a bit bloated, and `find("-std", 1, true)` only find plain text, it doesn't handle pattern matching.

Issue #3342, opened 2023-02-07
Improve to configure working and build directories


waruqi:
xmake will support two ways of configuring working directory.

#### Working directory

The current working directory where xmake is running, and it is also the directory where the project cache configuration .xmake is stored.

```
- workdir
  - .xmake
  - build
```

#### Build directory (-o/--buildir)

Build the output directory of the build artifacts.

```
build
└── macosx
    └── x86_64
        └── release
            ├── demo
```

#### Project directory (-P/--project)

The root directory of the project's source code, which is also the directory containing the root xmake.lua file.

```
- projectdir
  - xmake.lua
```

### Internal working directory mode (default)

This is the behaviour of the existing version, which has never changed.

#### The working directory is the project root directory by default.

```
- projectdir (workdir)
  - build (generated)
  - .xmake (generated)
  - src
  - xmake.lua
```

```console
$ cd projectdir
$ xmake
```

#### Set the other build directory outside the working directory

we can set the given build directory, but the working directory is still the project root directory.

```
- projectdir (workdir)
  - .xmake (generated)
  - src
  - xmake.lua
- build
```

```console
$ cd projectdir
$ xmake f -o ../build
```

### External working directory mode (new)

We can use `-P projectdir` to set the given project directory and it will enable external working directory mode.

Lock the current directory as the working directory, and xmake will cache the specified project directory.

#### The working directory is the current directory by default.

```
- workdir
  - build (generated)
  - .xmake (generated)
- projectdir
  - src 
  - xmake.lua
```

```console
$ cd workdir
$ xmake f -P ../projectdir
$ xmake
```

#### Set the other build directory outside the working directory

```
- build (generated)
- workdir
  - .xmake (generated)
- projectdir
  - src 
  - xmake.lua
```

```console
$ cd workdir
$ xmake f -o ../build
```

waruqi:
```
$ xmake update -s github:xmake-io/xmake#config
$ cd workdir
$ xmake f -P ../projectdir
$ xmake
$ xmake run
...
```

waruqi:
@davidchisnall you can try it, I have merged into dev branch.

davidxiaozhi:
帮忙测试验证
 xmake.lua 存在且能执行 xmake 构建
![image](https://user-images.githubusercontent.com/6194818/217593541-128d5cae-73ee-4e89-849f-c11a9741784f.png)

[...]

PR #3344, opened 2023-02-07
Improve to configure options and build directory

waruqi:
https://github.com/xmake-io/xmake/issues/3342


waruqi:
```
$ xmake update -s github:xmake-io/xmake#config
$ cd workdir
$ xmake f -P ../projectdir
$ xmake
...
```

davidchisnall:
Sorry, I'm coming to this very late, but I can confirm that this works beautifully.  This makes it much easier to use xmake because now I can have multiple parallel directories each containing a different build configuration and they do not interfere and they do not pollute my source tree with build artefacts.  Thank you very much for adding this!

PR #3348, opened 2023-02-08
Fix dependent packages loading when cross-compiling

SirLynix:
When configuring xmake for android from Linux with a binary package (plat=linux), its dependencies will by fetched for android.`

```lua
add_requires("nzsl~host, {binary = true})
add_requires("nzsl")
```
(nzsl depends on nazarautils)

I noticed this because xmake wanted me to reinstall the same binary package over and over again because its dependency hash didn't match.

I added a print here:
```lua
-- load all required packages
function _load_packages(requires, opt)

    -- no requires?
    if not requires or #requires == 0 then
        return {}
    end

    -- load packages
    local packages = {}
    local packages_nodeps = {}
    for _, requireitem in ipairs(load_requires(requires, opt.requires_extra, opt)) do

        -- load package
        local requireinfo = requireitem.info
        local requirepath = opt.requirepath and (opt.requirepath .. "." .. requireitem.name) or requireitem.name
        local package     = _load_package(requireitem.name, requireinfo, table.join(opt, {requirepath = requirepath}))
        print(package:name(), package:plat())
```

prints
```
nzsl linux
nazarautils android

...

nzsl android
nazarautils android
```

this PR fixes the issue by setting the plat/arch values of requires_extra for dependencies, I'm not sure it's the best fix but it works:

```
nzsl linux
nazarautils linux

...

nzsl android
nazarautils android
```

### DIFF
--- a/xmake/modules/private/action/require/impl/package.lua
+++ b/xmake/modules/private/action/require/impl/package.lua
@@ -883,8 +883,18 @@ function _load_packages(requires, opt)
 
                     -- load dependent packages and do not load system/3rd packages for package/deps()
                     local packagedeps = {}
+
+                    -- make sure dependent packages are loaded with the same plat/arch
+                    local requires_extraconf = package:extraconf("deps") or {}
+                    local requires_extra = {}
+                    for _, depname in pairs(package:get("deps")) do
+                        requires_extra[depname] = requires_extraconf[depname] or {}
+                        requires_extra[depname].plat = requires_extra[depname].plat or package:plat()
+                        requires_extra[depname].arch = requires_extra[depname].arch or package:arch()
+                    end
+
                     local deps, plaindeps = _load_packages(package:get("deps"), {requirepath = requirepath,
-                                                        requires_extra = package:extraconf("deps") or {},
+                                                        requires_extra = requires_extra,
                                                         parentinfo = requireinfo,
                                                         nodeps = opt.nodeps,
                                                         system = false})


waruqi:
no `{binary = true}`, it should be `{host = true}`

it should has warnings tips. like this

```
warning: add_requires("xxx") has unknown option: {binary=true}!
```

SirLynix:
sorry I put the wrong code from my memory, here's the real code I used:
```lua
if is_plat("android", "iphoneos", "wasm") then
	-- compile shaders using host-binaries
	if has_config("compile_shaders") then
		add_requires("nzsl~host", { kind = "binary" })
	end
end
```

it's a binary package, which is the same as host = true if I understand correctly.

Plus you can see that `package:plat()` returns "linux" and it gets "android" dependencies

waruqi:
> it's a binary package, which is the same as host = true if I understand correctly.

no, it's different, you should use `{host = true}`

```
add_requires("nzsl~host", { host = true })
```

`kind = "binary"` simply tells xmake to install the package as a binary, which is not a static or dynamic library, but it may also be a cross-compiled binary (even if perhaps it cannot be run on the host).

And `host = true` tells xmake to compile the package only as a host platform program, which may be a binary or a library, but it, and all its dependencies, will be forced to compile and run as a host program.

Translated with www.DeepL.com/Translator (free version)

SirLynix:
I tested with `add_requires("nzsl~host", { kind = "binary", host = true })` and it works as intended, however there still seem to be an issue as `kind = "binary"` made it so `nzsl~host` :plat() returns "linux" instead of "android", but this didn't affect dependent packages.

This is an inconstancy between the two behaviors.

waruqi:
It works for me.

```
add_requires("libpng", {system = false, kind = "binary", host = true})
target("testdep")
    set_kind("binary")
    add_files("src/*.cpp")
    add_packages("libpng")
```

```
ruki-2:testdep ruki$ xmake f -p android --ndk=~/files/android-ndk-r20b/ -c
checking for architecture ... armeabi-v7a
checking for Android SDK directory ... ~/Library/Android/sdk
checking for Build Tools Version of Android SDK ... 33.0.0
checking for NDK directory ... /Users/ruki/files/android-ndk-r20b
checking for SDK version of NDK ... 21
in xmake-repo:
  -> libpng v1.6.37 [macosx, x86_64, binary]. <---------------- macos, not android
please input: y (y/n/m)
```

SirLynix:
It works for me with both host = true and binary = true.

But if I use binary=true only, I still get a Linux binary (and not Android binary), with :kind() being "linux", however its dependency are loaded as android.

waruqi:
> It works for me with both host = true and binary = true.
> 
> But if I use binary=true only, I still get a Linux binary (and not Android binary), with :kind() being "linux", however its dependency are loaded as android.

Yes, I just made a bit of a mistake because the xmake-repo repository binary is mostly running on the host, so I forced it to gate as the host platform as well.

But setting kind = "binary" currently does not pass to its dependency packages. But host = true will make all dependencies set to the host platform as well.

waruqi:
Yes, this may be a bit hacky, and other types such as host binary may be extended later to improve it. But at the moment I haven't encountered a situation where I need to build a target host package, because usually it can't be run and used directly. So I didn't deal with it for now.

PR #3377, opened 2023-02-16
improve windows cross-compilation for meson

xq114:


xmake/modules/package/tools/meson.lua
```diff
@@ -54,7 +54,7 @@ function _is_cross_compilation(package)
     if package:is_plat("macosx") and not package:is_arch(os.subarch()) then
         return true
     end
-    if package:is_plat("windows") and package:is_arch("arm64") then
+    if package:is_plat("windows") and not package:is_arch(os.arch()) then
```
waruqi, line 57: 现在有 package:is_cross 了，应该可以直接用，你试试

https://github.com/xmake-io/xmake/blob/e4716ca9e43b1fc5b7eedbdef96f88b361153da6/xmake/core/package/package.lua#L625

PR #3387, opened 2023-02-18
improve find cmake package

star-hengxing:
- Improve component string order, see https://cmake.org/cmake/help/latest/command/find_package.html#id8
- When I print `opt`, I can't find `link_libraries = "Houdini"` and `search_mode = "config"`. Only `envs` pass into configs.
```lua
add_requires("cmake::Houdini", {alias = "houdini", system = true, configs = {
    search_mode = "config",
    link_libraries = "Houdini",
    envs = {CMAKE_PREFIX_PATH = "C:\\Program Files\\Side Effects Software\\Houdini 19.5.493\\toolkit\\cmake"}
}})
```
How can I fix it?

### DIFF
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -22,10 +22,12 @@
 function main()
     return
     {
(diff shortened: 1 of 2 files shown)


waruqi:
Fix what? You can debug it.

PR #3403, opened 2023-02-21
Fix ps1 install script

OpportunityLiu:
- Remove old completion
- Use latest completion from xmake-io/xmake repo (https://github.com/xmake-io/xmake-docs/blob/master/assets/scripts/pscompletions.text is never updated)


OpportunityLiu:
好像还有个 `scripts\profile-win.ps1`，是 `xmake update` 在用

PR #3423, opened 2023-02-25
Replace the ternary operator with the if/else statement

Someonesz:
In earlier versions of PowerShell, the ternary operator (?:) is not supported. Replace the ternary operator with the if/else statement

------

较早的 PowerShell 版本中，三目运算符（?:）不被支持。使用 if/else 语句替换三目运算符。


waruqi:
thanks

PR #3465, opened 2023-03-06
update find_vulkansdk.lua

45degree:
linux will not generate the two environment variables($VK_SDK_PATH, $VULKAN_SDK) when you install Vulkan from the system package manager(e.g. archlinux), so you should find the Vulkan path from /usr and /usr/local

### DIFF
--- a/xmake/modules/detect/sdks/find_vulkansdk.lua
+++ b/xmake/modules/detect/sdks/find_vulkansdk.lua
@@ -47,7 +47,7 @@ function main(opt)
     local libname = (is_host("windows") and "vulkan-1" or "vulkan")
     local libsuffix = ((is_host("windows") and arch == "x86") and "lib32" or "lib")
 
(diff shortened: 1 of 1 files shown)


xmake/modules/detect/sdks/find_vulkansdk.lua
```diff
@@ -47,6 +47,12 @@ function main(opt)
     local libname = (is_host("windows") and "vulkan-1" or "vulkan")
     local libsuffix = ((is_host("windows") and arch == "x86") and "lib32" or "lib")
 
+    if is_host("linux") then
+        -- we attempt to find vulkan from /usr, e.g. /usr/include/vulkan/vulkan.h
+        table.insert(paths, "/usr");
+        table.insert(paths, "/usr/local")
+    end
```
waruqi, line 54: paths 是在上面设置的，放到上面位置，不要分散。

PR #3473, opened 2023-03-07
fix xmake doxygen bug，add doxygen to function _generate_doxyfile()

guijiyang:
xmake doxygen can't exexcute properly
with output below:
```
xmake doxygen -vD
error: @programdir\core\main.lua:302: @programdir\plugins\doxygen\main.lua:34: attempt to index a nil value (global 'doxygen')
```
_generate_doxyfile function block use doxygen parameter which need to be passed outside


waruqi:
commit 用英文

waruqi:
thanks

PR #3493, opened 2023-03-10
extra include_directories and debug flags for cmake::find_package

### LINKED ISSUE #3492: add an optional internal project name for cmake find_package

I'm using xmake to find an external third-party with cmake suppot. The library is called `deal.II` but within the `deal.IIConfig.cmake`  file, all variable names start with `DEAL_II` like:

```cmake
SET(DEAL_II_PACKAGE_NAME "deal.II")
SET(DEAL_II_INCLUDE_DIRS ...)
```

However the offered `find_package` module could not detect the actual link and include path
https://github.com/xmake-io/xmake/blob/43908d7c753b6502cf1f2f589cbc52b508738347/xmake/modules/package/manager/cmake/find_package.lua#L84-L93

If possible, please consider to add an extra optional `config_name` to `add_requires`  interface. And the `_find_package` function should be:

```lua
local config_name = opt.config_name or name
cmakefile:print("   target_include_directories(%s PRIVATE ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS})", 
    testname, config_name, config_name) 
cmakefile:print("   target_include_directories(%s PRIVATE ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS})", 
    testname, config_name:upper(), config_name:upper())
```

And the use of the interface should look like:

```lua
-- library that use an extra config name
add_requires("cmake::deal.II", {configs = {system = true, config_name = "DEAL_II"}})
-- library that work as before
add_requires("cmake::ZLIB")
```

My current workaround is to create a `deal_II` directory and copy all `deal.IIConfig...` files, rename them as `deal_IIConfig...`. While using xmake, use `deal_II` as the library name:

```
add_requires("cmake::deal_II")
```

### Additional context

_No response_

RichardYCJ:
Once merge, this PR should close #3492. 

# About `include_directories`
Since cmake has no convention on variables that saves the include directories and link libraries, some libraries have different variables rather than `XXX_INCLUDE_DIRS` or `XXX_LIBRARIES` defined within `XXXConfig.cmake`. To apply to this situation, use:

```lua
-- search for library deal.II defined with DEAL_II_INCLUDE_DIRS / DEAL_II_LIBRARIES
add_requires("cmake::deal.II", {system = true,
    configs = {include_directories = "${DEAL_II_INCLUDE_DIRS}",
               link_libraries = "${DEAL_II_LIBRARIES}"}})

-- targets are also supported
add_requires("cmake::deal.II", {system = true,
    configs = {include_directories = "${DEAL_II_INCLUDE_DIRS}",
               link_libraries = "deal_II"}})
```

About `debug`
Set to true if want to find package under `CMAKE_BUILD_TYPE=Debug`. Could use to spread the debug config automatically according to xmake mode, use like:

```lua
add_requires("cmake::ZLIB", {debug = is_mode("debug")})

-- add it into configs is also ok
add_requires("cmake::ZLIB", {configs = {debug = is_mode("debug")}})
```

### DIFF
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -23,12 +23,12 @@ function main()
     return
     {
         link_libraries = {description = "Set the cmake package dependencies, e.g. {\"abc::lib1\", \"abc::lib2\"}"},
+        include_dirs   = {description = "Set the cmake package include directories, e.g. {\"${ZLIB_INCLUDE_DIRS}\"}"},
         search_mode    = {description = "Set the cmake package search mode, e.g. {\"config\", \"module\"}"},
         components     = {description = "Set the cmake package components, e.g. {\"regex\", \"system\"}"},
         moduledirs     = {description = "Set the cmake modules directories."},
         presets        = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs           = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
-        find_name      = {description = "Set the name used for ${find_name}_LIBS/_INCLUDE_DIRS in cmake"},
         debug          = {description = "Set the CMAKE_BUILD_TYPE to Debug if set to true"},
     }
 end
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -29,7 +29,6 @@ function main()
         moduledirs          = {description = "Set the cmake modules directories."},
         presets             = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs                = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
-        debug               = {description = "Set the CMAKE_BUILD_TYPE=Debug if set to ture."}
     }
 end
 
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -29,7 +29,6 @@ function main()
         moduledirs     = {description = "Set the cmake modules directories."},
         presets        = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs           = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
-        debug          = {description = "Set the CMAKE_BUILD_TYPE to Debug if set to true"},
     }
 end
 
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -29,6 +29,7 @@ function main()
         presets        = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs           = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
         find_name      = {description = "Set the name used for ${find_name}_LIBS/_INCLUDE_DIRS in cmake"},
+        debug          = {description = "Set the CMAKE_BUILD_TYPE to Debug if set to true"},
     }
 end
 
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -28,6 +28,7 @@ function main()
         moduledirs     = {description = "Set the cmake modules directories."},
         presets        = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs           = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
+        find_name      = {description = "Set the name used for ${find_name}_LIBS/_INCLUDE_DIRS in cmake"},
     }
 end
 
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -29,6 +29,7 @@ function main()
         moduledirs          = {description = "Set the cmake modules directories."},
         presets             = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs                = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
+        debug               = {description = "Set the CMAKE_BUILD_TYPE=Debug if set to ture."}
     }
 end
 
--- a/xmake/modules/package/manager/cmake/configurations.lua
+++ b/xmake/modules/package/manager/cmake/configurations.lua
@@ -22,13 +22,13 @@
 function main()
     return
     {
-        link_libraries = {description = "Set the cmake package dependencies, e.g. {\"abc::lib1\", \"abc::lib2\"}"},
-        include_dirs   = {description = "Set the cmake package include directories, e.g. {\"${ZLIB_INCLUDE_DIRS}\"}"},
-        search_mode    = {description = "Set the cmake package search mode, e.g. {\"config\", \"module\"}"},
-        components     = {description = "Set the cmake package components, e.g. {\"regex\", \"system\"}"},
-        moduledirs     = {description = "Set the cmake modules directories."},
-        presets        = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
-        envs           = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
+        link_libraries      = {description = "Set the cmake package dependencies, e.g. {\"abc::lib1\", \"abc::lib2\"}"},
+        include_directories = {description = "Set the cmake package include directories, e.g. {\"${ZLIB_INCLUDE_DIRS}\"}"},
+        search_mode         = {description = "Set the cmake package search mode, e.g. {\"config\", \"module\"}"},
+        components          = {description = "Set the cmake package components, e.g. {\"regex\", \"system\"}"},
+        moduledirs          = {description = "Set the cmake modules directories."},
+        presets             = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
+        envs                = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
     }
 end
 
--- a/xmake/modules/package/manager/cmake/find_package.lua
+++ b/xmake/modules/package/manager/cmake/find_package.lua
@@ -107,12 +107,9 @@ function _find_package(cmake, name, opt)
 
     -- run cmake
     local envs = configs.envs or opt.envs
-    if (configs.debug) then
-        if (not envs) then
-            envs = {CMAKE_BUILD_TYPE = "Debug"}
-        elseif (not envs.CMAKE_BUILD_TYPE) then
-            envs.CMAKE_BUILD_TYPE = "Debug"
-        end
+    if opt.mode == "debug" then
+        envs = envs or {}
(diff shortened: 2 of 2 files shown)


xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -78,19 +78,20 @@ function _find_package(cmake, name, opt)
...
+    local findname = configs.find_name or name
     cmakefile:print("find_package(%s REQUIRED %s)", requirestr, componentstr)
     cmakefile:print("if(%s_FOUND)", name)
     cmakefile:print("   add_executable(%s test.cpp)", testname)
     cmakefile:print("   target_include_directories(%s PRIVATE ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS})",
-        testname, name, name)
+        testname, findname, findname)
     cmakefile:print("   target_include_directories(%s PRIVATE ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS})",
-        testname, name:upper(), name:upper())
+        testname, findname:upper(), findname:upper())
```
SirLynix, line 88: I think we should keep `configs.find_name` as is, instead of calling upper() on it, some packages may have lowercase find_name.

RichardYCJ: we actually check both findname (line 86) and findname:upper() (line 88), this is for `findname = name` but the library have all uppercase variables.

RichardYCJ:
I finally take @waruqi's scheme. It is simple to use and need not to consider too much edge cases. 

For packages who needs to change to debug 
```lua
add_requires("cmake::zlib", {configs = {debug = is_mode("debug")})
```

For other build types use:
```lua
add_requires("cmake::zlib", {configs = {envs = {CMAKE_BUILD_TYPE = "RelWithDebInfo"}}})
```

This flag respects the `envs.CMAKE_BUILD_TYPE`, the default does nothing just as before.

xmake/modules/package/manager/cmake/configurations.lua
```diff
@@ -28,6 +28,8 @@ function main()
         moduledirs     = {description = "Set the cmake modules directories."},
         presets        = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs           = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
+        find_name      = {description = "Set the name used for ${find_name}_LIBS/_INCLUDE_DIRS in cmake"},
+        debug          = {description = "Set the CMAKE_BUILD_TYPE to Debug if set to true"},
```
waruqi, line 32: debug is builtin configuration, we need not define it. and you can get `opt.mode == debug` directly

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -100,6 +101,10 @@ function _find_package(cmake, name, opt)
 
     -- run cmake
     local envs = configs.envs or opt.envs
+    -- decide whether to enable debug build type
+    if (not envs.CMAKE_BUILD_TYPE and configs.debug) then
```
waruqi, line 105: use `opt.mode == "debug"`

xmake/modules/package/manager/cmake/configurations.lua
RichardYCJ: this flag should actually be called as `spread_debug`. Once set, it will find debug version of external library if xmake is under debug mode.

Do we need this flag, or we just set `CMAKE_BUILD_TYPE` to `Debug` automatically internally?

xmake/modules/package/manager/cmake/find_package.lua
RichardYCJ: This is done. But `opt.mode` seems always to be `release`

xmake/modules/package/manager/cmake/configurations.lua
RichardYCJ: I end up removing `debug` flags, since `CMKAE_BUILD_TYPE` is a global variable and will spread to all linked libraries. This is consistent with cmake behavior.

waruqi: you need pass to configs.debug

```
add_requires("cmake::ZLIB", {configs = {debug = true}})
```

```
{
  require_version = "latest",
  mode = "debug",   <------- debug
  configs = {
    debug = true,
    shared = false,
    pic = true
  },
  arch = "x86_64",
```

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -81,26 +81,37 @@ function _find_package(cmake, name, opt)
...
-    cmakefile:print("   target_include_directories(%s PRIVATE ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS})",
-        testname, name, name)
-    cmakefile:print("   target_include_directories(%s PRIVATE ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS})",
-        testname, name:upper(), name:upper())
+    -- setup include directories
+    local includedirs = ""
+    if configs.include_dirs then
+        includedirs = table.concat(table.wrap(configs.include_dirs), " ")
+    else
+        includedirs = string.gsub("${X_INCLUDE_DIR} ${X_INCLUDE_DIRS}", "X", name)
```
waruqi, line 89: please use string.format

```
includedirs = ("${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS}")format(name, name)
```

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -81,26 +81,37 @@ function _find_package(cmake, name, opt)
...
+        includedirs = includedirs .. string.gsub(" ${X_INCLUDE_DIR} ${X_INCLUDE_DIRS}", "X", name:upper())
```
waruqi, line 90:
```
includedirs = ("${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS} ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS}"):format(name, name, name:upper(), name:upper())
```

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -81,26 +81,37 @@ function _find_package(cmake, name, opt)
...
+        linklibs = linklibs .. string.gsub(" ${X_LIBRARY} ${X_LIBRARIES} ${X_LIBS}", "X", name:upper())
     end
+    cmakefile:print("   target_link_libraries(%s PRIVATE %s)", testname, linklibs)
     cmakefile:print("endif(%s_FOUND)", name)
     cmakefile:close()
 
     -- run cmake
-    local envs = configs.envs or opt.envs
-    try {function() return os.vrunv(cmake.program, {workdir}, {curdir = workdir, envs = envs}) end}
+    local envs = configs.envs or opt.envs or {}
```
waruqi, line 109: if envs is nil, you cannot set `{}`, it will override all current envs in os.vrunv().

xmake/modules/package/manager/cmake/configurations.lua
```diff
@@ -23,6 +23,7 @@ function main()
     return
     {
         link_libraries = {description = "Set the cmake package dependencies, e.g. {\"abc::lib1\", \"abc::lib2\"}"},
+        include_dirs   = {description = "Set the cmake package include directories, e.g. {\"${ZLIB_INCLUDE_DIRS}\"}"},
```
waruqi, line 26: please use `include_directories` Consistent with the cmake interface.

RichardYCJ: I tried but the variable is too long and will break the alignment, change all?

xmake/modules/package/manager/cmake/find_package.lua
RichardYCJ: ok, revert this

xmake/modules/package/manager/cmake/configurations.lua
waruqi: change all

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -81,25 +81,36 @@ function _find_package(cmake, name, opt)
...
+        linklibs = linklibs .. string.gsub(" ${X_LIBRARY} ${X_LIBRARIES} ${X_LIBS}", "X", name:upper())
     end
+    cmakefile:print("   target_link_libraries(%s PRIVATE %s)", testname, linklibs)
     cmakefile:print("endif(%s_FOUND)", name)
     cmakefile:close()
 
     -- run cmake
     local envs = configs.envs or opt.envs
+    -- decide whether to enable debug build type
+    if (not envs.CMAKE_BUILD_TYPE and opt.mode == "debug") then
```
waruqi, line 111:
```
if opt.mode == "debug" then
    envs = envs or {}
    envs.CMAKE_BUILD_TYPE = envs.CMAKE_BUILD_TYPE or "Debug"
end
```

RichardYCJ: remove this since #3497 is solved.

RichardYCJ: fixed

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -81,25 +81,39 @@ function _find_package(cmake, name, opt)
...
+        linklibs = ("${%s_LIBRARY} ${%s_LIBRARIES} ${%s_LIBS}"):format(name, name, name)
+        linklibs = linklibs .. (" ${%s_LIBRARY} ${%s_LIBRARIES} ${%s_LIBS}"):format(name:upper(), name:upper(), name:upper())
     end
+    cmakefile:print("   target_link_libraries(%s PRIVATE %s)", testname, linklibs)
     cmakefile:print("endif(%s_FOUND)", name)
     cmakefile:close()
 
     -- run cmake
     local envs = configs.envs or opt.envs
+    if (configs.debug) then
```
waruqi, line 110: you should use 
```
if opt.mode == "debug"
```

because `add_requires("xxx", {debug = true})` is also supported.

waruqi: and remove `()`

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -81,25 +81,39 @@ function _find_package(cmake, name, opt)
...
+        if (not envs) then
+            envs = {CMAKE_BUILD_TYPE = "Debug"}
+        elseif (not envs.CMAKE_BUILD_TYPE) then
+            envs.CMAKE_BUILD_TYPE = "Debug"
+        end
```
waruqi, line 115: just only two lines

```
envs = envs or {}
envs.CMAKE_BUILD_TYPE = envs.CMAKE_BUILD_TYPE or "Debug"
```

RichardYCJ: It confuses me, earlier you said set `envs` to `{}` would overwrite the whole envs (see [this](https://github.com/xmake-io/xmake/pull/3493/files/001e4781aaaea799f53a4ab76903ba3930ad1b24#diff-da48836594df046b6b3465cf6d19e24de62a47e2b978f210d246842706506922R109)). Only a `nil` or a table with value is accepted?

waruqi: envs has always values if it's debug mode, so we can set `envs = envs or {}`, because it is always non-empty.

but if it's release mode, envs cannot be `{}`

xmake/modules/package/manager/cmake/find_package.lua
RichardYCJ: It seems that `opt.configs.debug` can be set automatically according to outer `opt.mode`, if so, should we recommand using just `{debug = true}` and delete it from `configurations.lua`?

waruqi: They are equivalent and you can use them all. And you have removed it from the configuration

RichardYCJ:
Sorry it takes so long to implement this simple feature in such a mess 😢. I'm new to xmake and Lua and not familiar with the internal variable transfer and conventions. Thanks for the patience and guidance from @waruqi! 🎉

waruqi:
thanks

PR #3555, opened 2023-03-23
support different qt versions by --qmake config

EvinceMoi:
add `--qmake` config option to specifiy Qt SDK, in case you have multiple version installed in system like ArchLinux

for example:
```sh 
# this will find the default version(qt5)
xmake f --qmake=/usr/bin/qmake  
```

or 
```sh
# this will find qt version 6
xmake f --qmake=/usr/bin/qmake6
```

### DIFF
--- a/xmake/modules/detect/sdks/find_qt.lua
+++ b/xmake/modules/detect/sdks/find_qt.lua
@@ -177,6 +177,10 @@ function _find_qt(sdkdir, sdkver)
         return
     end
 
+    return _find_qt_by_qmake(qmake)
+end
+
+function _find_qt_by_qmake(qmake)
     -- get qt environments
     local qtenvs = _get_qtenvs(qmake)
     if not qtenvs then
@@ -241,7 +245,12 @@ function main(sdkdir, opt)
     end
 
     -- find qt
-    local qt = _find_qt(sdkdir or config.get("qt") or global.get("qt") or config.get("sdk"), opt.version or config.get("qt_sdkver"))
+    local qt
+    if config.get("qmake") then
+        qt = _find_qt_by_qmake(config.get("qmake"))
+    else
+        qt = _find_qt(sdkdir or config.get("qt") or global.get("qt") or config.get("sdk"), opt.version or config.get("qt_sdkver"))
+    end
     if qt then
 
         -- save to config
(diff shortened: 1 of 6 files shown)


EvinceMoi:
only tested on ArchLinux, as the `find_qt` works by extracting info from `qmake -query`, i think other platforms may work too

waruqi:
please do not add `--qmake=`, we just need use `xmake f --qt_sdkver=6.0` to improve to find_qt/qmake.

```
xmake f --qt_sdkver=6.0
```

it will attempt to find qmake6 first in find_qt.

EvinceMoi:
I don't think this is a good idea, as the name of `qmake` may vary from different distro or different version of qt. 

For example, the current default version of qt in Arch is qt5, so the name `qmake` is attached to qt5, and `qmake6` for qt6. Some day in the future, the `qmake` may be assigned to qt6, and there will be a new `qmake7` for qt7.

My point is as `qmake -query` containing all the necessary infos of qt sdk, and internally xmake use it to do actually `find_qt`, it is clear to provide specific `qmake` rather than trying to find it in sdkdir

waruqi:
I wouldn't consider adding the `--qmake` parameter for now.

There are already so many configuration parameters in xmake config that it would be difficult to maintain and for users to know which ones to use if they were to add new ones to improve the lookup of each sdk.

Also, the existing `--qt_sdkver=` parameter would conflict with `--qmake=`.

like this

```
xmake f --qt_sdkver=7.0 --qmake=/usr/bin/qmake6
```

it's invalid configuration. For users, should I use `--qt_sdkver=6.0` or `--qmake=/usr/bin/qmake6` to select the qt version? He has to look through the documentation and keep trying to get it right.

> For example, the current default version of qt in Arch is qt5, so the name qmake is attached to qt5, and qmake6 for qt6. Some day in the future, the qmake may be assigned to qt6, and there will be a new qmake7 for qt7.

This is not a problem and we are perfectly adaptive to deal with them. like this 

```
local sdkver_major = semver.new(sdkver):major()
qmake = find_tool("qmake" .. sdkver_major, {paths = sdkdir and path.join(sdkdir, "bin")})
```

waruqi:
I have improved it to find qt6. https://github.com/xmake-io/xmake/commit/2b26ccfe4deeccdf73d405254bb81becbefc7da8

please run `xmake update -s dev` to update it.

```console
xmake f --qt_sdkver=6.0 -c
xmake
```

And if we only installed qmake6, did not install qmake.

xmake will also find qmake6 automatically.

```
$ xmake f -c
$ xmake
checking for platform ... linux
checking for Qt SDK directory ... /usr
checking for Qt SDK version ... 6.3.1
[ 14%]: compiling.qt.ui src/mainwindow.ui
[ 28%]: compiling.qt.moc src/mainwindow.h
[ 42%]: cache compiling.release src/main.cpp
[ 42%]: cache compiling.release src/mainwindow.cpp
[ 71%]: linking.release demo
[100%]: build ok, spent 4.56s
```

PR #3563, opened 2023-03-26
Improve d projects

pdagobert:
This PR fixes two bugs in D projects, dub packages are not working on windows and pdb are not generated in debug builds. I have also added a test for dub packages.

I have a few questions : 
- The original D tests (eg dlang/console) are skipped on many platforms, any idea why?
- Maybe it'd be worth adding a test for checking pdb generation? I can add that if you think it's worth it, but I'd need a bit of direction on how to a achieve that. 
- I haven't fix pdb with gdc. I suppose the fix is the same as with ldc, but since I can't test it (gdc won't work on windows) I didn't include it. Should I?
- I noticed that in release the boundscheck:off flag is enabled. This flag trades correctness against performances. I consider this flag just lik fastmath, something that should be enabled explicitly, rathan than by default.. Especially since it's not the case with dub. What is the rational behind that?

Thanks.

### DIFF
--- a/xmake/modules/core/tools/dmd.lua
+++ b/xmake/modules/core/tools/dmd.lua
@@ -83,6 +83,8 @@ function nf_symbol(self, level)
             _g.symbol_maps = maps
         end
         return maps[level .. '_' .. kind] or maps[level]
+    elseif kind == "dcld" then
+        return "-g"
     end
 end
 
--- a/xmake/modules/core/tools/dmd.lua
+++ b/xmake/modules/core/tools/dmd.lua
@@ -72,7 +72,7 @@ function nf_strip(self, level)
 end
 
 -- make the symbol flag
-function nf_symbol(self, level)
+function nf_symbol(self, level, target)
     local kind = self:kind()
     if language.sourcekinds()[kind] then
         local maps = _g.symbol_maps
@@ -83,7 +83,7 @@ function nf_symbol(self, level)
             _g.symbol_maps = maps
         end
         return maps[level .. '_' .. kind] or maps[level]
-    elseif kind == "dcld" then
+    elseif (kind == "dcld" or kind == "dcsh") and target:is_plat("windows") and level == "debug" then
         return "-g"
     end
 end
--- a/xmake/modules/core/tools/ldc2.lua
+++ b/xmake/modules/core/tools/ldc2.lua
@@ -49,7 +49,7 @@ function nf_optimize(self, level)
 end
 
 -- make the symbol flag
-function nf_symbol(self, level)
+function nf_symbol(self, level, target)
     local kind = self:kind()
     if language.sourcekinds()[kind] then
         local maps = _g.symbol_maps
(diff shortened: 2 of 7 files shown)


xmake/modules/core/tools/dmd.lua
```diff
@@ -83,6 +83,8 @@ function nf_symbol(self, level)
             _g.symbol_maps = maps
         end
         return maps[level .. '_' .. kind] or maps[level]
+    elseif kind == "dcld" then
+        return "-g"
```
waruqi, line 87: if should be only for windows (pdb) and you need add `if level == "debug" ` or use map `{debug = "-g"}`

```
function nf_symbol(self, level, target)
    local kind = self:kind()
    if language.sourcekinds()[kind] then
        local maps = _g.symbol_maps
        if not maps then
            maps = {
                debug  = {"-g", "-debug"}
            }
            _g.symbol_maps = maps
        end
        return maps[level .. '_' .. kind] or maps[level]
    elseif (kind == "dcld" or kind == "dcsh") and target:is_plat("windows") and level == "debug" then
        return "-g"
    end
end

```

xmake/modules/core/tools/ldc2.lua
```diff
@@ -61,5 +61,7 @@ function nf_symbol(self, level)
             _g.symbol_maps = maps
         end
         return maps[level .. '_' .. kind] or maps[level]
+    elseif kind == "dcld" then
+        return "-g"
```
waruqi, line 65: and here

waruqi:
> This PR fixes two bugs in D projects, dub packages are not working on windows and pdb are not generated in debug builds. I have also added a test for dub packages.
> 
> I have a few questions :
> 
> * The original D tests (eg dlang/console) are skipped on many platforms, any idea why?

Too many tests can make ci tests too slow. xmake currently focuses on testing c++-related features, while other language features are tested on some of the faster ci platforms.

For example, the freebsd ci is very slow and unstable, so we can only do basic tests on it, otherwise it may affect our daily development and maintenance schedule.

> * Maybe it'd be worth adding a test for checking pdb generation? I can add that if you think it's worth it, but I'd need a bit of direction on how to a achieve that.

It is not possible to do all the testing on the ci, some optional features are tested and often only local verification is required.

> * I haven't fix pdb with gdc. I suppose the fix is the same as with ldc, but since I can't test it (gdc won't work on windows) I didn't include it. Should I?

If you can't test it yet, don't include it.

> * I noticed that in release the boundscheck:off flag is enabled. This flag trades correctness against performances. I consider this flag just lik fastmath, something that should be enabled explicitly, rathan than by default.. Especially since it's not the case with dub. What is the rational behind that?


But reducing some of the detection should also improve performance and reduce some of the binary size.

You can remove it in fastest, but I think smallest/aggressive should keep it.

If you don't want to turn it on, you can use `set_optimize("faster")`

pdagobert:
Ah, makes sense. I changed the dub test to match the console test then.

Thanks for clarifying set_optimize, I didn't realize release was the equivalent of fastest, I thought it was faster. And since fastest enables fastmath, having boundscheck:off makes sense I guess. I'll use set_optimize("faster").

xmake/modules/core/tools/dmd.lua
```diff
@@ -83,6 +83,8 @@ function nf_symbol(self, level)
             _g.symbol_maps = maps
```
waruqi, line 83: you need add `target` argument

```
function nf_symbol(self, level, target)
```

waruqi:
dub is not installed on macos ci

PR #3664, opened 2023-04-21
Modify README(_CN).md examples style

laper32:
IIRC, following writing style are same:
```lua

-- Old
target("console")
    set_kind("binary")
    add_files("src/*.c")
    if is_mode("debug") then
        add_defines("DEBUG")
    end

-- New
target("console", 
  function() 
    set_kind("binary")
    add_files("src/*.c")
    if is_mode("debug") then
        add_defines("DEBUG")
    end
  end
)
```

Since new writing style is better human-reading, should we convert all existing writing style to new if possible?

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,24 +371,32 @@ For more details see: [Remote Compilation](https://xmake.io/#/features/remote_bu
 ```lua
 add_rules("mode.debug", "mode.release")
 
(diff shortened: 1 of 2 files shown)


waruqi:
You can use new style, but please do not modify it.

PR #3738, opened 2023-05-12
Fix librarydeps.compatible_checked

SirLynix:
When setting package:data_set("librarydeps.compatible_checked", true), `_compatible_with_previous_librarydeps` returns nil instead of true

### DIFF
--- a/xmake/modules/private/action/require/impl/package.lua
+++ b/xmake/modules/private/action/require/impl/package.lua
@@ -995,7 +995,7 @@ function _compatible_with_previous_librarydeps(package, opt)
     -- has been checked?
     local compatible_checked = package:data("librarydeps.compatible_checked")
     if compatible_checked then
-        return
+        return true
     end
 
     -- check strict compatibility for librarydeps?


xmake/modules/private/action/require/impl/package.lua
```diff
@@ -995,7 +995,7 @@ function _compatible_with_previous_librarydeps(package, opt)
     -- has been checked?
     local compatible_checked = package:data("librarydeps.compatible_checked")
```
waruqi, line 996: ohmm. I've forgotten why I judged it in the first place.

I can't find where it was set. librarydeps.compatible_checked is supposed to use caching to speed up compatibility checking.

package:data is a private data interface and should not be set by the user.

waruqi: I don't think it's doing anything, at least for now, and it should be removed.

SirLynix: I'm using it at the end of a on_fetch callback.
```lua

    on_fetch(function (package, opt)
        local nazara = os.getenv("NAZARA_ENGINE_PATH")
        if not nazara or not os.isdir(nazara) then
            return
        end

-- ...

        package:data_set("librarydeps.compatible_checked", true)

        return fetchInfo
    end)
```

because otherwise, xmake wants to install the package even though it found it because its dependencies hash don't match (which is not a problem here). We already discussed this here: https://github.com/xmake-io/xmake/issues/2719 and I ran into this issue again, with no solution. I though setting `"librarydeps.compatible_checked"` would help.

waruqi: There is no mention of using package:data("librarydeps.compatible_checked") to bypass it. As I recall, there was no good discussion of a solution.

Also, package:data("librarydeps.compatible_checked") is a private interface, and calling it directly by the user would be very hacky and difficult to maintain later. As it is, I've forgotten what it does.

SirLynix: It looks like it was introduced here: https://github.com/xmake-io/xmake/commit/2a8af8db1d8b6e05f2a751a3320a1c45596ac0e5 with no purpose (probably a leftover of another implementation).

So yeah, I think it should be removed. However I still have no way to disable strict dependency checking with on_fetch, which makes `on_fetch` useless when you use strict dependencies.

SirLynix: The issue is that my dependent packages use `set_policy("package.strict_compatibility", true)` as well

waruqi: you can modify here

```diff
    local strict_compatibility = project.policy("package.librarydeps.strict_compatibility")
    if strict_compatibility == nil then
        strict_compatibility = package:policy("package.librarydeps.strict_compatibility")
    end
+    if strict_compatibility == false then
+       return true
+    end
```

waruqi: I removed it and add this check. you can try it. https://github.com/xmake-io/xmake/commit/34eb729eb862e7415028c84970f3b2c830d95cd9

and you can add new pr to improve it.

PR #3739, opened 2023-05-13
improve to install modules #3730

waruqi:
https://github.com/xmake-io/xmake/issues/3730


waruqi:
@Arthapz I have improved the file structure of the module installation so that it can support module files of the same name in different directories.

Arthapz:
Okay nice :)

PR #3741, opened 2023-05-14
fix: when the incoming targetname == '.', the name of the current dir `.`

fwqaaq:
When  the incoming targetname == `.`, the name of the current dir `.`.

I hope that it can be generated directly in the current dir and I have made some minor modifitications. But the name `targetname` doesn't seem to be good.

### DIFF
--- a/xmake/actions/create/main.lua
+++ b/xmake/actions/create/main.lua
@@ -68,9 +68,11 @@ function _create_project(language, templateid, targetname)
     end
     assert(tempinst and tempinst:scriptdir(), "invalid template id: %s!", templateid)
 
-    -- get project directory
-    local projectdir = path.absolute(option.get("project") or path.join(os.curdir(), targetname))
-    if not os.isdir(projectdir) then
+    -- get project directory, if targetname == '.', then get path ot the current dir
+    local projectdir = path.absolute(option.get("project") or
+        path.join(os.curdir(), targetname == "." and "" or targetname))
+    print(projectdir)
+    if not os.isdir(projectdir) and targetname ~= "." then
         -- make the project directory if not exists
         os.mkdir(projectdir)
     end
@@ -104,6 +106,11 @@ function _create_project(language, templateid, targetname)
         raise("template(%s): project not found!", templateid)
     end
 
+    -- if targetname == ".", get the name of the current directory
+    if targetname == "." then
+        targetname = path.filename(projectdir)
+    end
+
     -- get the builtin variables
     local builtinvars = _get_builtinvars(tempinst, targetname)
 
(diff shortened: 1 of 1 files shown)


xmake/actions/create/main.lua
```diff
@@ -68,9 +68,11 @@ function _create_project(language, templateid, targetname)
     end
     assert(tempinst and tempinst:scriptdir(), "invalid template id: %s!", templateid)
 
-    -- get project directory
-    local projectdir = path.absolute(option.get("project") or path.join(os.curdir(), targetname))
-    if not os.isdir(projectdir) then
+    -- get project directory, if targetname == '.', then get path ot the current dir
+    local projectdir = path.absolute(option.get("project") or
+        path.join(os.curdir(), targetname == "." and "" or targetname))
+    print(projectdir)
```
fwqaaq, line 74:
```suggestion
    local projectdir = path.absolute(option.get("project") or
        path.join(os.curdir(), targetname == "." and "" or targetname))
```

waruqi:
we should not pass `.` to target name.

Please use `xmake create -P .`

fwqaaq:
> we should not pass . to target name.

But if I use `xmake create -l c .`, it has no problem and generates a project with a target of `.`.

So I think it is necessary to add this logic or error handling.

waruqi:
> > we should not pass . to target name.
> 
> But if I use `xmake create -l c .`, it has no problem and generates a project with a target of `.`.
> 
> So I think it is necessary to add this logic or error handling.

`.` is invalid target name, you should use `xmake create -l c -P .`

you can add some check error for it.

fwqaaq:
> . is invalid target name, you should use `xmake create -l c -P .`

But if you use `xmake create -l c xxx` that is Ok, emmm..., what should we do about this?

waruqi:
> > . is invalid target name, you should use `xmake create -l c -P .`
> 
> But if you use `xmake create -l c xxx` that is Ok, emmm..., what should we do about this?

If you want to create project files in current directory. you can use `-P .`

otherwise it will create sub-directory.   It has nothing to do with whether the target name is changed or not. You should not set targetname as .

fwqaaq:
>otherwise it will create sub-directory. 

Not so. It just a problem with the name of the project. Just like this:

```bash
mkdir test && cd test
xmake create -l c .
```

```lua
target(".")
    set_kind("binary")
    add_files("src/*.c")
```

waruqi:
> > > `.` is invalid target name, you should use `xmake create -l c -P .`
> > > you can add some check error for it.
> 
> You closed this pr 😢

If you want to add error check, you can open a new pr.

PR #3821, opened 2023-06-06
Long Path checkbox

A2va:
Sometimes when xmake is building a package on Windows, it fails because the path is too long to handle. This issue can also occur when installing a package from a repository with a lot of submodules. Therefore, this PR attempts to mitigate the problem by providing a setting in the NSIS installer.

The NSIS configuration file is modified to:

* Enable the Long Path option in the registry.
* Execute a git command to enable the long path option.

Unfortunately, the execution of git opens a terminal for a short time (~100ms). I did some research to remove this behavior, but it seems that the only way is to use nsExec plugin which is already present in the codebase but not enabled by default. Therefore, if you prefer to have a silent execution, you would have to compile the plugin first (That's why I have set this PR in draft mode).

### DIFF
--- a/scripts/installer.nsi
+++ b/scripts/installer.nsi
@@ -285,6 +285,19 @@ Section "XMake (required)" InstallExeutable
 
 SectionEnd
 
+Section "Enable Long Path" LongPath
+  ; Get git path from the registry
+  ReadRegStr $R0 ${HKLM} "SOFTWARE\GitForWindows" "InstallPath"
+  ${If} $NOADMIN == "false"
+    ; In admin enable long path on both git system config and in registry
+    WriteRegDWORD ${HKLM} "SYSTEM\CurrentControlSet\Control\FileSystem" "LongPathsEnabled" 1
+    Exec '"$R0\cmd\git.exe" config --system core.longpaths true'
+  ${Else}
+    ; Enable only in git global config
+    Exec '"$R0\cmd\git.exe" config --global core.longpaths true'
+  ${EndIf}
+SectionEnd
+
 Section "Add to PATH" InstallPath
 
   ${If} $NOADMIN == "false"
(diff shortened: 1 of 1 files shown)


scripts/installer.nsi
```diff
@@ -285,6 +285,19 @@ Section "XMake (required)" InstallExeutable
 
 SectionEnd
 
+Section "Enable Long Path" LongPath
+  ; Get git path from the registry
+  ReadRegStr $R0 ${HKLM} "SOFTWARE\GitForWindows" "InstallPath"
+  ${If} $NOADMIN == "false"
+    ; In admin enable long path on both git system config and in registry
+    WriteRegDWORD ${HKLM} "SYSTEM\CurrentControlSet\Control\FileSystem" "LongPathsEnabled" 1
+    Exec '"$R0\cmd\git.exe" config --system core.longpaths true'
```
waruqi, line 294: If git is not installed on the user's system, or if the installation path is not cmd\git.exe, the installation may fail.

And, xmake may install git in its own package installation directory.

waruqi: Only for installing the xmake-repo package does git need to support long paths, but for some packages that need to handle long paths, we will turn on long paths to support it. 

https://github.com/xmake-io/xmake-repo/blob/e1581652ef8cdfb3605d341eebea240a05a580f9/packages/l/libtorch/xmake.lua#L45

So I don't think git's setting of longpath is necessary, it's automatically supported by xmake when installing packages.

waruqi: Also, this setting may have an effect on other programs and it is recommended that more detailed descriptive information be provided to tell the user the context in which it is used.

A2va: I didn't know about that policy, so I will remove the execution of git. The description will be modified to inform the user that it needs to reboot to make sure that all program have registered the new key value. 
From the Microsoft docs:
> The registry key will not be reloaded during the lifetime of the process. In order for all apps on the system to recognize the value of the key, a reboot might be required because some processes may have started before the key was set.

waruqi:
ok, thanks

PR #3824, opened 2023-06-08
Make runpath recursive on Windows

SirLynix:


xmake/modules/private/action/run/make_runenvs.lua
```diff
@@ -35,24 +35,30 @@ function _make_runpath_on_windows(target)
...
+        for _, pkg in ipairs(target:orderpkgs()) do
+            for _, linkdir in ipairs(pkg:get("linkdirs")) do
+                insert(linkdir)
+            end
+        end
+        for _, dep in ipairs(target:orderdeps()) do
+            if dep:kind() == "shared" then
+                insert(dep:targetdir())
+            end
+            insert_target(dep)
```
waruqi, line 56: If two deps have the same subdeps, this will cause them to be added repeatedly. We should do some target de-duplication to avoid adding the same dep target repeatedly

SirLynix: Didn't think of that, this should fix it.

PR #3917, opened 2023-07-03
Toolchains env

A2va:
* Add some toolchains env
* Detect llvm from path

### DIFF
--- a/xmake/scripts/xrepo/envs/ndk.lua
+++ b/xmake/scripts/xrepo/envs/ndk.lua
@@ -1,0 +1,1 @@
+add_requires("ndk")
--- a/xmake/scripts/xrepo/envs/zig.lua
+++ b/xmake/scripts/xrepo/envs/zig.lua
@@ -1,0 +1,1 @@
+add_requires("zig")
--- a/xmake/toolchains/llvm/check.lua
+++ b/xmake/toolchains/llvm/check.lua
@@ -66,6 +66,21 @@ function main(toolchain)
                 sdkdir = path.directory(bindir)
             end
         end
+        -- attempt to get it from $PATH
+        if not sdkdir then
+            local pathenv = os.getenv("PATH")
+            if pathenv then
+                local buildhash_pattern = string.rep('%x', 32)
+                local match_pattern = "[\\/]packages[\\/]%w[\\/].*llvm.*[\\/][^\\/]+[\\/]" .. buildhash_pattern .. "[\\/]bin"
+                for _, p in ipairs(path.splitenv(pathenv)) do
+                    if (p:find(match_pattern) or p:find(string.ipattern("llvm[%w%-%_%+]*[\\/]bin"))) and
+                        path.filename(p) == "bin" and os.isdir(p) then
+                        sdkdir = path.directory(p)
+                        break
+                    end
+                end
+            end
+        end
     end
 
     -- find cross toolchain from external envirnoment
--- a/xmake/toolchains/llvm/check.lua
+++ b/xmake/toolchains/llvm/check.lua
@@ -66,21 +66,6 @@ function main(toolchain)
                 sdkdir = path.directory(bindir)
             end
         end
-        -- attempt to get it from $PATH
-        if not sdkdir then
-            local pathenv = os.getenv("PATH")
-            if pathenv then
-                local buildhash_pattern = string.rep('%x', 32)
-                local match_pattern = "[\\/]packages[\\/]%w[\\/].*llvm.*[\\/][^\\/]+[\\/]" .. buildhash_pattern .. "[\\/]bin"
-                for _, p in ipairs(path.splitenv(pathenv)) do
-                    if (p:find(match_pattern) or p:find(string.ipattern("llvm[%w%-%_%+]*[\\/]bin"))) and
-                        path.filename(p) == "bin" and os.isdir(p) then
-                        sdkdir = path.directory(p)
-                        break
-                    end
-                end
-            end
-        end
     end
 
     -- find cross toolchain from external envirnoment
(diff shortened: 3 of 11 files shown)


xmake/toolchains/llvm/check.lua
```diff
@@ -66,6 +66,21 @@ function main(toolchain)
...
+                local match_pattern = "[\\/]packages[\\/]%w[\\/].*llvm.*[\\/][^\\/]+[\\/]" .. buildhash_pattern .. "[\\/]bin"
+                for _, p in ipairs(path.splitenv(pathenv)) do
+                    if (p:find(match_pattern) or p:find(string.ipattern("llvm[%w%-%_%+]*[\\/]bin"))) and
+                        path.filename(p) == "bin" and os.isdir(p) then
+                        sdkdir = path.directory(p)
+                        break
+                    end
+                end
+            end
+        end
```
waruqi, line 83: Please don't do this. you should use https://github.com/xmake-io/xmake/blob/master/tests/projects/package/toolchain_llvm/xmake.lua

A2va: Why would that be more of a problem than mingw? Who already has this feature.
https://github.com/xmake-io/xmake/blob/95df1f53c56d12b940e04b326ee05e090e3fed1a/xmake/modules/detect/sdks/find_mingw.lua#L48

waruqi: This was added a long time ago before the package toolchain feature was supported, but it is no longer recommended. We now recommend the use of package toolchains.

waruqi: llvm package is toolchain kind, bin path will be added automatically. so `xrepo env -b llvm shell` has been supported. you can try it.

A2va: Yes it's supported but not detected and I think that's where there are improvements to be made.
After `xrepo env -b llvm shell` 
```
xmake f --toolchain=clang -> works , clang is on PATH
xmake f --toolchain=llvm -> didn't work
```
So it works for some toolchain because the executable is added on PATH and others no. 
There a function to enter in a package env but none to tell if this is case for a certain package.
In that case llvm could be resolved with this approximative idea:
```lua
if not sdkdir and packagenv.entered("llvm") then
    sdkdir = get install dir from package
end
```
Anyway, this isn't the right place to talk about this anymore, could I open an feature issue ?

waruqi: it works for me.

```console
ruki-2:testqt ruki$ xrepo env -b llvm shell
checking for Codesign Identity of Xcode ... Apple Development: waruqi@gmail.com (T3NA4MRVPU)
checking for SDK version of Xcode for macosx (x86_64) ... 13.0
checking for Minimal target version of Xcode for macosx (x86_64) ... 12.6
[llvm] testqt > echo $PATH
/Users/ruki/.local/bin:/Users/ruki/.xmake/packages/l/llvm/14.0.0/cd7cd141302c4ebc806462ecc47cdc1
9/bin:/Users/ruki/projects/personal/xmake/build:/Users/ruki/.local/bin:/Users/ruki/.rvm/gems/rub
y-2.6.3/bin:/Users/ruki/.rvm/gems/ruby-2.6.3@global/bin:/Users/ruki/.rvm/rubies/ruby-2.6.3/bin:/
Library//Java/JavaVirtualMachines/jdk1.8.0_291.jdk/Contents/Home//bin:/usr/local/opt/make/libexe
c/gnubin:/Users/ruki/projects/working/flutter/flutter_v1.9.1+hotfix.6-stable/flutter/bin:/Users/
ruki/projects/working/flutter/flutter_v1.9.1+hotfix.6-stable/flutter/bin/cache/dart-sdk/bin:/Use
rs/ruki/projects/working/flutter/depot_tools:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Libra
ry/Apple/usr/bin:/Users/ruki/.cargo/bin:/Users/ruki/.rvm/bin:/Users/ruki/.rvm/bin
[llvm] testqt > which llvm-ar
/Users/ruki/.xmake/packages/l/llvm/14.0.0/cd7cd141302c4ebc806462ecc47cdc19/bin/llvm-ar
[llvm] testqt > llvm-nm --version
llvm-nm, compatible with GNU nm
LLVM (http://llvm.org/):
  LLVM version 14.0.0
  Optimized build.
  Default target: x86_64-apple-darwin21.6.0
  Host CPU: skylake
[llvm] testqt >
```

xmake/scripts/xrepo/envs/ndk.lua
```diff
@@ -0,0 +1 @@
+add_requires("ndk")
```
waruqi, line 1: Some environments are not necessary built-in, users can add their own configuration according to their needs. https://xrepo.xmake.io/#/getting_started?id=manage-virtual-environments

We will only consider some of the more commonly used toolchain environments that are built in.

for example. emscripten, zig, ...

waruqi: so please remove it.

xmake/scripts/xrepo/envs/zig.lua
```diff
@@ -0,0 +1 @@
+add_requires("zig")
```
waruqi, line 1: Alternatively, if there is only one package, it is possible to bind them directly, even without adding any environment. So all these environments are not necessary.

you can run `xrepo env -b zig shell` without this script.

waruqi:
you need not any envs for single package.

```
ruki-2:tmp ruki$ xrepo env -b "zig 0.10.1" -v shell
[zig 0.10.1] tmp > zig version
0.10.1
```

PR #3919, opened 2023-07-03
change `$SHELL` with `$<SHELL>_VERSION`

### LINKED ISSUE #3915: `bash` completions error: `compctl: command not found`
### Xmake Version

xmake v2.7.9+HEAD.c879226

macOS 13.4 (22F66)

### Describe Bug

Running `bash` from an OS whose default shell is `zsh` causes completions to fail to load.
image

### Expected Behavior

I think completions should load based on the currently running shell, not the global default one.

[Xmake chooses completions based on `$SHELL`](https://github.com/xmake-io/xmake/blob/master/xmake/scripts/profile-unix.sh#L22) which returns the default shell, not the currently running one.

image

The fix could involve replacing `$SHELL` in `profile-unix.sh`. I tried `$0` since it fixes the `bash` issue, but breaks `zsh` completions.

### Project Configuration

_No response_

_No response_

vic1707:
fixes #3915, now completions are loaded based on current shell not default one.
(tested on macos and linux via a docker container)

### DIFF
--- a/xmake/scripts/profile-unix.sh
+++ b/xmake/scripts/profile-unix.sh
@@ -19,9 +19,9 @@
 #
 
 # register completions
-if   [[ "$SHELL" = */zsh ]]; then
+if   [[ -n "$ZSH_VERSION" ]]; then
   . "$XMAKE_PROGRAM_DIR/scripts/completions/register-completions.zsh"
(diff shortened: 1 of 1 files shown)


waruqi:
thanks

PR #3964, opened 2023-07-13
Fix tar extraction in mingw windows

calebkiage:
MinGW has a unix compatible tar program on windows. The tar program from the msys registry doesn't have a bundled 7zip extractor.

------

### DIFF
--- a/xmake/core/base/os.lua
+++ b/xmake/core/base/os.lua
@@ -1002,8 +1002,9 @@ end
 -- get the system null device
 function os.nuldev(input)
 
+    -- mingw and cygwin have /dev/null
     if input then
-        if os.host() == "windows" then
+        if os.host() == "windows" and not os.is_subhost("msys", "cygwin") then
             -- init the input nuldev
             if xmake._NULDEV_INPUT == nil then
                 -- create an empty file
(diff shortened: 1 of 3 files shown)


xmake/modules/utils/archive/extract.lua
```diff
@@ -33,7 +33,7 @@ import("extension", {alias = "get_archive_extension"})
 function _extract_using_tar(archivefile, outputdir, extension, opt)
 
     -- the tar of windows can only extract "*.tar"
-    if os.host() == "windows" and extension ~= ".tar" then
+    if os.host() == "windows" and not os.is_subhost("msys", "cygwin") and extension ~= ".tar" then
```
waruqi, line 36: no, msys2/cygwin's tar does not decompress many non-.tar files. So I disabled it as well. Had a lot of problems with it before

xmake/core/base/os.lua
```diff
@@ -1002,8 +1002,9 @@ end
 -- get the system null device
 function os.nuldev(input)
 
+    -- mingw and cygwin have /dev/null
     if input then
-        if os.host() == "windows" then
+        if os.host() == "windows" and not os.is_subhost("msys", "cygwin") then
```
waruqi, line 1007: I know msys2 has /dev/null. But it seems to have had some problems before. The compatibility wasn't very good, so I disabled it as well.

waruqi: Also, xmake's os.exec is always a direct call to CreateProcess, which doesn't support /dev/null for msys2. xmake doesn't use mingw's fork to create processes.

xmake/modules/utils/archive/extract.lua
calebkiage: Ok. I'll remove this

xmake/core/base/os.lua
calebkiage: I'll remove this too.

PR #3966, opened 2023-07-13
Include directories for header-only libraries in CMake only support INTERFACE access types

mohabouje:
CMake only allows the use of the INTERFACE access type to include directories in header-only libraries.

However, Xmake generates a faulty CMakeLists.txt file when a header-only library has dependencies. This results in the following error message: "target_include_directories may only set INTERFACE properties on INTERFACE targets".

The root cause of this issue is that the include directory function is using a PRIVATE access type instead of INTERFACE. To address this problem, the proposed change checks whether the target is a header-only library and adjusts the access type accordingly.


waruqi:
thanks

PR #3991, opened 2023-07-22
Set cppstd for msvc in conan 2.0

tgsong:
When using Conan 2.0 on Windows, it always try to build from source if `cppstd` is not specified. Adding `compiler.cppstd=14` will make Conan pull prebuilt binaries if it's available, see https://github.com/conan-io/conan/issues/12387.


xmake/modules/package/manager/conan/v2/install_package.lua
```diff
@@ -178,6 +178,9 @@ function _conan_generate_compiler_profile(profile, configs, opt)
         local vs = assert(config.get("vs"), "vs not found!")
         profile:print("compiler=msvc")
         profile:print("compiler.version=" .. assert(vsvers[vs], "unknown msvc version!"))
+        if vs >= 2015 then
```
waruqi, line 181: please add comment, `@see https://github.com/conan-io/conan/issues/12387`

PR #4029, opened 2023-07-31
vsproject generation: handle filegroups merging

SirLynix:
I noticed an issue when generating a vs project using filegroups that were not the same across modes.

For example:
```lua
rule("natvis")
    on_config(function (target)
        for name, pkg in pairs(target:pkgs()) do
            local includedir = path.join(pkg:installdir(), "include")
            local natvis = os.files(path.join(includedir, "**.natvis"))
            if #natvis > 0 then
                local groups = table.wrap(target:get("filegroups"))
                if not table.find(groups, name) then
                    target:add("headerfiles", natvis, { install = false })
                    target:add("filegroups", name, { rootdir = includedir })
                end
            end
        end
    end)

add_rules("mode.debug", "mode.release", "natvis")

set_runtimes(is_mode("debug") and "MDd" or "MD")

add_requires("joltphysics")

target("example")
    add_files("src/*.cpp")
    add_packages("joltphysics")
```

this works fine, but will add different filegroups in debug and in release mode (because of different vs_runtime), with the current code one filegroup will override the other and it will generate an absolute path filter chain.

this fixes it by merging filegroups confs or a target across modes.

### DIFF
--- a/xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
@@ -103,6 +103,10 @@ function _make_filter(filepath, target, vcxprojdir)
                         break
                     end
                 end
+                -- stop once a rootdir matches
+                if filter then
+                    break
+                end
             end
         end
     end
--- a/xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
@@ -100,14 +100,15 @@ function _make_filter(filepath, target, vcxprojdir)
                         if filter and filter == '.' then
                             filter = nil
                         end
-                        break
+                        goto found_filter
                     end
                 end
                 -- stop once a rootdir matches
                 if filter then
-                    break
+                    goto found_filter
                 end
             end
+            ::found_filter::
         end
     end
     if not filter and not is_plain then
--- a/xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
@@ -76,30 +76,32 @@ function _make_filter(filepath, target, vcxprojdir)
             local extraconf = filegroups_extraconf[filegroup] or {}
             local rootdir = extraconf.rootdir
             assert(rootdir, "please set root directory, e.g. add_filegroups(%s, {rootdir = 'xxx'})", filegroup)
-            if not path.is_absolute(rootdir) then
-                rootdir = path.absolute(rootdir, scriptdir)
-            end
-            local fileitem = path.relative(filepath, rootdir)
-            local files = extraconf.files or "**"
-            local mode = extraconf.mode
-            for _, filepattern in ipairs(files) do
-                filepattern = path.pattern(path.absolute(path.join(rootdir, filepattern)))
-                if filepath:match(filepattern) then
-                    if mode == "plain" then
-                        filter = path.normalize(filegroup)
-                        is_plain = true
-                    else
-                        -- file tree mode (default)
-                        if filegroup ~= "" then
-                            filter = path.normalize(path.join(filegroup, path.directory(fileitem)))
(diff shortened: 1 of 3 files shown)


waruqi:
> but will add different filegroups in debug and in release mode (because of different vs_runtime)

Why does vs_runtime affect file groups?

xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
```diff
@@ -76,30 +76,32 @@ function _make_filter(filepath, target, vcxprojdir)
...
-                filepattern = path.pattern(path.absolute(path.join(rootdir, filepattern)))
-                if filepath:match(filepattern) then
-                    if mode == "plain" then
-                        filter = path.normalize(filegroup)
-                        is_plain = true
-                    else
-                        -- file tree mode (default)
-                        if filegroup ~= "" then
-                            filter = path.normalize(path.join(filegroup, path.directory(fileitem)))
+            for _, rootdir in ipairs(table.wrap(rootdir)) do
```
waruqi, line 79: If there is more than one rootdir, and the first rootdir already makes the filter have a value. This causes the for-loop to continue until the filter corresponding to the last rootdir is obtained. I think that's a bit of a problem.

waruqi: How do we decide which rootdir filter value to use?

SirLynix:
> > but will add different filegroups in debug and in release mode (because of different vs_runtime)
> 
> Why does vs_runtime affect file groups?

because it changes package hash, here in debug, joltphysics MDd will be used and in release joltphysics MD will be used, this leads to two .natvis being added in two folders.

Without this change:
image

```xml
  <ItemGroup>
    <ClInclude Include="C:\Users\lynix\AppData\Local\.xmake\packages\j\joltphysics\v3.0.1\1805f236549648afaea1cb8a6cbf8b2f\include\Jolt\Jolt.natvis">
      <Filter>C:\Users\lynix\AppData\Local\.xmake\packages\j\joltphysics\v3.0.1\1805f236549648afaea1cb8a6cbf8b2f\include\Jolt</Filter>
    </ClInclude>
    <ClInclude Include="C:\Users\lynix\AppData\Local\.xmake\packages\j\joltphysics\v3.0.1\80c3536877aa426e81d7c74744646e74\include\Jolt\Jolt.natvis">
      <Filter>joltphysics\Jolt</Filter>
    </ClInclude>
```

With this change:
image
```xml
  <ItemGroup>
    <ClInclude Include="C:\Users\lynix\AppData\Local\.xmake\packages\j\joltphysics\v3.0.1\1805f236549648afaea1cb8a6cbf8b2f\include\Jolt\Jolt.natvis">
      <Filter>joltphysics\Jolt</Filter>
    </ClInclude>
    <ClInclude Include="C:\Users\lynix\AppData\Local\.xmake\packages\j\joltphysics\v3.0.1\80c3536877aa426e81d7c74744646e74\include\Jolt\Jolt.natvis">
      <Filter>joltphysics\Jolt</Filter>
    </ClInclude>
```

xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
SirLynix: I think we can safely stop once a rootdir matches

SirLynix: I added a break condition, it ends the loop once it found a matching rootdir

xmake/plugins/project/vstudio/impl/vs201x_vcxproj_filters.lua
```diff
@@ -76,29 +76,35 @@ function _make_filter(filepath, target, vcxprojdir)
...
-                            filter = path.normalize(path.directory(fileitem))
+                            -- file tree mode (default)
+                            if filegroup ~= "" then
+                                filter = path.normalize(path.join(filegroup, path.directory(fileitem)))
+                            else
+                                filter = path.normalize(path.directory(fileitem))
+                            end
                         end
+                        if filter and filter == '.' then
+                            filter = nil
```
waruqi, line 101: and here, we need also break rootdir loop

PR #4126, opened 2023-08-26
[Modules C++20] try to fix headerunits on msvc

Arthapz:
This fixes headerunits on VS 17.8 preview 1, and maybe VS 17.7

VS changed the way they handle headerunits

from MSVC STL server:

"there's your problem:
```
-headerUnit:quote C:\Dev\repos\header-unit-regression-VS17.8\src\header.hpp=build\.gens\user_headerunit\windows\x64\release\rules\modules\cache\be9473a8\header.hpp.ifc
```

The argument here points at the full header path, but it's written in source as #include "header.hpp".  In this release the compiler is more strict about the mapping between how it is written in source due to how easily we can optimize that path.  If you want a full path then you need to use /headerUnit without a : designating the type."


waruqi:
Thanks! 👍

PR #4174, opened 2023-09-06
Fix cpp file recompilation when using modules

Arthapz:
When using a complex module scenario, .cpp files are rebuilt every xmake call, this is due to the ordering of require_flags that may change between two xmake instance

i fixed it on clang and on msvc when using opt.expand, but i'm not comfortable enough with lua to fix it when opt.expand is not set so i may need a little help about that :D

as for GCC i'll check it later, GCC modules development is very slow, i'm waiting for improvements on their side before refine the GCC support in XMake

### DIFF
--- a/xmake/rules/c++/modules/modules_support/clang.lua
+++ b/xmake/rules/c++/modules/modules_support/clang.lua
@@ -947,6 +947,7 @@ function get_requiresflags(target, requires)
         ::continue::
     end
     if #flags > 0 then
+        table.sort(flags, function(left, right) return left < right end)
         return table.unique(flags)
     end
 end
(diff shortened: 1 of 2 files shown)


waruqi:
We don't need to sort it explicitly. Because pairs are used in lua to traverse keys in a random order, it is possible that the order of flags will be random as well.

We should find the pairs that affect requiresflags and replace them with table.orderpairs, which should solve the problem completely.

Arthapz:
it should be in fileconfig ? or in private.build.object maybe ? because on module side, after the get_requireflags there is no "pair(requireflags)", only fileconfig:add for .cpp files

waruqi:
https://github.com/xmake-io/xmake/pull/4178

I've only found one pairs that needs to be fixed, and I think msvc's get_requireflags should already be ordered, with orderpairs and orderdeps fully used in it, and there shouldn't be a problem.

waruqi:
I have merged that patch.

PR #4251, opened 2023-09-28
fix toolset in llvm toolchain for objC

Arthapz:


xmake/toolchains/llvm/xmake.lua
```diff
@@ -30,9 +30,9 @@ toolchain("llvm")
 
     -- set toolset
     set_toolset("cc",     "clang")
-    set_toolset("cxx",    "clang", "clang++")
-    set_toolset("mxx",    "clang", "clang++")
-    set_toolset("mm",    "clang", "clang++")
+    set_toolset("cxx",    "clang++", "clang")
```
waruqi, line 33: why? clang can also compile c++ code.

Arthapz: well it's more logical to use clang++ when compiling c++ :D

waruqi: I don't think it should be modified unless it's to fix some unavoidable bug.

Tools like autotools, boost/b2, etc. can detect gcc better, and I'm not sure how much passing in g++/clang++ would break anything. It would introduce a lot of uncertainty.

xmake/rules/c++/modules/modules_support/clang.lua
```diff
@@ -84,7 +84,7 @@ function _get_clang_scan_deps(target)
             local dir = path.directory(program)
             local basename = path.basename(program)
             local extension = path.extension(program)
-            program = (basename:gsub("clang", "clang-scan-deps")) .. extension
+            program = ((basename:gsub("clang", "clang-scan-deps")) .. extension):gsub("+", "")
```
waruqi, line 87: There's probably a lot of other places that could be broken.

xmake/toolchains/llvm/xmake.lua
```diff
@@ -31,8 +31,8 @@ toolchain("llvm")
     -- set toolset
     set_toolset("cc",     "clang")
     set_toolset("cxx",    "clang", "clang++")
-    set_toolset("mxx",    "clang", "clang++")
-    set_toolset("mm",    "clang", "clang++")
+    set_toolset("mxx",    "clang++", "clang")
+    set_toolset("mm",     "clang")
```
waruqi, line 35: and here

Arthapz: if i remove it, i'm getting "error: cannot get program for mxx"

xmake/toolchains/llvm/xmake.lua
```diff
@@ -32,7 +32,7 @@ toolchain("llvm")
     set_toolset("cc",     "clang")
     set_toolset("cxx",    "clang", "clang++")
     set_toolset("mxx",    "clang", "clang++")
-    set_toolset("mm",    "clang", "clang++")
+    set_toolset("mm",     "clang")
```
waruqi, line 35: and here. clang++ cannot compile objc? why?

Arthapz: you didn't add clang++ for cc so i did the same for objc

PR #4291, opened 2023-10-14
Add try option to download

A2va:

------

### DIFF
--- a/xmake/modules/net/http/download.lua
+++ b/xmake/modules/net/http/download.lua
@@ -113,6 +113,10 @@ function _curl_download(tool, url, outputfile, opt)
     table.insert(argv, outputfile)
 
     -- download it
+    if opt.try then
+        local ok = try {function () os.vrunv(tool.program, argv); return true end}
+        return ok == nil and false or true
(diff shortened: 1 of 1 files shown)


xmake/modules/net/http/download.lua
```diff
@@ -113,6 +113,10 @@ function _curl_download(tool, url, outputfile, opt)
     table.insert(argv, outputfile)
 
     -- download it
+    if opt.try then
+        local ok = try {function () os.vrunv(tool.program, argv); return true end}
+        return ok == nil and false or true
```
waruqi, line 118: why? we need not add it. you can use 

```lua
local ok = try {function () http.download(...); return true end}
```

like this

https://github.com/xmake-io/xmake/blob/23f5f45fdbc266eee5fb295641ed6b1703a2111c/xmake/modules/private/action/require/impl/actions/download.lua#L317

waruqi: It's not possible to add a try parameter to all api's, so please use try/catch {} whenever possible.

PR #4324, opened 2023-10-26
Fix asan and lto config handling

SirLynix:
This fixes two things:

1) when `project.policy("build.sanitizer.address")` is true, there is no way to disable asan on a package, `add_requires("xx", { configs { asan = false } })` won't work because of the `or` (it should check for nil)

https://github.com/xmake-io/xmake/blob/3d9cf6af8a3fae36dbb5e8fe055a54000ddcadfb/xmake/modules/private/action/require/impl/package.lua#L506

2) When asan is set to false, it won't match installed packages with asan key set to nil even though they are identical. So when asan/lto key is set to false, it will remove the asan/lto key from config

### DIFF
--- a/xmake/modules/private/action/require/impl/package.lua
+++ b/xmake/modules/private/action/require/impl/package.lua
@@ -655,8 +655,9 @@ function _inherit_parent_configs(requireinfo, package, parentinfo)
             if requireinfo_configs.pic == nil then
                 requireinfo_configs.pic = parentinfo_configs.pic
             end
-            if not requireinfo_configs.pic then
-                requireinfo_configs.pic = nil
+            -- remove pic entry if pic is true (to prevent cache mismatch), as it the default behavior
+            if requireinfo_configs.pic == true then
+                requireinfo_configs.pic = nil -- pic is enabled by default
             end
         end
         if parentinfo.plat then
--- a/xmake/modules/private/action/require/impl/package.lua
+++ b/xmake/modules/private/action/require/impl/package.lua
@@ -502,8 +502,17 @@ function _init_requireinfo(requireinfo, package, opt)
         if project.policy("package.inherit_external_configs") then
             requireinfo.configs.vs_runtime = requireinfo.configs.vs_runtime or get_config("vs_runtime")
         end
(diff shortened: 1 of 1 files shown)


xmake/modules/private/action/require/impl/package.lua
```diff
@@ -646,6 +655,9 @@ function _inherit_parent_configs(requireinfo, package, parentinfo)
             if requireinfo_configs.pic == nil then
                 requireinfo_configs.pic = parentinfo_configs.pic
             end
+            if not requireinfo_configs.pic then
+                requireinfo_configs.pic = nil
```
waruqi, line 659: it's enabled by default if pic is nil

SirLynix: Fixed

waruqi:
try this patch. https://github.com/xmake-io/xmake/pull/4326

SirLynix:
I can't test it now but I think it should work, thank you, will test soon

SirLynix:
after applying the patch:
```
in xmake-repo:
  -> ffmpeg 5.1.2 [shared:y, asan:n]
  -> openssl#1 1.1.1-t [asan:n, from:libcurl,python#3,python#4]
  -> libcurl 8.4.0 [shared:y, asan:n]
  -> expat#1 2.5.0 [asan:n, from:wayland#1]
  -> wayland#1 1.19.0 [asan:n]
please input: y (y/n/m)
```

it looks like asan = false keys were not properly evicted

waruqi:
> > try this patch. #4326
> 
> after applying the patch:
> 
> ```
> in xmake-repo:
>   -> ffmpeg 5.1.2 [shared:y, asan:n]
>   -> openssl#1 1.1.1-t [asan:n, from:libcurl,python#3,python#4]
>   -> libcurl 8.4.0 [shared:y, asan:n]
>   -> expat#1 2.5.0 [asan:n, from:wayland#1]
>   -> wayland#1 1.19.0 [asan:n]
> please input: y (y/n/m)
> ```
> 
> it looks like asan = false keys were not properly evicted (asan is missing from `_get_default_config_value_of`)

try it again

waruqi:
I have merged it.

SirLynix:
I can confirm it's working, thank you!

PR #4376, opened 2023-11-09
Add package.install_locally policy

SirLynix:
This adds a new policy allowing one to install new packages in the project folder instead of XMAKE_PKG_INSTALLDIR, but still try find packages in XMAKE_PKG_INSTALLDIR if they're not in the project folder.

This is useful in case where you want packages to be temporary installed, or installed with the project, to preserve disk space.

Note: this implementation works but can probably be improved a lot

### DIFF
--- a/xmake/core/package/package.lua
+++ b/xmake/core/package/package.lua
@@ -632,8 +632,19 @@ end
 
 -- is local package?
 -- we will use local installdir and cachedir in current project
-function _instance:is_local()
-    return self:is_source_embed() or self:is_binary_embed() or self:is_thirdparty()
+function _instance:is_local(opt)
+    if self:is_source_embed() or self:is_binary_embed() or self:is_thirdparty() then
+        return true
+    end
+
+    if not opt or not opt.ignorelocal then
+        local project = package._project()
+        if project and project.policy("package.install_locally") then
+            return true
+        end
+    end
+
+    return false
 end
 
 -- use external includes?
@@ -729,10 +740,10 @@ function _instance:sourcedir()
 end
 
 -- get the build directory
-function _instance:buildir()
+function _instance:buildir(opt)
     local buildir = self._BUILDIR
     if not buildir then
-        if self:is_local() then
+        if self:is_local(opt) then
             local name = self:name():lower():gsub("::", "_")
             local rootdir = path.join(config.buildir({absolute = true}), ".packages", name:sub(1, 1):lower(), name, self:version_str())
             buildir = path.join(rootdir, "cache", "build_" .. self:buildhash():sub(1, 8))
@@ -745,7 +756,7 @@ function _instance:buildir()
 end
 
 -- get the cached directory of this package
-function _instance:cachedir()
+function _instance:cachedir(opt)
     local cachedir = self._CACHEDIR
     if not cachedir then
         cachedir = self:get("cachedir")
@@ -756,7 +767,7 @@ function _instance:cachedir()
                 -- strip `>= <=`
                 version_str = version_str:gsub("[>=<]", "")
             end
-            if self:is_local() then
+            if self:is_local(opt) then
                 cachedir = path.join(config.buildir({absolute = true}), ".packages", name:sub(1, 1):lower(), name, version_str, "cache")
             else
                 cachedir = path.join(package.cachedir(), name:sub(1, 1):lower(), name, version_str)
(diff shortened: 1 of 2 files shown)


xmake/core/package/package.lua
```diff
@@ -1711,13 +1726,34 @@ end
...
+
+    local installdir = self:installdir({readonly = true})
+    fetchopt.installdir = installdir
+    local fetchinfo = self._find_tool(name, fetchopt)
+    if fetchinfo then
+        return fetchinfo
+    end
+
+    -- try from global install dir (if package has install_locally policy)
+    installdir = self:installdir({readonly = true, ignorelocal = true, ignorecache = true})
```
waruqi, line 1747: why? we should not find twice.

xmake/core/package/package.lua
```diff
@@ -632,8 +632,19 @@ end
...
-function _instance:is_local()
-    return self:is_source_embed() or self:is_binary_embed() or self:is_thirdparty()
+function _instance:is_local(opt)
+    if self:is_source_embed() or self:is_binary_embed() or self:is_thirdparty() then
+        return true
+    end
+
+    if not opt or not opt.ignorelocal then
+        local project = package._project()
+        if project and project.policy("package.install_locally") then
```
waruqi, line 642: and package policy

xmake/core/package/package.lua
waruqi: we need only modify is_local

SirLynix: we need to find twice because the package may be installed globally.

for example if I have libsdl already installed, and I'm using package.install_locally on a project using libsdl and other packages, I only want the other packages to be installed in the project folder. Otherwise it's not different from just setting XMAKE_PKG_INSTALLDIR to another value.

only changing package:is_local() means xmake-repo fetch won't work as it uses the installdir to fetch the package manifest.

waruqi: If libsdl has been installed to a global path, then its `is_local()` should return false instead of looking twice.

SirLynix: yeah but how do you know if the package has been installed to XMAKE_PKG_INSTALLDIR without first trying to find it?

This is a project policy, not a per-package policy.

waruqi: we can improve `package:installdir()` to support `package:installdir({local = true})`

```lua
function _instance:installdir(...)
    -- TODO opt
    local installdir = self._INSTALLDIR
    if not installdir then
        installdir = self:get("installdir")
        if not installdir then
            local name = self:name():lower():gsub("::", "_")
            if opt.local or self:is_local() then.  <----
```

then check if `package:manifest_file()`  and `package:manifest_file({local = true})` exists in self:is_local()

SirLynix: shouldn't `package:is_local()` should return true if the policy is in effect? because yeah we can force it but we would need to do the same in _instance:cachedir() and _instance:buildir() (which are testing package:is_local() for now)

waruqi: so if we set local policy globally, is_local is always true. and xmake should always install to local, even if it has been installed globally. we should also reinstall it to local. do not twice find.

SirLynix: this will make xmake reinstall all packages already installed in XMAKE_PKG_INSTALLDIR to the project. this is no different from setting XMAKE_PKG_INSTALLDIR to the project folder.

this policy aims to reduce disk space usage by allowing to test xmake projects on a separate drive than the main one (where XMAKE_PKG_INSTALLDIR points to) and be able to remove a project and all the dependencies it installed quickly.  
Making xmake reinstall already installed packages will just make it lose build time and space.

waruqi: > shouldn't should return true if the policy is in effect? 

If we just want to incrementally install packages locally, then as I just said, we just need to determine if the global manifest file exists inside is_local, and if it does, and we're still using the globally installed packages, then the current instance of the package is still a global package, and is_local will return false even though we've set up a local policy.

_instance:cachedir() and _instance:buildir()  will be changed automaticall when is_local() return true or false

waruqi: There are also some problems with calling manifest_file inside is_local. is_local will be called too early, when the buildhash is still uncertain, and this will generate some incorrect installdir.

I don't think this feature is well implemented. The current patch causes is_local to return true, but it may still be using the library files in the global directory. This inconsistent state is also unacceptable, and it causes twice find_tool/package.

I think priority should be given to maintainability, consistency of package state. If the implementation of this feature is hard to maintain and be implemented, I have to disregard it for now.

SirLynix: Another idea would be to let is_local untouched and add is_local_install (which would check policy), which would be checked by :installdir()/:cachedir()/:buildir().

The main issue is to make xmake try to find the package in the project folder and then in the XMAKE_PKG_INSTALLDIR, since xmake-repo package finder tries to use :installdir()

waruqi: I know, but I also need to consider the complexity and maintainability of the code implementation.

There is no implementation that satisfies me at the moment and I need some time. When I have time I will think about how to implement it better. But right now I'm busy with other issues and xpack.

waruqi:
Try this patch: https://github.com/xmake-io/xmake/pull/4485

SirLynix:
Thanks a lot, but I'm away from home for a week so I won't be able to test it before next week.

SirLynix:
Thanks a lot, but I'm away from home for a week so I won't be able to test it before next week.

SirLynix:
It looks like it's working great!

PR #4421, opened 2023-11-24
Fix mingw with meson

A2va:

------

### DIFF
--- a/xmake/modules/package/tools/meson.lua
+++ b/xmake/modules/package/tools/meson.lua
@@ -307,7 +307,7 @@ function _get_configs(package, configs, opt)
     end
 
     -- add cross file
-    if package:is_cross() then
+    if package:is_cross() or package:is_plat("mingw") then
         table.insert(configs, "--cross-file=" .. _get_cross_file(package, opt))
     end
 


xmake/modules/package/tools/meson.lua
```diff
@@ -307,7 +307,7 @@ function _get_configs(package, configs, opt)
     end
 
     -- add cross file
-    if package:is_cross() then
+    if package:is_cross() or package:is_plat("mingw") then
```
waruqi, line 310: is_cross should be true for mingw, we need not add it.

https://github.com/xmake-io/xmake/blob/6b400aa1f7a578694079ccb84813a02192cb28e6/xmake/core/package/package.lua#L693

plat: mingw
host: windows
subhost: msys

A2va: It false for mingw because it goes here
https://github.com/xmake-io/xmake/blob/6b400aa1f7a578694079ccb84813a02192cb28e6/xmake/core/package/package.lua#L689-L691

waruqi: ok

PR #4424, opened 2023-11-26
Fix runjobs

ShifftC:
'runjobs' would try to index a nil value if no options where given due to a typo in the initialization of 'opt'.


waruqi:
thanks

PR #4460, opened 2023-12-01
Add native file meson

A2va:
Previously, meson didn't use the toolchain selected by the user with `--toolchain` or depending on the platform, like mingw.

### DIFF
--- a/xmake/modules/package/tools/meson.lua
+++ b/xmake/modules/package/tools/meson.lua
@@ -84,12 +84,116 @@ function _translate_flags(package, flags)
     return flags
 end
 
+function _insert_cross_configs(package, file, opt)
+    -- host machine
+    file:print("[host_machine]")
+    if opt.host_machine then
+        file:print("%s", opt.host_machine)
+    elseif package:is_plat("iphoneos", "macosx") then
+        local cpu
+        local cpu_family
+        if package:is_arch("arm64") then
+            cpu = "aarch64"
+            cpu_family = "aarch64"
+        elseif package:is_arch("armv7") then
+            cpu = "arm"
+            cpu_family = "arm"
+        elseif package:is_arch("x64", "x86_64") then
+            cpu = "x86_64"
+            cpu_family = "x86_64"
+        elseif package:is_arch("x86", "i386") then
+            cpu = "i686"
+            cpu_family = "x86"
+        else
+            raise("unsupported arch(%s)", package:arch())
+        end
+        file:print("system = 'darwin'")
+        file:print("cpu_family = '%s'", cpu_family)
+        file:print("cpu = '%s'", cpu)
+        file:print("endian = 'little'")
+    elseif package:is_plat("android") then
+        local cpu
+        local cpu_family
+        if package:is_arch("arm64-v8a") then
+            cpu = "aarch64"
+            cpu_family = "aarch64"
+        elseif package:is_arch("armeabi-v7a") then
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/meson.lua
```diff
@@ -165,113 +269,15 @@ function _get_cross_file(package, opt)
             file:print("c_link_args=['%s']", table.concat(linkflags, "', '"))
             file:print("cpp_link_args=['%s']", table.concat(linkflags, "', '"))
```
waruqi, line 270: we should set ant cc/ld for native host.

if you want to support `--toolchain`, you can refer 

https://github.com/xmake-io/xmake/blob/04f15ffea7ba45d06be9d0a69bad1e1e8509ccd0/xmake/modules/package/tools/cmake.lua#L719-L729

A2va: What do you mean by "ant" ?

waruqi: ant?

waruqi: typo. sorry. we should not set cc/ld for native host.

A2va: Why not ? Even with cmake the path to the host toolchain is given.
https://github.com/xmake-io/xmake/blob/8fcc525666412aefa0457bf866922b8e97818ea8/xmake/modules/package/tools/cmake.lua#L573-L579
And if the path to the correct tooolchain is not set meson keeps using the default detected one. 
Example :
On Windows a project set with a llvm toolchain, meson will use msvc if we didn't tell to do so. I have not tested on linux, but I think it's the same problem, if the sdk was set the another path that the default system toolchain meson will use the wrong one.

waruqi: > Why not ? Even with cmake the path to the host toolchain is given.

Only when toolchain is specified explicitly, because it changes the default host compiler.

```console
xmake f --toolchain=clang
```

or `set_toolchains("clang")`

But if toolchains are not explicitly set, we only need to use the default system compiler and do not need to explicitly specify cc. Neither cmake nor autoconf specifies it.


https://github.com/xmake-io/xmake/blob/8fcc525666412aefa0457bf866922b8e97818ea8/xmake/modules/package/tools/cmake.lua#L721

 > On Windows a project set with a llvm toolchain, meson will use msvc if we didn't tell to do so. I have not tested on linux, but I think it's the same problem, if the sdk was set the another path that the default system toolchain meson will use the wrong one.

so you changed toolchain, you can set cc/ld.

```lua
    elseif package:config("toolchains") then
        -- we still need find system libraries,
        -- it just pass toolchain environments if the toolchain is compatible with host
        if _is_toolchain_compatible_with_host(package) then
        else
            _get_configs_for_cross(package, configs, opt)
        end
    else
```

But your patch will always modify cc/ld , even if we don't set toolchains explicitly.

waruqi: We should judge `package:config("toolchains")` and only modify cc/ld when it is not empty

A2va:
I have a small question about `is_compatible_with_host` function why it doesn't true for the clang toolchain on Windows ? Clang is able to create windows programs from a windows host, right?


https://github.com/xmake-io/xmake/blob/8fcc525666412aefa0457bf866922b8e97818ea8/xmake/modules/private/utils/toolchain.lua#L28

PR #4463, opened 2023-12-02
add build.c++.clang.libcxx policy to enable libc++ when compiling with clang

Arthapz:


Arthapz:
i close it, after some discussion with @SirLynix a toolchain seems to be more suitable to handle packages, i'll make an other PR with a libc++ toolchain

PR #4474, opened 2023-12-06
add --cxxstl= flag to configure clang/llvm c++ stl library

Arthapz:
introduce --cxxstl flag (named after ndk_cxxstl to retain consistency) to choose which stl library to use for clang

msstl is a no-op and is only here to differentiate package builded with libc++ / libstdc++ and microsoft stl

### DIFF
--- a/tests/projects/c++/libc++/src/main.cpp
+++ b/tests/projects/c++/libc++/src/main.cpp
@@ -1,0 +1,9 @@
+#include <iostream>
+
+using namespace std;
+
+int main(int argc, char** argv)
+{
+    cout << "hello world!" << endl; 
+    return 0;
+}
--- a/tests/projects/c++/libc++/test.lua
+++ b/tests/projects/c++/libc++/test.lua
@@ -1,0 +1,45 @@
+import("lib.detect.find_tool")
+import("core.base.semver")
+
+function _build()
+    local ci = (os.getenv("CI") or os.getenv("GITHUB_ACTIONS") or ""):lower()
+    if ci == "true" then
+        os.exec("xmake -rvD")
+    else
+        os.exec("xmake -r")
+    end
+end
+
+function main(t)
+    local clang = find_tool("clang")
+    if clang then
+        os.exec("xmake clean -a")
+        os.exec("xmake f --toolchain=clang -c")
+        _build()
+        os.exec("xmake clean -a")
+        os.exec("xmake f --toolchain=clang --cxxstl=\"libc++\" -c")
+        _build()
+        os.exec("xmake clean -a")
+        os.exec("xmake f --toolchain=clang --cxxstl=\"libstdc++\" -c")
+        _build()
+        if is_plat("windows") then
+            os.exec("xmake clean -a")
+            os.exec("xmake f --toolchain=clang --cxxstl=\"msstl\" -c")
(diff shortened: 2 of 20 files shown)


waruqi:
We shouldn't add it as an option because there are already too many options and it's only used for clang/llvm, but cxxstl is a generic name.

At least ndk_cxxstl has a prefix that identifies it as only being used in the ndk toolchain.

Also, `--cxxstl=` and get_config are only valid for global platform toolchains, and there is no way to individually control the toolchain for each target binding(target1 using llvm, target2 using gcc). It has significant limitations.

I would recommend policy to configure it, which not only converges the command line options, but also supports targets at the same time.

for example:

```lua
set_policy("build.c++.libcxx", "libc++")
set_policy("build.c++.libcxx", "libstdc++")
```

we can also only set it for clang/gcc

```lua
set_policy("build.c++.clang.libcxx", "libc++")
set_policy("build.c++.gcc.libcxx", "libstdc++")
```

or 

```console
$ xmake f --policies=build.c++.clang.libcxx:libc++
```

Also, if you plan to add new interfaces and options, you should open an issue first , and we need to discuss how to design the new interface. Instead of opening a pr directly, this can effectively avoid some unnecessary work and improve development efficiency.

SirLynix:
It's also possible to [change the standard library with GCC](https://releases.llvm.org/5.0.0/projects/libcxx/docs/UsingLibcxx.html#using-libc-with-gcc).

whether it's a config or a policy, this setting should be passed as a config key (like asan/toolchain/etc.) because C++ libraries using different libcxx are not compatible.

I also feel that if someone decides to use libc++ on Linux, package fetching from system should be disabled (as on Linux, all packages will use libstdc++)

waruqi:
> It's also possible to [change the standard library with GCC](https://releases.llvm.org/5.0.0/projects/libcxx/docs/UsingLibcxx.html#using-libc-with-gcc).
> 
> whether it's a config or a policy, this setting should be passed as a config key (like asan/toolchain/etc.) because C++ libraries using different libcxx are not compatible.
> 
> I also feel that if someone decides to use libc++ on Linux, package fetching from system should be disabled (as on Linux, all packages will use libstdc++)

However, the default libcxx used by gcc and clang may be different, e.g. gcc uses libstdc++ by default and clang uses libc++ by default.

Users may also need to choose different libcxx for different compilers.

If we use the `-cxxstl=` global option, we can't configure it flexibly.

There are also some compilers that may not support libc++.

Therefore, it is better to use `set_policy("build.c++.clang.libcxx", "libc++")` to configure libcxx for a specific compiler.

Furthermore, policies can be passed to packages, which can be configured globally for targets and packages, or individually for a target.

SirLynix:
> However, the default libcxx used by gcc and clang may be different, e.g. gcc uses libstdc++ by default and clang uses libc++ by default.

Clang uses libstdc++ by default on Linux for compatibility.

Making it a per-compiler policy seems not very accessible to me. A `set_cxxstl("libc++")` target-config seems more interesting (with maybe the same support as flags: `set_cxxstl("clang:libc++")` and `set_cxxstl("gcc:libstdc++")`?) as an API.

It's a somewhat popular configuration for devs using Clang to set libc++ as their standard library.

Issue #2948, opened 2022-10-20
Failed to install on OpenBSD


ChihHao-Su:
### Xmake Version

2.7.2

### Operating System Version and Architecture

OpenBSD 7.2 amd64

### Describe Bug

I tried installing xmake on OpenBSD, following the manual: [https://xmake.io/#/guide/installation](), I got:
```
☁  boost_stacktrace_test  bash <(wget https://xmake.io/shget.text -O -)

正重新導向輸出至「「wget-log」」。
grep: unknown option -- Pgrep:
unknown option -- Pusage: grep [-abcEFGHhIiLlnoqRsUVvwxZ] [-A num] [-B num] [-C[num]] [-e pattern]
        [-f file] [-m num] [--binary-files=value] [--context[=num]]
        [--label=name] [--line-buffered] [--null] [pattern] [file ...]

usage: grep [-abcEFGHhIiLlnoqRsUVvwxZ] [-A num] [-B num] [-C[num]] [-e pattern]
        [-f file] [-m num] [--binary-files=value] [--context[=num]]
        [--label=name] [--line-buffered] [--null] [pattern] [file ...]
grep: grep: unknown option -- Punknown option -- P

usage: grep [-abcEFGHhIiLlnoqRsUVvwxZ] [-A num] [-B num] [-C[num]] [-e pattern]
        [-f file] [-m num] [--binary-files=value] [--context[=num]]
        [--label=name] [--line-buffered] [--null] [pattern] [file ...]
usage: grep [-abcEFGHhIiLlnoqRsUVvwxZ] [-A num] [-B num] [-C[num]] [-e pattern]
        [-f file] [-m num] [--binary-files=value] [--context[=num]]
        [--label=name] [--line-buffered] [--null] [pattern] [file ...]
xmake, A cross-platform build utility based on Lua.
Copyright (C) 2015-present Ruki Wang, tboox.org, xmake.io
                         _
    __  ___ __  __  __ _| | ______
    \ \/ / |  \/  |/ _  | |/ / __ \
     >  <  | \__/ | /_| |   <  ___/
    /_/\_\_|_|  |_|\__ \|_|\_\____|
                         by ruki, xmake.io

   👉  Manual: https://xmake.io/#/getting_started
   🙏  Donate: https://xmake.io/#/sponsor

downloading https://fastly.jsdelivr.net/gh/xmake-mirror/xmake-releases@v2.7.2/xmake-v2.7.2.xz.run ..
[...]

PR #4508, opened 2023-12-15
Fix xm_engine_get_program_file for NetBSD

cmeerw:
NetBSD defines `KERN_PROC_PATHNAME`, but the `sysctl` (that should only be used for OpenBSD) doesn't work with NetBSD. Using `/proc/curproc/exe` works and is the preferred option.

### DIFF
--- a/core/src/xmake/engine.c
+++ b/core/src/xmake/engine.c
@@ -54,10 +54,8 @@
 // proc/self
 #if defined(TB_CONFIG_OS_LINUX)
 #   define XM_PROC_SELF_FILE        "/proc/self/exe"
-#elif defined(TB_CONFIG_OS_BSD) && !defined(__OpenBSD__)
-#   if defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


waruqi:
It does not work for freebsd

cmeerw:
> It does not work for freebsd

#2948 seemed to suggest that `sysctl` with `KERN_PROC_PATHNAME` is only used for OpenBSD, but looks like it's also used for FreeBSD then

But that also means that `/proc/curproc/file` doesn't work on FreeBSD? So I guess we should set `XM_PROC_SELF_FILE` only on NetBSD then?

cmeerw:
According to https://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe/1024937#1024937 `/proc/curproc/file` is indeed problematic on FreeBSD, so that makes FreeBSD and OpenBSD for the `sysctl` option and NetBSD for `/proc/curproc/exe' (with Dragonfly BSD for `/proc/curproc/file`)

waruqi:
thanks

Issue #1539, opened 2021-07-29
Add support for c sharp (dotnet core, etc)


1bardesign:
### Problem Description
C# `.cs` files (for dotnet core or other C# tooling) are reported as unrecognised:
```
[max@hostname example]$ xmake
error: unknown source file: dotnet-project/Program.cs
```

### Solution
`.cs` files should be recognised as C# and built with whatever installed tooling. dotnet core is likely to be lowest friction for xmake as it can be run directly from the commandline.

### Alternative
Currently I build the dotnet project separately and have a script to put all the build artefacts in the right spot. This sucks compared to just running xmake.

### Additional context
I'm not totally sure how `add_deps` etc should be handled. Ideally native dependencies from the same project would be put somewhere that the C# project can find it with `[DllImport ("libname")]`, and similar - this works remarkably well with c/cpp currently :)

1bardesign:
I've got a primitive integration working with the following
```lua
rule("csharp")
	set_extensions(".csproj")
	on_build_file(function (target, sourcefile)
		os.runv("dotnet", {"build", sourcefile, "-o", target:targetdir()})
	end)
	on_clean(function (target, sourcefile)
		os.runv("dotnet", {"clean", sourcefile, "-o", target:targetdir()})
	end)
	on_run(function (target, sourcefile)
		os.runv("dotnet", {"run", sourcefile, "-o", target:targetdir()})
	end)
	on_link(function () end)
	rule_end()

target("test-cs")
	set_kind("binary")
	add_rules("csharp")
	add_files("dotnet/**.csproj")
```

This fails to run the program:
```
[max@hostname example]$ xmake run test-cs
error: please run `$xmake [target]` to build the following targets first:
  -> test-cs

```

but at least builds it in the right place

waruqi:
> This fails to run the program:

you need run `xmake build` and ensure `target:targetfile()` is generated.

1bardesign:
With the run rule above i get a cryptic warning about argv:
```
error: invalid argv[?
```

Without the special run rule it works, if the exact name of the final assembly matches the rule name, which i guess makes sense.

I'm not sure how best to handle running the dotnet assembly - it works as above but seems very fragile (the final assembly depends on the name of the `.csproj`, so they need to match). This is sort of expected, as `xmake` is just wrapping `dotnet`'s build logic, but is not particularly nice. You can probably comment on what would be the best approach.

[...]

PR #4510, opened 2023-12-16
Add naive csharp project support for vsxmake

Kethers:
# Note

This is a very naive and basic support for C# only in windows, and much much far from completely supports C# ecosystem like .Net Core or more advanced topics.

## Motivation

I'm learning [TheCherno's Game Engine Series](https://www.youtube.com/playlist?list=PLlrATfBNZ98dC-V-N3m0Go4deliWHPFwT) courses. In the video Cherno uses premake as build tool. I decide to use xmake since it's getting popular in Chinese game & graphics development community. Everything works fine until the course step into C# Scripting and xmake [hasn't supported csharp yet #1539](https://github.com/xmake-io/xmake/issues/1539). So I have to modified the xmake to make it work.

## How it works in short

When building vsxmake solution, xmake generates .sln and .vcxproj files that uses XML (via customized template). When builds the target project, it uses MSBuild to analyze the .vcxproj. In the .vcxproj template it imports the `Xmake.targets` and the latter one use `<Target>` semantic to override the real build actions like config, build, clean, etc.

To support CSharp project, here’s the work that needed:

1. Frontend: Add languages/rules for csharp (not completely done)
2. Modified the .sln files to recognize the .csproj
3. Add .csproj XML templates to generate the .csproj file
4. Backend: Add cross-platform compiling pipeline. (of course not done since it’s too hard and too much work to make it done)

For step 1, currently  `set_languages(csharp)` and `add_files("xx.cs")` works fine, and `set_kind` only supports binary and shared.

For step 2 and 3, currently MSBuild.exe can parse the .sln and .csproj correctly. IntelliSense and Property setting window works fine (tested on VS2022 and Rider 2023.3.1).

For step 4, currently I use the `Microsoft.CSharp.targets` instead of `Xmake.CSharp.targets` to build targets, which means it doesn’t override the default build actions. Thus the build actions in the `rules/csharp/xmake.lua` will not be executed at all.

## Optimizations & bug fixes

1. The project reference in `#target#.vcxproj` is commented since initial commit. I don’t see why since uncomment it also works fine. For .vcxproj it’s just a slight improvement when using IDE. However for .csproj, it’s a necessary semantic otherwise IntelliSense disfunctions when referencing another C# project via `add_deps(...)`.
2. `set_objectdir("...")` doesn’t work correctly on .vcxproj because of hard-coded path, now it works fine.
3. In project property window, language standard display of c++20 is not correct. (expected: c++20, actually is: c++latest)

Maybe these changes can be extracted as another new pull request.

## Test
```shell
xmake project -k vsxmake
```
[CSharpTest.zip](https://github.com/xmake-io/xmake/files/13691682/CSharpTest.zip)

## Future work

Due to the limitation of time and my ability, there are many things that to be working on, and needs community collaborations:

1. When writing `#target#.csproj`, `Xmake.Common.props` and `Xmake.Csharp.targets`(this file is not deployed), I take C++ project template files for reference. Thus in these files many C++ settings are left. I didn’t delete them for the reason of compatibility (lol, too noob to fully understood the xmake c++ build pipeline and relevant api mechanics). These legacy code should be cleaned up.

2. For the similar reason, files in `languages/csharp/` and `rules/csharp` also contain a lot legacy code that copy from c++ corresponding files that should be cleaned up.
3. Backend compilation override, and cross-platform, more C# ecosystem support and more apis for C# builds.

## External reference

1. [Premake](https://github.com/premake/premake-core) implementation of .csproj file
2. MSDN of MSBulid and .csproj XML


xmake/languages/csharp/load.lua
```diff
@@ -0,0 +1,129 @@
...
+-- get apis
+function _get_apis()
+    local apis = {}
+    apis.values = {
+        -- target.add_xxx
+        "target.add_links"
+    ,   "target.add_syslinks"
+    ,   "target.add_cflags"
+    ,   "target.add_cxflags"
+    ,   "target.add_cxxflags"
```
waruqi, line 30: there is not cflags/cxxflags for csharp, please remove it. and you shoud add `add_csflags`

xmake/languages/csharp/xmake.lua
```diff
@@ -0,0 +1,155 @@
...
+--
+-- Copyright (C) 2015-present, TBOOX Open Source Group.
+--
+-- @author      Kethers
+-- @file        xmake.lua
+--
+
+language("csharp")
+    add_rules("csharp")
+    set_sourcekinds {cs = {".cs"}}
```
waruqi, line 23: set_sourceflags {cs = {"csflags"}}

xmake/languages/csharp/load.lua
```diff
@@ -0,0 +1,129 @@
...
+    ,   "package.add_linkdirs"
+    ,   "package.add_includedirs" --@note we need not uses paths for package, see https://github.com/xmake-io/xmake/issues/717
+    ,   "package.add_sysincludedirs"
+    ,   "package.add_frameworkdirs"
+        -- toolchain.add_xxx
+    ,   "toolchain.add_links"
+    ,   "toolchain.add_syslinks"
+    ,   "toolchain.add_cflags"
+    ,   "toolchain.add_cxflags"
+    ,   "toolchain.add_cxxflags"
```
waruqi, line 80: and here

xmake/languages/csharp/load.lua
```diff
@@ -0,0 +1,129 @@
...
+    ,   "package.add_arflags"
+    ,   "package.add_shflags"
+    ,   "package.add_defines"
+    ,   "package.add_undefines"
+    ,   "package.add_frameworks"
+    ,   "package.add_rpathdirs"
+    ,   "package.add_linkdirs"
+    ,   "package.add_includedirs" --@note we need not uses paths for package, see https://github.com/xmake-io/xmake/issues/717
+    ,   "package.add_sysincludedirs"
+    ,   "package.add_frameworkdirs"
```
waruqi, line 74: please remove package apis first.

xmake/languages/csharp/load.lua
```diff
@@ -0,0 +1,129 @@
...
+        -- option.add_xxx
+    ,   "option.add_linkdirs"
+    ,   "option.add_includedirs"
+    ,   "option.add_sysincludedirs"
+    ,   "option.add_frameworkdirs"
+    }
+    apis.dictionary = {
+        -- option.add_xxx
+        "option.add_csnippets"
+    ,   "option.add_cxxsnippets"
```
waruqi, line 120: please remove them

xmake/languages/csharp/xmake.lua
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    set_menu {
+                config =
+                {
+                    {category = "Cross Complation Configuration/Compiler Configuration"                             }
+                ,   {nil, "cc",            "kv", nil,          "The C Compiler"                                     }
+                ,   {nil, "cxx",           "kv", nil,          "The C++ Compiler"                                   }
+                ,   {nil, "cpp",           "kv", nil,          "The C Preprocessor"                                 }
```
waruqi, line 123: csharp need not them

xmake/languages/csharp/xmake.lua
```diff
@@ -0,0 +1,155 @@
...
+                ,   {category = "Cross Complation Configuration/Linker Configuration"                               }
+                ,   {nil, "ld",            "kv", nil,          "The Linker"                                         }
+                ,   {nil, "ar",            "kv", nil,          "The Static Library Linker"                          }
+                ,   {nil, "sh",            "kv", nil,          "The Shared Library Linker"                          }
+                ,   {nil, "ranlib",        "kv", nil,          "The Static Library Index Generator"                 }
+
+                ,   {category = "Cross Complation Configuration/Compiler Flags Configuration"                       }
+                ,   {nil, "cflags",        "kv", nil,          "The C Compiler Flags"                               }
+                ,   {nil, "cxflags",       "kv", nil,          "The C/C++ compiler Flags"                           }
+                ,   {nil, "cxxflags",      "kv", nil,          "The C++ Compiler Flags"                             }
```
waruqi, line 134: and here

xmake/plugins/project/vsxmake/getinfo.lua
```diff
@@ -143,6 +145,11 @@ function _make_targetinfo(mode, arch, target)
     -- save defines
     targetinfo.defines       = _make_arrs(_get_values_from_target(target, "defines"))
 
+    -- save symbols, optimize, strip info
+    targetinfo.symbols = tostring(target:get("symbols"))
+    targetinfo.optimize = tostring(target:get("optimize"))
+    targetinfo.strip = tostring(target:get("strip"))
```
waruqi, line 151: please align `=`

waruqi: please remove `tostring`, get() maybe return nil or table.

xmake/plugins/project/vstudio/impl/vsinfo.lua
```diff
@@ -92,6 +98,7 @@ local vsinfo =
     ,   solution_version    = "12"
     ,   toolset_version     = "v143"
     ,   sdk_version         = "10.0.19041.0"
+    ,   dotnetframework_ver = "4.7.2"
```
waruqi, line 101: please rename tp dotnet_version

xmake/plugins/project/vsxmake/langtype.lua
```diff
@@ -0,0 +1,86 @@
...
+
+    return false
+end
+
+function iscpp(language)
+    if language and _g.cxxmaps[language] then
+        return true
+    end
+
+    return false
```
waruqi, line 85: please remove this file

xmake/plugins/project/vsxmake/getinfo.lua
```diff
@@ -486,20 +493,39 @@ function main(outputdir, vsinfo)
...
                 _target.vcxprojdir = path.join(vsinfo.solution_dir, targetname)
+                _target.csprojdir = path.join(vsinfo.solution_dir, targetname)
+                local xxprojdir = _target.vcxprojdir
+                
+                local targetinfo = _make_targetinfo(mode, arch, target)
+                _target.languages = targetinfo.languages;
+                _target.symbols = targetinfo.symbols;
+                _target.optimize = targetinfo.optimize;
+                _target.strip = targetinfo.strip;
+                if langtype.isc(_target.languages) or langtype.iscpp(_target.languages) then
```
waruqi, line 504: we should not use languages to judge c/c++/cs, it maybe nil or table.

waruqi: we should judge it from target:sourcebatches() and sourcekind

xmake/rules/csharp/xmake.lua
```diff
@@ -0,0 +1,58 @@
...
+    -- $(XmakeProgramDir)\scripts\vsxmake\vsproj\Xmake.CSharp.targets
+    on_build(function(target)
+        
+    end)
+    
+    on_clean(function(target)
+
+    end)
+    
+    on_link(function () end)
```
waruqi, line 35: function (target)

xmake/rules/csharp/xmake.lua
```diff
@@ -0,0 +1,58 @@
...
+--
+
+rule("csharp.build")
+    set_sourcekinds("cs")
+    -- TODO: now csharp builds is using MSBuild.exe, below code does nothing
+    -- if you want this file to be used, go to #target#.csproj file and replace
+    -- $(MSBuildToolsPath)\Microsoft.CSharp.targets
+    -- $(XmakeProgramDir)\scripts\vsxmake\vsproj\Xmake.CSharp.targets
+    on_build(function(target)
+
```
waruqi, line 28: vsxmake project should call xmake to build csharp.

waruqi:
I tried it and it worked, but it didn't call the xmake command to compile the cs project, but directly used vs to build it.

If so, you should implement it in vs project generator (xmake project -k vs) instead of vsxmake.

The vsxmake generator is only used to wrap xmake cli. The xmake command is directly called in the vs project to compile the csharp program.

Kethers:
> I tried it and it worked, but it didn't call the xmake command to compile the cs project, but directly call vs to build it.
> 
> If so, you should implement it in vs project generator (xmake project -k vs) instead of vsxmake.
> 
> The vsxmake generator is only used to wrap xmake cli. The xmake command is directly called in the vs project to compile the csharp program.

Thanks for the reply. I am aware of the fact that it not completely fits in the vsxmake build paradigm. I am trying to figure out a 
 good way to orchestrate them and there's no doubt that it needs deeper understanding of how xmake build cli functions in the backend and how C# build toolchain works and furthermore to abstract and integrate it into xmake backend. That's a lot of work and it takes time.
Present build backend is definitely more fittable to the vs proj generator. However I will work on vsxmake first since it's more recommendable to use than the old vs in the wiki (although it's easier to implement than vsxmake).
I might be too reckless to pull this unaccomplished request just for helping somebody like me out. At the moment just left it for some time until everything is done:
- [ ] C# build backend integration into xmake
- [ ] Use `Xmake.CSharp.targets` instead of `Microsoft.CSharp.targets` in .csproj template.
- [ ] csflags support and LangVersion support

SirLynix:
I think it's worth a shot to implement it in for the vs project generator first since it's not going away (vsxmake generator doesn't fully replace it) and it would be great to have C# support in both native vs project and vsxmake/xmake backend.

waruqi:
> I think it's worth a shot to implement it in for the vs project generator first since it's not going away (vsxmake generator doesn't fully replace it) and it would be great to have C# support in both native vs project and vsxmake/xmake backend.

right, we can make vs generator to support csharp first.

kgs233:
I think if you want to use xmake as a build system for C#, you should start with csc (compilation tool), if you simply generate MSbuild project from xmake.lua, then this process may not make much sense, because MSbuild is simple and clear enough (at least in my opinion).

Kethers:
> I think if you want to use xmake as a build system for C#, you should start with csc (compilation tool), if you simply generate MSbuild project from xmake.lua, then this process may not make much sense, because MSbuild is simple and clear enough (at least in my opinion).

Thanks for the advice. I know csc.exe. If you check the output log (set detail mode to at least normal) of c# project in vs you will see that it's calling csc.exe with source files plus some compilation options. My initial purpose was to make c# scripting orchestra with xmake and vs2022/rider so I just stopped when msbuild parse .csproj correctly. I haven't go that deep into xmake's compiler backend to see how it works. Furthermore c# ecosystem also have .Net core which might use ''dotnet build'' command. That should be taken into consideration becsuse it's more cross-platform than the old framework one (expired and no more update/newer version), especially when it comes to integrating to a cross-platform build system. Unfortunately at present I am busy doing other stuff and don't have not enough time for this pr. I will accomplish it in the future but I can't make time promise.

ArthurVasseur:
Hello, are there any updates on this PR?

Kethers:
> Hello, are there any updates on this PR?

No updates yet. Too busy to spend time on this. It's been almost a year and xmake has updated serveral versions. This PR needs more work and considerations to deploy. I will try to make some updates in next year.

waruqi:
try this patch https://github.com/xmake-io/xmake/pull/7398

PR #4513, opened 2023-12-16
Skip root user check on Haiku OS

cmeerw:
Haiku OS doesn't have any non-root users, so it doesn't make sense to warn about running as the root user.

### DIFF
--- a/xmake/core/main.lua
+++ b/xmake/core/main.lua
@@ -266,7 +266,7 @@ function main.entry()
     end
 
     -- check run command as root
-    if not option.get("root") and os.getenv("XMAKE_ROOT") ~= 'y' then
+    if not option.get("root") and os.getenv("XMAKE_ROOT") ~= 'y' and os.host() ~= 'haiku' then
         if os.isroot() then
             errors = [[Running xmake as root is extremely dangerous and no longer supported.
 As xmake does not drop privileges on installation you would be giving all
--- a/xmake/core/main.lua
+++ b/xmake/core/main.lua
@@ -238,6 +238,11 @@ function main._exit(ok, errors)
     return retval
 end
 
+-- limit root? @see https://github.com/xmake-io/xmake/pull/4513
+function main._limit_root()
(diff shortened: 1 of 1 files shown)


xmake/core/main.lua
```diff
@@ -266,7 +266,7 @@ function main.entry()
     end
 
     -- check run command as root
-    if not option.get("root") and os.getenv("XMAKE_ROOT") ~= 'y' then
+    if not option.get("root") and os.getenv("XMAKE_ROOT") ~= 'y' and os.host() ~= 'haiku' then
```
waruqi, line 269: why?

waruqi:
> Haiku OS doesn't have any non-root users, so it doesn't make sense to warn about running as the root user.

so you should improve os.isroot() to support for haiku.

SirLynix:
What if some code errors if os.isroot() returns false on haiku later? for example if we check root privileges before installing to /usr

cmeerw:
> > Haiku OS doesn't have any non-root users, so it doesn't make sense to warn about running as the root user.
> 
> so you should improve os.isroot() to support for haiku.

I am not convinced that changing isroot is preferable.

Currently, code checks `os.isroot()` to ask the question "Do I have the privileges to make system-wide changes?" (in several `install_package.lua`) On Haiku OS the answer to that question should be "yes", so I'd think `os.isroot()` should return `true` on Haiku OS. On the other hand, there is no non-root user on Haiku OS, so warning about running as root doesn't seem helpful.

I am just worried, that trying to return `false` for `os.isroot()` on Haiku OS will actually cause more maintenance issues in the future as basically all those `os.isroot()` calls will need special handling on Haiku OS.

waruqi:
> I am just worried, that trying to return false for os.isroot() on Haiku OS will actually cause more maintenance issues in the future as basically all those os.isroot() calls will need special handling on Haiku OS.

why? os.isroot() just return false on Haiku OS. and privilege.get() return true. 

Only two places need to be changed only. No other special handling.

waruqi:
well.

PR #4517, opened 2023-12-18
Fix vstudio arm64 dectect under Developer Command Prompt

dragonflylee:
Fix find_vstudio detect arm64 toolchain under Developer Command Prompt


waruqi:
thanks

PR #4523, opened 2023-12-20
Fix package detection and install on msys2 for ucrt64 and clang64 platforms

ImperatorS79:
When xmake was run inside the ucrt64 shell, incorrect packages would be installed when requesting xmake to install packages from pacman.

### DIFF
--- a/xmake/modules/package/manager/pacman/find_package.lua
+++ b/xmake/modules/package/manager/pacman/find_package.lua
@@ -30,6 +30,7 @@ function _find_package_from_list(list, name, pacman, opt)
     -- mingw + pacman = cygpath available
     local cygpath = nil
     local pathtomsys = nil
+    local msystem = nil
     if is_subhost("msys") and opt.plat == "mingw" then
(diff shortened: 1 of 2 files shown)


xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -30,13 +30,15 @@ function _find_package_from_list(list, name, pacman, opt)
...
     local pathtomsys = nil
+    local msystem = nil
     if is_subhost("msys") and opt.plat == "mingw" then
         cygpath = find_tool("cygpath")
         if not cygpath then
             return
         end
         pathtomsys = os.iorunv(cygpath.program, {"--windows", "/"})
         pathtomsys = pathtomsys:trim()
+        msystem = string.lower(os.getenv("MSYSTEM"))
```
waruqi, line 41:
```lua
msystem = os.getenv("MSYSTEM")
if msystem then
    msystem = msystem:lower()
end
```

PR #4573, opened 2024-01-05
Add loong64 support

linqigang888:
This patch adds support for loong64 architecture. I am forwarding it from a bug report submitted to [Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1059959). It has been applied in a patch for Debian also.


waruqi:
I have added this patch in dev branch. see https://github.com/xmake-io/xmake/pull/4569

PR #4639, opened 2024-01-22
fix qt.qrc rule

RC1844:
Fix the qt.qrc rule and add multiple qrc files with the same name, which will overwrite each other.


waruqi:
thanks

PR #4693, opened 2024-02-04
support --sdk for --runtimes

Arthapz:
when setting --runtimes if multiple llvm are installed on the system and --sdk is provided, llvm still use the system one,
so we need to manually pass the flags to ensure the right libc++ lib / include are used

### DIFF
--- a/xmake/modules/core/tools/clang.lua
+++ b/xmake/modules/core/tools/clang.lua
@@ -243,14 +243,18 @@ function nf_runtime(self, runtime, opt)
             maps["c++_shared"]    = "-stdlib=libc++"
             maps["stdc++_static"] = "-stdlib=libstdc++"
             maps["stdc++_shared"] = "-stdlib=libstdc++"
-            -- clang on windows fail to add libc++ includepath when using -stdlib=libc++ so we manually add it
-            -- @see https://github.com/llvm/llvm-project/issues/79647
-            if self:is_plat("windows") then
-                local llvm_rootdir = _get_llvm_rootdir(self)
-                if llvm_rootdir then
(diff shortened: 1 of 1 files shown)


xmake/modules/core/tools/clang.lua
```diff
@@ -243,14 +243,18 @@ function nf_runtime(self, runtime, opt)
...
-            -- clang on windows fail to add libc++ includepath when using -stdlib=libc++ so we manually add it
-            -- @see https://github.com/llvm/llvm-project/issues/79647
-            if self:is_plat("windows") then
-                local llvm_rootdir = _get_llvm_rootdir(self)
-                if llvm_rootdir then
-                    maps["c++_static"] = table.join(maps["c++_static"], "-cxx-isystem" .. path.join(llvm_rootdir, "include", "c++", "v1"))
-                    maps["c++_shared"] = table.join(maps["c++_shared"], "-cxx-isystem" .. path.join(llvm_rootdir, "include", "c++", "v1"))
-                end
+            local llvm_rootdir
+            if get_config("sdk") then
```
waruqi, line 247: but it does not support `toolchain/set_sdkdir`

Arthapz: i think it's okay now, can u confirm ? (i'm not sure how to get toolchain sdkdir)

xmake/modules/core/tools/clang.lua
```diff
@@ -238,19 +239,20 @@ function nf_runtime(self, runtime, opt)
     end
     if not self:is_plat("android") then -- we will set runtimes in android ndk toolchain
         maps = maps or {}
+        local llvm_rootdir = get_config("sdk") or self:get("sdk")
```
waruqi, line 242:
```lua
local llvm_rootdir = self:toolchain():sdkdir() or get_config("sdk")
```

PR #4695, opened 2024-02-04
Fix libc++.modules.json finding

Arthapz:
when libc++.modules.json is in subdir of lib folder, it fail to find it (for example on debian platform)

### DIFF
--- a/xmake/rules/c++/modules/modules_support/clang/compiler_support.lua
+++ b/xmake/rules/c++/modules/modules_support/clang/compiler_support.lua
@@ -209,7 +209,7 @@ function get_stdmodules(target)
                 -- @see https://github.com/llvm/llvm-project/pull/76451 (has been revert, so we need to wait)
                 local clang_path = path.directory(get_clang_path(target))
                 local clang_lib_path = path.join(clang_path, "..", "lib")
-                local modules_json_path = find_file("libc++.modules.json", clang_lib_path)
+                local modules_json_path = find_file("**.modules.json", clang_lib_path)
                 if modules_json_path then
                     local modules_json = json.decode(io.readfile(modules_json_path))
                     local std_module_directory = path.directory(modules_json.modules[1]["source-path"])


xmake/rules/c++/modules/modules_support/clang/compiler_support.lua
```diff
@@ -209,7 +209,7 @@ function get_stdmodules(target)
                 -- @see https://github.com/llvm/llvm-project/pull/76451 (has been revert, so we need to wait)
                 local clang_path = path.directory(get_clang_path(target))
                 local clang_lib_path = path.join(clang_path, "..", "lib")
-                local modules_json_path = find_file("libc++.modules.json", clang_lib_path)
+                local modules_json_path = find_file("**.modules.json", clang_lib_path)
```
waruqi, line 212: why not use `-print-library-module-manifest-path`?

```
def print_std_module_manifest_path : Flag<["-", "--"], "print-library-module-manifest-path">,
  HelpText<"Print the path for the C++ Standard library module manifest">,
  Visibility<[ClangOption, CLOption]>;
```

https://github.com/llvm/llvm-project/pull/76451/files

waruqi: It looks like it might also be modules-asan.json

```
    // For example
    //  StringRef modules = [&] {
    //    const SanitizerArgs &Sanitize = TC.getSanitizerArgs(C.getArgs());
    //    if (Sanitize.needsAsanRt())
    //      return "modules-asan.json";
    //    return "modules.json";
    //  }();
```

Arthapz: > why not use `-print-library-module-manifest-path`?
> 
> ```
> def print_std_module_manifest_path : Flag<["-", "--"], "print-library-module-manifest-path">,
>   HelpText<"Print the path for the C++ Standard library module manifest">,
>   Visibility<[ClangOption, CLOption]>;
> ```
> 
> https://github.com/llvm/llvm-project/pull/76451/files

it has been reverted, i plan to support it when it's re-merged
https://github.com/llvm/llvm-project/commit/82f424f766be00b037a706a835d0a0663a2680f1

waruqi:
ok

PR #4696, opened 2024-02-05
Add ldflags to export attributes loop

samayala22:
This fixes https://github.com/xmake-io/xmake-repo/issues/3224.

Verified with the example project provided in the issue on Ubuntu with GCC compiler. 

Documentation about how to contribute to xmake is poor so I dont know how to run tests locally. I just compiled xmake, modified `inherit_links.lua` and tested on my project.


waruqi:
wait for this patch https://github.com/xmake-io/xmake/pull/4698

PR #4707, opened 2024-02-05
Support moduleonly libraries

Arthapz:
Implement https://github.com/xmake-io/xmake/pull/4673#issuecomment-1926947794
need https://github.com/xmake-io/xmake/pull/4692 so i'll rebase on modules it when merge

```lua
add_rules("mode.release", "mode.debug")
set_languages("c++20")

target("bar2")
    set_kind("moduleonly")
    add_files("*.mpp") -- implicitly public
```

PR #4728, opened 2024-02-10
implement module reusage based on flag comparison

Arthapz:
this PR is based on https://github.com/xmake-io/xmake/pull/4707, i'll rebase on modules branch after merging https://github.com/xmake-io/xmake/pull/4707


CHANGELOG.md
```diff
@@ -1742,6 +1742,7 @@
 ### Bugs 修复
 
 * [#4596](https://github.com/xmake-io/xmake/issues/4596): 修复远程构建缓存
+* [#4689](https://github.com/xmake-io/xmake/issues/4689): 修复目标依赖继承
 
 ## v2.8.6
```
waruqi, line 1748: we should not contain this patch

waruqi:
I have merged #4707

Arthapz:
image
do you have an idea to what can cause relinking of the target ?

waruqi:
you can use `xmake f --policies=diagnosis.check_build_deps` to enable builddeps info to debug it. or debug here. 

https://github.com/xmake-io/xmake/blob/c601bfa8240467a0a7d5ba463dd321b2be1d1bdd/xmake/modules/core/project/depend.lua#L109

https://github.com/xmake-io/xmake/blob/c601bfa8240467a0a7d5ba463dd321b2be1d1bdd/xmake/actions/build/kinds/binary.lua#L77

Arthapz:
So to explain this PR, the objective is to reuse BMI of deps targets if flags are compatible
there is no official list of flags that affect BMI we'll empirically maintain the list based on experience and futur bug reports
i've already added all the flags that i can think of that will not affect BMI generation

xmake/rules/c++/modules/xmake.lua
```diff
@@ -119,6 +119,7 @@ rule("c++.build.modules.builder")
 
                 -- cull external modules objectfile
                 compiler_support.cull_objectfiles(target, modules, sourcebatch)
+                table.sort(sourcebatch.objectfiles)
```
waruqi, line 122: This seems to solve the problem, but I feel like I should find what is actually causing the objectfiles to get out of order, maybe somewhere in the pairs call. Then use orderpairs instead of it.

xmake/rules/c++/modules/modules_support/builder.lua
```diff
@@ -81,23 +80,94 @@ function _build_headerunits(target, headerunits, opt)
...
+  flags2 = compiler_support.strip_flags(target, flags2)
+
+  if #flags1 ~= #flags2 then
+      return false
+  end
+
+  table.sort(flags1)
+  table.sort(flags2)
+
+  for i = 1,#flags1 do
```
waruqi, line 102: add spaces ` , `

xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
```diff
@@ -65,6 +65,41 @@ function load(target)
...
+        "-Q",
+    }
+    if not target:policy("build.c++.modules.tryreuse.discriminate_on_defines") then
+        table.join2(strippable_flags, {"-D", "-U"})
+    end
+    local output = {}
+    local last_flag_I = false
+    for _, flag in ipairs(flags) do
+        local strip = false
+
```
waruqi, line 88: remove empty line

PR #4804, opened 2024-03-06
Fix toolchain ld path

Arthapz:
when using --sdk alongside of --runtimes="c++" to build with a non-system llvm, we need to fix rpath to allow executables ton find libc++

### DIFF
--- a/xmake/modules/core/tools/clang.lua
+++ b/xmake/modules/core/tools/clang.lua
@@ -195,7 +195,7 @@ end
 function _get_llvm_rootdir(self)
     local llvm_rootdir = _g._LLVM_ROOTDIR
     if llvm_rootdir == nil then
-        local outdata = try { function() return os.iorun(self:program() .. " -print-resource-dir") end }
+        local outdata = try { function() return os.iorunv(self:program(), {"-print-resource-dir"}, {envs = self:runenvs()}) end }
         if outdata then
             llvm_rootdir = path.normalize(path.join(outdata:trim(), "..", "..", ".."))
             if not os.isdir(llvm_rootdir) then
@@ -211,7 +211,7 @@ end
 function _get_llvm_target_triple(self)
     local llvm_targettriple = _g._LLVM_TARGETTRIPLE
     if llvm_targettriple == nil then
-        local outdata = try { function() return os.iorun(self:program() .. " -print-target-triple") end }
+        local outdata = try { function() return os.iorunv(self:program(), {"-print-target-triple"}, {envs = self:runenvs()}) end }
         if outdata then
             llvm_targettriple = outdata:trim()
         end
--- a/xmake/modules/core/tools/clang.lua
+++ b/xmake/modules/core/tools/clang.lua
@@ -295,9 +295,9 @@ function nf_runtime(self, runtime, opt)
                     if triple_libdir then
                         maps["c++_shared"] = table.join(maps["c++_shared"], nf_rpathdir(self, triple_libdir))
                     end
-                    if target:kind() == "shared" and self:is_plat("macosx", "iphoneos", "watchos") then
+                    if target:is_shared() and self:is_plat("macosx", "iphoneos", "watchos") then
                         maps["c++_shared"] = table.join(maps["c++_shared"], "-install_name")
-                        maps["c++_shared"] = table.join(maps["c++_shared"], "@rpath/" .. path.filename(target:filename()))
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "@rpath/" .. target:filename())
                     end
                 end
                 if runtime:endswith("_static") and _has_static_libstdcxx(self) then
--- a/xmake/modules/core/tools/clang.lua
+++ b/xmake/modules/core/tools/clang.lua
@@ -220,24 +220,6 @@ function _get_llvm_target_triple(self)
     return llvm_targettriple or nil
 end
 
--- make the rpathdir flag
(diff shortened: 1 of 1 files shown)


xmake/modules/core/tools/clang.lua
```diff
@@ -267,8 +280,23 @@ function nf_runtime(self, runtime, opt)
...
+                    maps["c++_shared"] = table.join(maps["c++_shared"], "-L" .. libdir)
+                    -- sometimes llvm runtimes are located in a target-triple subfolder
+                    local target_triple = _get_llvm_target_triple(self)
+                    local triple_libdir = (target_triple and os.isdir(path.join(libdir, target_triple))) and path.join(libdir, target_triple)
+                    if triple_libdir then
+                        maps["c++_static"] = table.join(maps["c++_static"], "-L" .. triple_libdir)
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "-L" .. triple_libdir)
+                    end
+                    -- add rpath to avoid the user need to set LD_LIBRARY_PATH by hand
+                    if not self:is_plat("windows") and not self:is_plat("mingw") then
```
SirLynix, line 294: `if not self:is_plat("windows", "mingw") then`

xmake/modules/core/tools/clang.lua
```diff
@@ -267,8 +280,23 @@ function nf_runtime(self, runtime, opt)
...
+                    local triple_libdir = (target_triple and os.isdir(path.join(libdir, target_triple))) and path.join(libdir, target_triple)
+                    if triple_libdir then
+                        maps["c++_static"] = table.join(maps["c++_static"], "-L" .. triple_libdir)
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "-L" .. triple_libdir)
+                    end
+                    -- add rpath to avoid the user need to set LD_LIBRARY_PATH by hand
+                    if not self:is_plat("windows", "mingw") then
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "-Wl,-rpath=" .. libdir)
+                        if triple_libdir then
+                            maps["c++_shared"] = table.join(maps["c++_shared"], "-Wl,-rpath=" .. triple_libdir)
```
waruqi, line 297: Has it been tested on macos? it should use -Xlinker -rpath -Xlinker path and `-install_name`.

https://github.com/xmake-io/xmake/blob/2c67c028dffe3e456e8b9ac16296ef042629b88b/xmake/modules/core/tools/gcc.lua#L358

https://github.com/xmake-io/xmake/blob/2c67c028dffe3e456e8b9ac16296ef042629b88b/xmake/modules/core/tools/gcc.lua#L499

I think we should use `toolchain:add("runenvs", ..)`to add LD_LIBRARY_PATH/DYLD_LIBRARY_PATH in clang/llvm toolchains. It will be better.

xmake/modules/core/tools/clang.lua
```diff
@@ -207,6 +207,19 @@ function _get_llvm_rootdir(self)
     return llvm_rootdir or nil
 end
 
+-- get llvm target triple
+function _get_llvm_target_triple(self)
+    local llvm_targettriple = _g._LLVM_TARGETTRIPLE
+    if llvm_targettriple == nil then
+        local outdata = try { function() return os.iorun(self:program() .. " -print-target-triple") end }
```
waruqi, line 214: we need pass `self:runenvs()` to os.iorunv

waruqi: and please use os.iorunv

xmake/modules/core/tools/clang.lua
Arthapz: it has been tested yes, but i'll try with your suggestions

Arthapz:
mmmh -Xlinker rpath -Xlinker for macOS doesn't seem to work

xmake/modules/core/tools/clang.lua
Arthapz: -install_name isn't automatically handled by the gcc inheritance ?

xmake/modules/core/tools/clang.lua
```diff
@@ -267,8 +280,25 @@ function nf_runtime(self, runtime, opt)
...
+                    if triple_libdir then
+                        maps["c++_static"] = table.join(maps["c++_static"], "-L" .. triple_libdir)
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "-L" .. triple_libdir)
+                    end
+                    -- add rpath to avoid the user need to set LD_LIBRARY_PATH by hand
+                    maps["c++_shared"] = table.join(maps["c++_shared"], nf_rpathdir(self, libdir))
+                    if triple_libdir then
+                        maps["c++_shared"] = table.join(maps["c++_shared"], nf_rpathdir(self, triple_libdir))
+                    end
+                    if target:kind() == "shared" and self:is_plat("macosx", "iphoneos", "watchos") then
```
waruqi, line 298: target:is_shared

xmake/modules/core/tools/clang.lua
```diff
@@ -267,8 +280,25 @@ function nf_runtime(self, runtime, opt)
...
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "-install_name")
+                        maps["c++_shared"] = table.join(maps["c++_shared"], "@rpath/" .. path.filename(target:filename()))
```
waruqi, line 300: remove path.filename

PR #4806, opened 2024-03-07
Fix clang scan deps modules mac os

Arthapz:
on some configs, xmake will try to use clang-scan-deps++ which can't exists so the fallback will always be used


xmake/rules/c++/modules/modules_support/clang/compiler_support.lua
```diff
@@ -218,7 +218,7 @@ function get_clang_scan_deps(target)
             local dir = path.directory(program)
             local basename = path.basename(program)
             local extension = path.extension(program)
-            program = (basename:gsub("clang", "clang-scan-deps")) .. extension
+            program = (basename:gsub("+", ""):gsub("clang", "clang-scan-deps")) .. extension
```
waruqi, line 221: `+` is lua pattern, it should be `%+`, or you can use `basename:rtrim("+")`

PR #4819, opened 2024-03-10
Improve hello world examples with good practices

Ttibsi:

Updating the hello world examples to not use `namespace std` or `std::endl` as they're considered bad practices.

### DIFF
--- a/xmake/templates/c++/console/project/src/main.cpp
+++ b/xmake/templates/c++/console/project/src/main.cpp
@@ -1,9 +1,6 @@
 #include <iostream>
 
-using namespace std;
-
-int main(int argc, char** argv)
-{
-    cout << "hello world!" << endl;
+int main(int argc, char** argv) {
+    std::cout << "hello world!\n";
     return 0;
 }
--- a/xmake/templates/c++/shared/project/src/main.cpp
+++ b/xmake/templates/c++/shared/project/src/main.cpp
@@ -1,9 +1,7 @@
 #include "foo.h"
 #include <iostream>
 
-using namespace std;
-
 int main(int argc, char** argv) {
-    cout << "add(1, 2) = " << add(1, 2) << endl;
+    std::cout << "add(1, 2) = " << add(1, 2) << "\n";
     return 0;
 }
--- a/xmake/templates/c++/static/project/src/main.cpp
+++ b/xmake/templates/c++/static/project/src/main.cpp
@@ -1,9 +1,7 @@
 #include "foo.h"
 #include <iostream>
 
-using namespace std;
-
 int main(int argc, char** argv) {
-    cout << "add(1, 2) = " << add(1, 2) << endl;
+    std::cout << "add(1, 2) = " << add(1, 2) << "\n";
     return 0;
 }


xmake/templates/c++/console/project/src/main.cpp
```diff
@@ -1,9 +1,6 @@
 #include <iostream>
 
-using namespace std;
-
-int main(int argc, char** argv)
-{
-    cout << "hello world!" << endl;
+int main(int argc, char** argv) {
+    std::cout << "hello world!\n";
```
waruqi, line 4: Why is it best practice to use `\n` instead of endl?

qudix: In this case it doesn't really matter, but `endl` also flushes the buffer.

waruqi: But for this example program, the endl is necessary, and we need to refresh the buffer so that the user can see "hello wolrd" in time.

Ttibsi: No, it doesn’t need to flush, it’s just printing to stdout. Using `std::endl` also adds performance overhead that, while it doesn’t matter for hello world, isn’t something that should be encouraged.

 If you explicitly need to flush, use `std::flush`. But that isn’t needed here

waruqi: This is just a hello world example program, we don't need to be concerned about performance, we just need to make sure that the program outputs correctly. We don't need to over-optimise, and we don't need to fearlessly optimise early.

waruqi: At least for the current hello world, endl is more versatile than `\n` across platforms and ensures more accurate output. On windows, line breaks should be `\r\n` rather than `\n`.

Ttibsi: Heads up, `std::endl` doesn’t automatically become `\r\n` on windows [according to the standard](https://eel.is/c++draft/ostream.manip#2). Relying on it for that is UB.

Ttibsi: Heads up, `std::endl` doesn’t automatically become `\r\n` on windows [according to the standard](https://eel.is/c++draft/ostream.manip#2). Relying on it for that is UB.

waruqi: We don't need to rely on it, we just need to know that endl ensures correct output on different platforms. Instead of using hard-coded \n

waruqi: Relying on \n, on the other hand, does not guarantee that it will be handled correctly on all platforms. This is especially true for windows, which, although mostly fine, is not a best practice, especially for the current hello world

SirLynix: the C++ standard says that endl is equal to writing "\n" + flushing the buffer, the translation between \n and \r\n happens at a lower level.

the C++ standard also says that std::cout is flushed on program exit, nothing bad can happen with using "\n".

However, it doesn't matter much. We can keep `std::endl` without much issue, but we should at least get rid of `using namespace std;` in my opinion.

Halalaluyafail3: `std::endl` is not any more portable in this regard. It outputs `\n`, then flushes. The handling of the correct end of line sequence is done for you by the stream itself. It being more portable is just a myth and this is mentioned in the cpp core guidelines: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slio50-avoid-endl.
> endl does not take care of producing a platform specific end-of-line sequence (like "\r\n" on Windows). So for a string stream, s << endl just inserts a single character, '\n'.

PR #4830, opened 2024-03-13
add partitions_implunit test

waruqi:

------

### DIFF
--- a/tests/projects/c++/modules/partitions_implunit/BasicPlane.Figures-Point.ixx
+++ b/tests/projects/c++/modules/partitions_implunit/BasicPlane.Figures-Point.ixx
@@ -1,0 +1,5 @@
+export module BasicPlane.Figures:Point;
+
+export struct Point {
+    int x, y;
+};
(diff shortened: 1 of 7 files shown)


Arthapz:
in this case it's not a partition implementation unit, but a module implementation unit

a partition implementation unit look like this

```cpp
// Foo.cpp
export module Foo;

import :Impl; // can't export import Impl because Impl is a implementation partition unit

export void foo() {
    bar();
}
```

```cpp
// Foo-Impl.mpp
module Foo:Impl;

export void bar() {
    std::cout << "Hello world" << std::endl;
}
```

Arthapz:
> > a partition implementation unit look like this
> 
> like this example? https://github.com/xmake-io/xmake/tree/master/tests/projects/c%2B%2B/modules/partitions2

yes, but we should remove the export on "export module math:details;  // math:details implementation partition"

waruqi:
> > > a partition implementation unit look like this
> > 
> > 
> > like this example? https://github.com/xmake-io/xmake/tree/master/tests/projects/c%2B%2B/modules/partitions2
> 
> yes, but we should remove the export on "export module math:details; // math:details implementation partition"

done.

Arthapz:
ah we should remove the "export" on bar function too according to https://learn.microsoft.com/en-us/cpp/build/reference/internal-partition?view=msvc-170

PR #4876, opened 2024-03-25
Add package:vrunv

SirLynix:
see https://github.com/xmake-io/xmake-repo/pull/3606


waruqi:
We should not add special package:vrunv, it will cause to modify lots of package test scripts.

try this patch. https://github.com/xmake-io/xmake/pull/4879

SirLynix:
It works great! Thank you

Issue #4889, opened 2024-03-27
os.exec misbehaves with programs trapping SIGINT


dkaszews:
### Xmake Version

xmake v2.8.9+20240321

### Operating System Version and Architecture

Ubuntu 22.04 aarch64

### Describe Bug

I am using xmake to compile RISC-V code and have created a phony target to run it with [spike simulator](https://github.com/riscv-software-src/riscv-isa-sim). To exit spike, you need to first enter interactive mode with Ctrl+C, then run `quit`. Under xmake, the Ctrl+C just kills xmake, so spike gets detached from but kept alive, making it spam I/O errors and impossible to exit without `kill` as anything you type goes into shell.

To reproduce, use the test script below:

```shell
$ ./trap.sh
Still alive
^CType "quit" to exit
quit

$ xmake run trap
Still alive
^CType "quit" to exit
$
trap.sh: line 5: read: read error: 0: Input/output error
Still alive
trap.sh: line 5: read: read error: 0: Input/output error
Still alive
trap.sh: line 5: read: read error: 0: Input/output error
Still alive
quit
Command 'quit' not found
trap.sh: line 5: read: read error: 0: Input/output error
Still alive
trap.sh: line 5: read: read error: 0: Input/output error
Still alive
...
```

### Expected Behavior

Xmake should trap `SIGINT` for the duration of `os.run` and wait for the spawned process to exit on its own, or send it `SIGTERM`/`SIGKILL` on repeated offences (configurable?). Other programs which trap `SIGKILL` include:

* vim - ignores it for most part; may be used with xmake with phony targets running git or opening files
* bitbake - requires second or third press which will not arrive under xmake; toolchain for Linux images

### Project Configuration

xmake.lua
```lua
target("trap")
    set_kind("phony")
    -- TODO: Ctrl-C does not kill inner process
    on_run(function (target)
        os.exec("$(projectdir)/trap.sh")
    end)
```

trap.sh
```shell
#!/usr/bin/env bash
function handler { echo 'Type "quit" to exit'; }
trap handler SIGINT
while true; do
    read -t 3 -p $'Still alive\n' x || continue
    echo $x | grep -q 'quit' && break
done
```

### Additional Information and Error Logs

Traping signals in Lua can be seen here: https://stackoverflow.com/a/34409274/2894535

waruqi:
[...]

PR #4895, opened 2024-03-28
Implement os.SIGDFL and os.SIGIGN

dkaszews:
Fixes: #4889 

Allows user to reset the previously overwritten `os.signal` or ignore it completely (not passed to process, same as with `{ exclusive = true }`.

Opens:
* More descriptive names like `signal_default/reset` and `signal_ignore` instead of Unix's `SIGDFL` and `SIGIGN`?
* Tests? May be tricky to test signals, would need to run xmake in a subprocess of its own.
* Add documentation

### DIFF
--- a/core/src/xmake/os/signal.c
+++ b/core/src/xmake/os/signal.c
@@ -49,6 +49,12 @@ typedef enum __xm_os_signal_e {
     XM_OS_SIGINT = 1
 }xm_os_signal_e;
 
+typedef enum __xm_os_signal_handler_e {
+    XM_OS_SIGFUN = 0,
+    XM_OS_SIGDFL = 1,
+    XM_OS_SIGIGN = 2,
+}xm_os_signal_handler_e;
+
 /* //////////////////////////////////////////////////////////////////////////////////////
  * globals
  */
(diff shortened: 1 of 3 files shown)


core/src/xmake/os/signal.c
```diff
@@ -49,6 +49,12 @@ typedef enum __xm_os_signal_e {
     XM_OS_SIGINT = 1
 }xm_os_signal_e;
 
+typedef enum __xm_os_signal_handler_e {
+    XM_OS_SIGFUN = 0,
+    XM_OS_SIGDFL = 1,
+    XM_OS_SIGIGN = 2,
+}xm_os_signal_handler_e;
```
waruqi, line 56: you should add them to xm_os_signal_e. and `XM_OS_SIGINT` has been one.

xmake/core/base/os.lua
```diff
@@ -56,6 +56,8 @@ os.SYSERR_NOT_ACCESS  = 3
 
 -- signal code
 os.SIGINT = 1
+os.SIGDFL = 1
```
waruqi, line 59: it will conflict with os.SIGINT

waruqi:
> More descriptive names like signal_default/reset and signal_ignore instead of Unix's SIGDFL and SIGIGN?

or we can use 

```lua
os.signal(os.SIGINT, {reset = true})
os.signal(os.SIGINT, {ignore = true})
```

waruqi:
> Tests? May be tricky to test signals, would need to run xmake in a subprocess of its own.

If you can help to add some tests, it will be better.

waruqi:
> Add documentation

you can add doc in https://github.com/xmake-io/xmake-docs/blob/master/manual/builtin_modules.md#os

dkaszews:
> ```lua
> os.signal(os.SIGINT, {reset = true})
> os.signal(os.SIGINT, {ignore = true})
> ```

I don't quite like using dictionary just for named value, it's easy to make a typo and not get any feedback. I think cleaner would be to create one function for each:

| name | signal | xmake action | process signalled |
| - | - | - | - |
| `os.set_signal(signo, fun)` | `fun` | callback | yes |
| `os.ignore_signal(signo)` | `SIG_IGN` | no action | no |
| `os.reset_signal(signo)` | `SIG_DLF` | killed | yes(1) |

Notes:
1. If traps and does not exit, then it's orphaned detached from I/O as per issue

waruqi:
I wouldn't consider adding too much API to the os module for signal. please use only `os.signal`

dkaszews:
> Or use a separate signal module.

Understood, how about putting it in `process` since it's related and has relatively small API?

waruqi:
> > Or use a separate signal module.
> 
> Understood, how about putting it in since it's related and has relatively small API?`process`

no, process is only used to run and manage a single child process.

waruqi:
see https://github.com/xmake-io/xmake/pull/4908

PR #4917, opened 2024-04-01
package.tools.cmake: Use newer Android toolchain and support runtimes

SirLynix:

------

### DIFF
--- a/xmake/modules/package/tools/cmake.lua
+++ b/xmake/modules/package/tools/cmake.lua
@@ -427,16 +427,9 @@ function _get_configs_for_android(package, configs, opt)
             table.insert(configs, "-DANDROID_NATIVE_API_LEVEL=" .. ndk_sdkver)
         end
         -- https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_STL_TYPE.html
-        local supported_runtimes = {
-            "gnustl_static", "gnustl_shared",
-            "c++_static", "c++_shared",
-            "stlport_static", "stlport_shared"
-        }
-        for _, runtime in ipairs(supported_runtimes) do
-            if package:has_runtime(runtime) then
-                table.insert(configs, "-DCMAKE_ANDROID_STL_TYPE=" .. runtime)
-                break
-            end
+        local runtime = package:runtimes()
+        if runtime then
+            table.insert(configs, "-DCMAKE_ANDROID_STL_TYPE=" .. runtime)
         end
         if is_host("windows") and opt.cmake_generator ~= "Ninja" then
             local make = path.join(ndk, "prebuilt", "windows-x86_64", "bin", "make.exe")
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/cmake.lua
```diff
@@ -419,15 +419,26 @@ function _get_configs_for_android(package, configs, opt)
...
         if ndk_sdkver then
             table.insert(configs, "-DANDROID_PLATFORM=android-" .. ndk_sdkver)
             table.insert(configs, "-DANDROID_NATIVE_API_LEVEL=" .. ndk_sdkver)
         end
-        if ndk_cxxstl then
-            table.insert(configs, "-DANDROID_STL=" .. ndk_cxxstl)
+        -- https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_STL_TYPE.html
+        local supported_runtimes = {
+            "none", "system", 
+            "gabi++_static", "gabi++_shared",
```
waruqi, line 432: package:has_runtime only support 

```
"gnustl_static", "gnustl_shared",
"c++_static", "c++_shared",
"stlport_static", "stlport_shared"
```

SirLynix: I know but maybe it will be extended in the future, so I put all runtimes supported by this cmake option

waruqi: Please add them when xmake actually supports them. Also, the runtime names of cmake and ndk are not necessarily the same as the runtimes of xmake, and may need to be mapped, so don't add some unknown runtime names for now.

SirLynix: Fair enough, done

xmake/modules/package/tools/cmake.lua
```diff
@@ -419,15 +419,24 @@ function _get_configs_for_android(package, configs, opt)
...
+            "gnustl_static", "gnustl_shared",
+            "c++_static", "c++_shared",
+            "stlport_static", "stlport_shared"
+        }
+        for _, runtime in ipairs(supported_runtimes) do
+            if package:has_runtime(runtime) then
+                table.insert(configs, "-DCMAKE_ANDROID_STL_TYPE=" .. runtime)
+                break
+            end
         end
```
waruqi, line 440:
```lua
local runtime = package:runtimes()
if runtime then
    table.insert(configs, "-DCMAKE_ANDROID_STL_TYPE=" .. runtime)
end
```

SirLynix: what if multiple runtimes have been set?

waruqi: only one runtime. https://github.com/xmake-io/xmake/blob/d1ee61015c30cd08455b8d9bf85970b4ecaa079a/xmake/toolchains/ndk/xmake.lua#L36

```lua
set_runtimes("c++_static", "MT")
```

-> package:runtimes() only return `c++_static`

SirLynix: I got confused by the plural form (runtimes instead of runtime).

waruqi: For some toolchains and platforms that support multiple runtimes, it may have multiple values, but for android/msvc, only one value is possible

PR #4957, opened 2024-04-09
Fix issue with fetching packages when doing cross-compilation

### LINKED ISSUE #4935: Unable to use asio vcpkg dependency for MacOS (cross compilation) with xmake version later than 2.8.2
### Xmake Version

2.8.9

macOS Sonoma 14.4.1

### Describe Bug

When adding the vcpk::asio dependency to a project it fails to install asio during the config phase when issuing `xmake config -v -y --arch=x86_64`.
Switching to an older version of xmake (2.8.2) it works as expected.

### Expected Behavior

I'd expect it to behave as 2.8.2 and install and link with the asio library.
2.8.2 is the latest version I have found where it works as expected.

### Project Configuration

```lua
add_requires(
    "vcpkg::asio"
)

set_policy("build.ccache", false)

target("asiotest")
    set_kind("binary")
    add_files("src/main.cpp")

    add_packages(
        "vcpkg::asio")
```

Output:
```
> xmake-test % xmake config -v -y --arch=x86_64 
checking for SDK version of Xcode for macosx (x86_64) ... 14.4
checking for Minimal target version of Xcode for macosx (x86_64) ... 14.4
checking for zig ... no
checking for zig ... no
checking for nim ... no
checking for nim ... no
checking for unzip ... /usr/bin/unzip
checking for git ... /usr/bin/git
checking for gzip ... /usr/bin/gzip
checking for tar ... /usr/bin/tar
checking for vcpkg ... /Users/dan.engelbrecht/github/microsoft/vcpkg/vcpkg
/Users/dan.engelbrecht/github/microsoft/vcpkg/vcpkg install asio:x64-osx
Computing installation plan...
The following packages are already installed:
    asio:x64-osx@1.29.0
asio:x64-osx is already installed
Total install time: 5.79 us
asio provides CMake targets:

  # this is heuristically generated, and may not be correct
  find_package(asio CONFIG REQUIRED)
  target_link_libraries(main PRIVATE asio::asio)


error: fetch vcpkg::asio-latest failed!
  => install vcpkg::asio latest .. failed
error: install failed!
```

bionicbeagle:
This change changes the behavior with cross compilation when using packages. Previously the package would not properly fetch or install the cross-compiled packages.

Fixes #4935

### DIFF
--- a/xmake/core/package/package.lua
+++ b/xmake/core/package/package.lua
@@ -1948,7 +1948,7 @@ function _instance:fetch(opt)
         end
 
         -- fetch it from the system and external package sources (disabled for cross-compilation)
-        if not fetchinfo and system ~= false and not self:is_cross() then
+        if not fetchinfo and system ~= false then
             fetchinfo = self:_fetch_library({system = true, require_version = require_ver, external = external, force = opt.force})
             if fetchinfo then
                 is_system = true


DanEngelbrecht (approved):

waruqi:
It will break something, It may introduce more host libraries from on_fetch as well as pkg-config, causing linking errors. Only vcpkg/conan supports fetching cross-compiled libraries.

If we want to improve this, we need to consider more details.

bionicbeagle:
I see. I guess I would consider making conan/vcpkg functional an improvement over the current behaviour which is a very obscure failure mode with every package manager. 

I'm not sufficiently familiar with the xmake internals to propose a solution but it seems like it would be useful to have a way to discriminate between the different package managers wrt their cross-compilation capabilities so that a better error could be produced in the cases where it is not supported, and let the package managers which support cross-compilation do their thing.

bionicbeagle:
Thanks. I'm wondering though if it would be valuable to at least make vcpkg/conan work in the meantime? As far as I understand the net impact of this change will be:

- xrepo/vcpkg/conan will work for cross-compilation just as it used to
- other package managers may fail in some manner

whereas the current code leaves us with:

- xrepo works for cross-compilation
- all other package managers end up with the obscure error in #4935, even if they are able to handle cross-compilation

with this change we are able to build for MacOS with latest xmake whereas without it we have to remain on 2.8.2 which is not ideal as we'd really like to leverage newer functionality.

waruqi:
> Thanks. I'm wondering though if it would be valuable to at least make vcpkg/conan work in the meantime? As far as I understand the net impact of this change will be:
> 
> * xrepo/vcpkg/conan will work for cross-compilation just as it used to
> * other package managers may fail in some manner

No, the impact of the current patch may be a bit severe, and more users will fail to compile due to finding the wrong system libraries from pkg-config than vcpkg/conan.

waruqi:
try https://github.com/xmake-io/xmake/pull/4963

PR #4927, opened 2024-04-03
add support for --runtimes for autoconf, cmake, make and meson packages

Arthapz:
draft because depends of https://github.com/xmake-io/xmake/pull/4921

PR #5025, opened 2024-04-27
fix CMake package defaultflags and runtimes support

Arthapz:
my last PR https://github.com/xmake-io/xmake/pull/4927 broke some cmake libraries from xrepo on windows (CMAKE_CXX_FLAGS override default flags)
so here a new version which use CMAKE_CXX_FLAGS_RELEASE and autodetect default flags

### DIFF
--- a/xmake/modules/package/tools/cmake.lua
+++ b/xmake/modules/package/tools/cmake.lua
@@ -723,43 +723,8 @@ function _get_configs_for_install(package, configs, opt)
     end
 end
 
--- get configs
-function _get_configs(package, configs, opt)
-    configs = configs or {}
-    opt._configs_str = string.serialize(configs, {indent = false, strip = true})
-    _get_configs_for_install(package, configs, opt)
-    _get_configs_for_generator(package, configs, opt)
-    if package:is_plat("windows") then
-        _get_configs_for_windows(package, configs, opt)
-    elseif package:is_plat("android") then
-        _get_configs_for_android(package, configs, opt)
-    elseif package:is_plat("iphoneos", "watchos") or
-        -- for cross-compilation on macOS, @see https://github.com/xmake-io/xmake/issues/2804
-        (package:is_plat("macosx") and (get_config("appledev") or not package:is_arch(os.subarch()))) then
-        _get_configs_for_appleos(package, configs, opt)
-    elseif package:is_plat("mingw") then
-        _get_configs_for_mingw(package, configs, opt)
-    elseif package:is_plat("wasm") then
-        _get_configs_for_wasm(package, configs, opt)
-    elseif package:is_cross() then
-        _get_configs_for_cross(package, configs, opt)
-    elseif package:config("toolchains") then
-        -- we still need find system libraries,
-        -- it just pass toolchain environments if the toolchain is compatible with host
-        if _is_toolchain_compatible_with_host(package) then
-            _get_configs_for_host_toolchain(package, configs, opt)
-        else
-            _get_configs_for_cross(package, configs, opt)
-        end
-    else
-        _get_configs_for_generic(package, configs, opt)
-    end
-    _get_configs_for_default_flags(package, configs, opt)
-    return configs
-end
-
 function _get_default_flags(package, configs, buildtype, opt)
-    local cmake_default_flags = _g.cmake_default_flags
+    local cmake_default_flags = _g.cmake_default_flags and _g.cmake_default_flags[buildtype]
     if not cmake_default_flags then
         local tmpdir = path.join(os.tmpdir(), package:name(), package:mode())
         local dummy_cmakelist = path.join(tmpdir, "CMakeLists.txt")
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/cmake.lua
```diff
@@ -755,6 +723,92 @@ function _get_configs_for_install(package, configs, opt)
     end
 end
 
+function _get_default_flags(package, configs, buildtype, opt)
+    local cmake_default_flags = _g.cmake_default_flags and _g.cmake_default_flags[buildtype]
+    if not cmake_default_flags then
+        local tmpdir = path.join(os.tmpdir(), package:name(), package:mode())
```
waruqi, line 729: `local tmpdir = os.tmpfile() .. ".dir"`

xmake/modules/package/tools/cmake.lua
```diff
@@ -755,6 +723,92 @@ function _get_configs_for_install(package, configs, opt)
...
+        table.join2(cmake_default_flags.cxxflags, outdata:match(format("CMAKE_CXX_FLAGS_%s is (.-)\n", buildtype)):replace("/MDd", ""):replace("/MD", ""):split(" "))
+        cmake_default_flags.ldflags = outdata:match("CMAKE_EXE_LINKER_FLAGS is (.-)\n"):split(" ")
+        table.join2(cmake_default_flags.ldflags, outdata:match(format("CMAKE_EXE_LINKER_FLAGS_%s is (.-)\n", buildtype)):split(" "))
+        cmake_default_flags.shflags = outdata:match("CMAKE_SHARED_LINKER_FLAGS is (.-)\n"):split(" ")
+        table.join2(cmake_default_flags.shflags, outdata:match(format("CMAKE_SHARED_LINKER_FLAGS_%s is (.-)\n", buildtype)):split(" "))
+        cmake_default_flags.arflags = outdata:match("CMAKE_STATIC_LINKER_FLAGS is (.-)\n"):split(" ")
+        table.join2(cmake_default_flags.arflags, outdata:match(format("CMAKE_STATIC_LINKER_FLAGS_%s is (.-)\n", buildtype)):split(" "))
+
+        _g.cmake_default_flags = _g.cmake_default_flags or {}
+        _g.cmake_default_flags[buildtype] = cmake_default_flags
```
waruqi, line 766: os.rm(tmpdir)

xmake/modules/package/tools/cmake.lua
```diff
@@ -755,6 +723,92 @@ function _get_configs_for_install(package, configs, opt)
...
+
+    message("CMAKE_SHARED_LINKER_FLAGS is ${CMAKE_SHARED_LINKER_FLAGS}")
+    message("CMAKE_SHARED_LINKER_FLAGS_%s is ${CMAKE_SHARED_LINKER_FLAGS_%s}")
+
+    message("CMAKE_STATIC_LINKER_FLAGS is ${CMAKE_STATIC_LINKER_FLAGS}")
+    message("CMAKE_STATIC_LINKER_FLAGS_%s is ${CMAKE_STATIC_LINKER_FLAGS_%s}")
+        ]], buildtype, buildtype, buildtype, buildtype, buildtype, buildtype, buildtype, buildtype, buildtype, buildtype))
+
+        local runenvs = opt.envs or buildenvs(package)
+        local cmake = find_tool("cmake", {envs = runenvs})
```
waruqi, line 750: we need not set runenvs when find cmake

xmake/modules/package/tools/cmake.lua
```diff
@@ -755,6 +723,92 @@ function _get_configs_for_install(package, configs, opt)
...
+        local _configs = table.join(configs, "-S " .. path.directory(dummy_cmakelist), "-B " .. tmpdir)
+        local _, outdata = os.iorunv(cmake.program, _configs, {envs = runenvs})
```
waruqi, line 752: `local outdata, errdata = os.iorunv`, this outdata is errdata here.

waruqi: and we should add try/catch, `local outdata = try {function () return os.iorunv end}`

and outdata maybe is nil.

```lua
if outdata then
end
```

xmake/modules/package/tools/cmake.lua
```diff
@@ -786,6 +840,20 @@ function _get_configs(package, configs, opt)
     else
         _get_configs_for_generic(package, configs, opt)
     end
+    local envs = _get_envs_for_default_flags(package, configs, opt)
+    local runtime_envs = _get_envs_for_runtime_flags(package, configs, opt)
+    if runtime_envs then
+        envs = envs or {}
+        for name, value in pairs(runtime_envs) do
+            envs[name] = table.join(envs[name], value)
```
waruqi, line 848: `join(envs[name] or {})`

xmake/modules/package/tools/cmake.lua
```diff
@@ -786,6 +840,20 @@ function _get_configs(package, configs, opt)
...
+        end
+    end
+    if envs then
+        for name, value in pairs(envs) do
+            envs[name] = table.concat(table.unique(value), " ")
```
waruqi, line 853: If value contains spaces, does it work?

xmake/modules/package/tools/cmake.lua
Arthapz: > `local outdata, errdata = os.iorunv`, this outdata is errdata here.

i don't know why, but cmake message output are on errdata

SirLynix: Probably because CMake fails to build the project as there's no project

Arthapz: okay, i fixed it, message() by defaut output on stderr, i added STATUS and now it output on stdout

waruqi:
It cannot get correct default flags for windows arm (cross).

https://github.com/xmake-io/xmake-repo/pull/4038#issuecomment-2116489448

waruqi:
https://github.com/xmake-io/xmake/pull/5115

xmake/modules/package/tools/cmake.lua
```diff
@@ -755,6 +723,95 @@ function _get_configs_for_install(package, configs, opt)
...
+    return envs
+end
+
+function _get_envs_for_runtime_flags(package, configs, opt)
+    local buildtype = _get_cmake_buildtype(package)
+    local envs = {}
+    local runtimes = package:runtimes()
+    if runtimes then
+        local fake_target = {is_shared = function(_) return false end,
+                             sourcekinds = function(_) return "c" end}
```
waruqi, line 804: why not use package? it will break if package toolchain is changed. 

https://github.com/xmake-io/xmake/issues/5542

PR #5079, opened 2024-05-09
improve msvc host #5077

waruqi:
https://github.com/xmake-io/xmake/issues/5077


waruqi:
```
xmake update -s github:xmake-io/xmake#msvc
```

PR #5082, opened 2024-05-10
try fixing build on msys2 gcc 14

c8ef:
- https://github.com/xmake-io/xmake-repo/actions/runs/9031275001/job/24817204958?pr=4001
- https://packages.msys2.org/package/mingw-w64-x86_64-gcc


waruqi:
thanks

PR #5142, opened 2024-05-24
Fix building xmake using xmake on mingw

tqfx:


core/src/xmake/process/openv.c
```diff
@@ -30,7 +30,7 @@
  */
 #include "prefix.h"
 #include "../io/prefix.h"
-#if defined(TB_CONFIG_OS_MACOSX) || defined(TB_CONFIG_OS_LINUX) || defined(TB_CONFIG_OS_BSD) || defined(TB_CONFIG_OS_HAIKU)
+#if defined(TB_CONFIG_OS_MACOSX) || defined(TB_CONFIG_OS_LINUX) || defined(TB_CONFIG_OS_BSD) || defined(TB_CONFIG_OS_HAIKU) || defined(__MINGW32__)
```
waruqi, line 33: TB_COMPILER_IS_MINGW

core/src/xmake/winos/logical_drives.c
```diff
@@ -81,5 +81,5 @@ tb_int_t xm_winos_logical_drives(lua_State* lua)
     data = tb_null;
 
     // ok
-    return 1;
+    return ok;
```
waruqi, line 84: the return value type is int

core/xmake.lua
```diff
@@ -65,15 +65,20 @@ option_end()
...
     add_cfuncs("readline")
     add_defines("XM_CONFIG_API_HAVE_READLINE")
 option_end()
 
 -- the curses option
 option("curses")
     set_description("Enable or disable curses library")
-    add_links("curses")
+    if is_plat("mingw") then
+        add_includedirs(path.join(get_config("mingw"), "include", "ncursesw"))
```
waruqi, line 77: mingw config maybe nil

core/xmake.lua
```diff
@@ -65,15 +65,23 @@ option_end()
...
+    add_cincludes("stdio.h", "readline/readline.h")
     add_cfuncs("readline")
     add_defines("XM_CONFIG_API_HAVE_READLINE")
 option_end()
 
 -- the curses option
 option("curses")
     set_description("Enable or disable curses library")
-    add_links("curses")
+    if is_plat("mingw") then
```
waruqi, line 76: we cannot use is_plat and get_config in option description scope. it does not work.

waruqi: we should handle it in before_check

https://github.com/tboox/tbox/blob/c6b0a56076941b8263e162c7fe7b0870ea44e09c/src/xmake.lua#L90

core/xmake.lua
```diff
@@ -65,15 +65,23 @@ option_end()
...
+        local mingw = get_config("mingw")
+        if mingw then
+            add_includedirs(path.join(mingw, "include", "ncursesw"))
```
waruqi, line 79: why not use `add_includedirs("ncursesw")` directly?

waruqi, line 79: we should include it in code.

```c
#ifdef TB_COMPILER_IS_MINGW
#   include <ncursesw/curses.h>
#else
#   include <curses.h>
#endif
```

and add it here

```lua
add_cincludes("ncursesw/curses.h")
```

PR #5183, opened 2024-06-03
Intel has changed its default path for icx

Altina-oz:
there is no "linux" anymore, for example: \\wsl.localhost\Ubuntu-24.04\opt\intel\oneapi\compiler\2024.1\bin

image


waruqi:
记得下次提到 dev branch

PR #5276, opened 2024-06-29
fix find vc6

ChrisCatCP:


SirLynix:
Both paths should be added, as it will not have the (x86) suffix on 32bits machines

xmake/modules/detect/sdks/find_vstudio.lua
```diff
@@ -343,10 +343,18 @@ function _find_vstudio(opt)
...
                 end
             end
             vcvarsall = find_file("vcvarsall.bat", paths) or find_file("vcvars32.bat", paths)
         end
+        if not vcvarsall and version == "6.0" then
+            local paths = {
+                "$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\Microsoft\\DevStudio\\6.0\\Products\\Microsoft Visual C++;ProductDir)\\Bin",
+                "$(reg HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\Microsoft\\VisualStudio\\6.0\\Setup\\Microsoft Visual C++;ProductDir)\\Bin"
+            }
+            vcvarsall = find_file("vcvars32.bat", paths)
```
waruqi, line 356: we should add these paths in https://github.com/xmake-io/xmake/blob/cb8554de6bc050bc325c0eb3e146409d5aae819d/xmake/modules/detect/sdks/find_vstudio.lua#L320

waruqi: and it should be only for x64

ChrisCatCP: WOW6432Node  only for x64

waruqi: > WOW6432Node only for x64

so we should add `if os.arch() == "x64"`

PR #5302, opened 2024-07-03
Improve Vala support

al1-ce:
When compiling multiple files valac fails to find symbols (i.e SomeClass from a.vala using in b.vala), changed build to process all source files at once

### DIFF
--- a/xmake/rules/vala/xmake.lua
+++ b/xmake/rules/vala/xmake.lua
@@ -116,17 +116,17 @@ rule("vala.build")
             -- if it's only a vala file
             if path.extension(sourcefile) == ".vala" then
                 table.insert(argv, path(sourcefile))
+                batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala %s", sourcefile)
             end
         end
 
-        batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala")
         batchcmds:vrunv(valac.program, argv)
     end)
 
-    on_buildcmd_file(function (target, batchcmds, sourcebatch, opt)
+    on_buildcmd_file(function (target, batchcmds, sourcefile, opt)
         -- Again, only vala files need special treatment
-        if path.extension(sourcebatch) == ".vala" then
-            local sourcefile_c = target:autogenfile((sourcebatch:gsub(".vala$", ".c")))
+        if path.extension(sourcefile) == ".vala" then
+            local sourcefile_c = target:autogenfile((sourcefile:gsub(".vala$", ".c")))
             local basedir = path.directory(sourcefile_c)
 
             batchcmds:mkdir(basedir)
@@ -134,7 +134,7 @@ rule("vala.build")
             local objectfile = target:objectfile(sourcefile_c)
             table.insert(target:objectfiles(), objectfile)
 
-            batchcmds:add_depfiles(sourcebatch)
+            batchcmds:add_depfiles(sourcefile)
             batchcmds:set_depmtime(os.mtime(objectfile))
             batchcmds:set_depcache(target:dependfile(objectfile))
 
--- a/xmake/rules/vala/xmake.lua
+++ b/xmake/rules/vala/xmake.lua
@@ -112,15 +112,27 @@ rule("vala.build")
 
         -- iterating through source files,
         -- otherwise valac would fail when compiling multiple files
+        local lastmtime = 0
+        local sourcefiles = {}
         for _, sourcefile in ipairs(sourcebatch.sourcefiles) do
             -- if it's only a vala file
             if path.extension(sourcefile) == ".vala" then
-                table.insert(argv, path(sourcefile))
+                local sourcefile_c = target:autogenfile((sourcefile:gsub(".vala$", ".c")))
                 batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala %s", sourcefile)
+                table.insert(argv, path(sourcefile))
+                table.insert(sourcefiles, sourcefile)
+                local sourcefile_c_mtime = os.mtime(sourcefile_c)
+                if sourcefile_c_mtime > lastmtime then
+                    lastmtime = sourcefile_c_mtime
+                end
             end
         end
 
-        batchcmds:vrunv(valac.program, argv)
+        if #sourcefiles > 0 then
+            batchcmds:vrunv(valac.program, argv)
+            batchcmds:add_depfiles(sourcefiles)
+            batchcmds:set_depmtime(lastmtime)
+        end
     end)
 
     on_buildcmd_file(function (target, batchcmds, sourcefile, opt)
@@ -134,12 +146,12 @@ rule("vala.build")
             local objectfile = target:objectfile(sourcefile_c)
             table.insert(target:objectfiles(), objectfile)
 
+            batchcmds:show_progress(opt.progress, "${color.build.object}compiling.c %s", sourcefile_c)
+            batchcmds:compile(sourcefile_c, objectfile, { configs = { force = { cflags = "-w" } } })
+
             batchcmds:add_depfiles(sourcefile)
             batchcmds:set_depmtime(os.mtime(objectfile))
             batchcmds:set_depcache(target:dependfile(objectfile))
-
-            batchcmds:show_progress(opt.progress, "${color.build.object}compiling.c %s", sourcefile_c)
-            batchcmds:compile(sourcefile_c, objectfile, { configs = { force = { cflags = "-w" } } })
         end
     end)
 
@@ -186,4 +198,3 @@ rule("vala")
     -- we attempt to extract symbols to the independent file and
     -- strip self-target binary if `set_symbols("debug")` and `set_strip("all")` are enabled
     add_deps("utils.symbols.extract")
-
--- a/xmake/rules/vala/xmake.lua
+++ b/xmake/rules/vala/xmake.lua
@@ -33,9 +33,9 @@ rule("vala.build")
         if not vapifile then
             local vapiname = target:values("vala.vapi")
             if vapiname then
-                vapifile = path.join(target:targetdir(), vapiname)
+                vapifile = path.absolute(path.join(target:targetdir(), vapiname))
             else
-                vapifile = path.join(target:targetdir(), target:name() .. ".vapi")
+                vapifile = path.absolute(path.join(target:targetdir(), target:name() .. ".vapi"))
             end
             target:data_set("vala.vapifile", vapifile)
         end
@@ -45,9 +45,9 @@ rule("vala.build")
         if not headerfile then
             local headername = target:values("vala.header")
             if headername then
-                headerfile = path.join(target:targetdir(), headername)
+                headerfile = path.absolute(path.join(target:targetdir(), headername))
             else
-                headerfile = path.join(target:targetdir(), target:name() .. ".h")
+                headerfile = path.absolute(path.join(target:targetdir(), target:name() .. ".h"))
             end
             target:data_set("vala.headerfile", headerfile)
         end
@@ -56,7 +56,7 @@ rule("vala.build")
             target:add("sysincludedirs", path.directory(headerfile), {public = true})
         end
     end)
-    before_buildcmd_files(function (target, batchcmds, sourcefiles_vala, opt)
+    before_buildcmd_files(function (target, batchcmds, sourcebatch, opt)
         -- Here we compile vala files into C code
 
         -- We have to compile entire project each time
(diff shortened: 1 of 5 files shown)


xmake/rules/vala/xmake.lua
```diff
@@ -58,67 +58,82 @@ rule("vala.build")
...
-        end
-        table.insert(argv, path(sourcefile_vala))
+
+        batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala")
         batchcmds:vrunv(valac.program, argv)
-        batchcmds:compile(sourcefile_c, objectfile)
+
+        for _, sourcefile_c in ipairs(sourcefiles_c) do
+            batchcmds:show_progress(opt.progress, "${color.build.object}compiling.c %s", sourcefile_c)
+            batchcmds:compile(sourcefile_c, target:objectfile(sourcefile_c))
```
waruqi, line 126: Using before_buildcmd_files will not compile them in parallel.

xmake/rules/vala/xmake.lua
```diff
@@ -58,67 +58,82 @@ rule("vala.build")
...
         -- add deps
-        batchcmds:add_depfiles(sourcefile_vala)
-        batchcmds:set_depmtime(os.mtime(objectfile))
-        batchcmds:set_depcache(target:dependfile(objectfile))
+        for _, sourcefile_vala in ipairs(sourcefiles_vala.sourcefiles) do
+            batchcmds:add_depfiles(sourcefile_vala)
+        end
+        for _, objectfile in ipairs(target:objectfiles()) do
+            batchcmds:set_depmtime(os.mtime(objectfile))
+            batchcmds:set_depcache(target:dependfile(objectfile))
```
waruqi, line 135: This will break incremental compilation.

xmake/rules/vala/xmake.lua
al1-ce: The problem is that you can't compile vala code separately

Imagine simple example:
```vala
// main.vala
int main() {
    myfunc(); // valac will throw error on this line
}
```

```vala
// b.vala
void myfunc() { stdout.print("impossible"); }
```

Vala compiler has no way of knowing if there's a `myfunc` defined anywhere if compiling files separately, that's why I'm making this PR

xmake/rules/vala/xmake.lua
al1-ce: Is there a way to set batchcmd depend from hash instead of modify time if to move compilation of **only** outputed C code into before_buildcmd_file?

xmake/rules/vala/xmake.lua
waruqi: But here you are also compiling them one by one instead of passing a list of source files to vala all at once, so how does vala know t their symbols?

Can you provide a full test project as attachment? I will look at it.

al1-ce: I was and am passing all vala files to compiler by adding them to argv

But, I refactored it and vala still has to go through entire project, but generated C code should be incrementally build now (and now Vala can properly directly import C symbols)

For project look at `tests/projects/vala/includec/src/main.vala` from my latest commit

xmake/rules/vala/xmake.lua
al1-ce: Doesn't matter anymore, see latest commit

xmake/rules/vala/xmake.lua
```diff
@@ -58,31 +56,40 @@ rule("vala.build")
             target:add("sysincludedirs", path.directory(headerfile), {public = true})
         end
     end)
-    before_buildcmd_file(function (target, batchcmds, sourcefile_vala, opt)
+    before_buildcmd_files(function (target, batchcmds, sourcefiles_vala, opt)
```
waruqi, line 59: please rename `sourcefiles_vala` to `sourcebatch`

xmake/rules/vala/xmake.lua
```diff
@@ -58,31 +56,40 @@ rule("vala.build")
...
 
-        -- get c source file for vala
-        local sourcefile_c = target:autogenfile((sourcefile_vala:gsub(".vala$", ".c")))
-        local basedir = path.directory(sourcefile_c)
+        local argv = {"-C", "-d", target:autogendir()}
+        -- iterating through source files,
+        -- otherwise valac would fail when compiling multiple files
+        for _, sourcefile_vala in ipairs(sourcefiles_vala.sourcefiles) do
+            -- if it's only a vala file
+            if string.match(sourcefile_vala, ".vala$") ~= nil then
```
waruqi, line 76: use path.extension

xmake/rules/vala/xmake.lua
```diff
@@ -103,22 +110,39 @@ rule("vala.build")
...
-        table.insert(argv, path(sourcefile_vala))
+
+        batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala")
         batchcmds:vrunv(valac.program, argv)
-        batchcmds:compile(sourcefile_c, objectfile)
+    end)
+
+    on_buildcmd_file(function (target, batchcmds, sourcefile_vala, opt)
+        -- Again, only vala files need special treatment
+        if string.match(sourcefile_vala, ".vala$") ~= nil then
```
waruqi, line 130: and here

xmake/rules/vala/xmake.lua
```diff
@@ -103,22 +110,39 @@ rule("vala.build")
...
-        batchcmds:set_depcache(target:dependfile(objectfile))
+            local objectfile = target:objectfile(sourcefile_c)
+            table.insert(target:objectfiles(), objectfile)
+
+            batchcmds:add_depfiles(sourcefile_vala)
+            batchcmds:set_depmtime(os.mtime(objectfile))
+            batchcmds:set_depcache(target:dependfile(objectfile))
+
+            batchcmds:show_progress(opt.progress, "${color.build.object}compiling.c %s", sourcefile_c)
+            batchcmds:compile(sourcefile_c, objectfile, { configs = { force = { cflags = "-w" } } })
```
waruqi, line 144: why add `-w`? we can use `set_warnings` in project xmake.lua

al1-ce: Because there's expected warnings from files that are generated from Vala and it's [recommended to disable them by Vala devs](http://docs.vala.dev/tutorials/programming-language/main/07-00-tools/07-01-valac.html):
```
The recommended approach is to just disable those warnings by passing options to the C compiler:
valac -X -w hello.vala
```
And this option only going to touch valac generated C files

waruqi:
Did you test other examples.

e.g. https://github.com/xmake-io/xmake/tree/dev/tests/projects/vala/staticlib

it does not work.

and others. https://github.com/xmake-io/xmake/tree/dev/tests/projects/vala

al1-ce:
Checking rn

xmake/rules/vala/xmake.lua
```diff
@@ -58,31 +56,40 @@ rule("vala.build")
...
+        -- incrementally built
 
         -- get valac
         import("lib.detect.find_tool")
         local valac = assert(find_tool("valac"), "valac not found!")
 
-        -- get c source file for vala
-        local sourcefile_c = target:autogenfile((sourcefile_vala:gsub(".vala$", ".c")))
-        local basedir = path.directory(sourcefile_c)
+        local argv = {"-C", "-d", target:autogendir()}
```
waruqi, line 71: `-d`? why not `-b basedir`?

xmake/rules/vala/xmake.lua
```diff
@@ -58,31 +56,40 @@ rule("vala.build")
...
+        -- iterating through source files,
+        -- otherwise valac would fail when compiling multiple files
+        for _, sourcefile_vala in ipairs(sourcefiles_vala.sourcefiles) do
+            -- if it's only a vala file
+            if string.match(sourcefile_vala, ".vala$") ~= nil then
+                table.insert(argv, path(sourcefile_vala))
```
waruqi, line 77: please add vala files list in argv tail

xmake/rules/vala/xmake.lua
al1-ce: - `-b` outputs all files into single directory
- `-d` outputs all files according to source tree
- 
We need `-d` because of these two lines

```lua
local sourcefile_c = target:autogenfile((sourcebatch:gsub(".vala$", ".c")))
local basedir = path.directory(sourcefile_c)
```

In this case for files in sub-directories `basedir` is going to be a sub-directory, for which `-b` flag doesn't account. It worked before because it was assumed that files were compiled one-by-one and directory for `-b` flag was a `basedir`

al1-ce:
Strange, vala/lua test doesn't compile even with upstream rule file...

al1-ce:
Implemented requested changes, all tests passing except for lua, but lua fails because of xmake (and maybe arch linux, it has headers who knows where) coz gcc fails to find lua includes from `-Llua` flag

Also fixed `vala.vapidir` and `vala.flags` not being applied and fixed `vala.vapifile` and `vala.header` locations being incorrect

xmake/rules/vala/xmake.lua
```diff
@@ -103,22 +99,48 @@ rule("vala.build")
...
+        end
+
+        batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala")
         batchcmds:vrunv(valac.program, argv)
-        batchcmds:compile(sourcefile_c, objectfile)
+    end)
+
+    on_buildcmd_file(function (target, batchcmds, sourcebatch, opt)
+        -- Again, only vala files need special treatment
+        if path.extension(sourcebatch) == ".vala" then
```
waruqi, line 128: please rename `sourcebatch` to `sourcefile` in on_buildcmd_file

sourcebatch is only for on_buildcmd_files

xmake/rules/vala/xmake.lua
```diff
@@ -103,22 +99,48 @@ rule("vala.build")
...
+        -- iterating through source files,
+        -- otherwise valac would fail when compiling multiple files
+        for _, sourcefile in ipairs(sourcebatch.sourcefiles) do
+            -- if it's only a vala file
+            if path.extension(sourcefile) == ".vala" then
+                table.insert(argv, path(sourcefile))
+            end
+        end
+
+        batchcmds:show_progress(opt.progress, "${color.build.object}compiling.vala")
```
waruqi, line 122: no sourcefile name in progress text. and Incremental compilation does not work.

```console
bash-5.2$ xmake
[ 25%]: compiling.vala
[100%]: build ok, spent 0.28s

bash-5.2$ xmake
[ 25%]: compiling.vala
[100%]: build ok, spent 0.28s
```

waruqi:
```console
$ cd staticlib

$ xmake
[ 20%]: compiling.vala
[ 40%]: archiving.release libmymath.a
[ 70%]: compiling.vala
[ 90%]: linking.release test
[100%]: build ok, spent 0.695s

$ xmake
[ 20%]: compiling.vala
[ 70%]: compiling.vala
[100%]: build ok, spent 0.517s
```

al1-ce: Because, as I explained, valac has to be called each time, since it acts on all of `.vala` files

Same for lack of fileblname, it compiles all vala files together

waruqi: you can show file progress one-by-one in for-loop.

al1-ce: That would be something like this though:
```
[50%] compiling vala: src/main.vala
[50%] compiling vala: src/other.vala
[50%] compiling vala: src/sub/mod.vala
...
```
And so on for all sourcefiles displayed at once

If that's fine I'll correct it to do that

waruqi: right, you can do it.

al1-ce: Ok, I'll also try getting it to at least not recompile if no files been modified

al1-ce:
Couldn't get valac build to run only when needed, but everything else should be correct now

waruqi:
But incremental compilation still doesn't work.

```console
bash-5.2$ xmake
[ 20%]: compiling.vala src/mymath.vala
[ 70%]: compiling.vala src/main.vala
[100%]: build ok, spent 0.588s

bash-5.2$ xmake
[ 20%]: compiling.vala src/mymath.vala
[ 70%]: compiling.vala src/main.vala
[100%]: build ok, spent 0.571s
```

al1-ce:
Oh, I think I'm starting to understand how batchcmds work, not completely though...
Hard to figure out, not that much documentation about it, except for examples which do not explain anything

waruqi:
> Oh, I think I'm starting to understand how batchcmds work, not completely though... Hard to figure out, not that much documentation about it, except for examples which do not explain anything

https://xmake.io/#/manual/custom_rule?id=custom-batch-compile-script-process-one-source-file-at-a-time

al1-ce:
> https://xmake.io/#/manual/custom_rule?id=custom-batch-compile-script-process-one-source-file-at-a-time
Again, shows how, but now why. Maybe worth adding some explaining comments to it?

Anyway, thanks for taking time to review and merge my PR

PR #5240, opened 2024-06-20
add staticlib2 test for modules

waruqi:
https://github.com/xmake-io/xmake/discussions/5229

PR #5376, opened 2024-07-22
Improve module objectfiles handling and moduleonly package

Arthapz:
this patch improve modules support by
    - ensuring that static / dynamic libraries public module objectfiles are beeing packed in the library
    - improving module culling so no std or std.compat module should be built if they are not imported
    - showing a warning when modules that are not std or std.compat are culled
    - raising an error when duplicated module name are detecteds


Arthapz:
need https://github.com/xmake-io/xmake/pull/5375

xmake/rules/c++/modules/modules_support/dependency_scanner.lua
```diff
@@ -200,7 +200,9 @@ function _get_package_modules(target, package, opt)
     for _, metafile in ipairs(metafiles) do
         package_modules = package_modules or {}
         local modulefile, name, metadata = _parse_meta_info(target, metafile)
-        package_modules[name] = {file = path.join(modulesdir, modulefile), metadata = metadata}
+        local package_data = package["_INFO"]
```
waruqi, line 203: please do not access private vars. we should call `package:libraryfiles()` or `package:get()` to get them.

Arthapz:
https://github.com/xmake-io/xmake/pull/5240 will no longer be necessary as this test is included in this PR

waruqi:
> #5240 will no longer be necessary as this test is included in this PR

I will close it when this patch is merged.

tests/projects/c++/modules/duplicate_name_detection/test.lua
```diff
@@ -0,0 +1,70 @@
...
+
+function _build()
+    try {function() 
+        if ci_is_running() then
+            os.iorun("xmake -rvD")
+        else
+            os.iorun("xmake -r")
+        end
+    end, catch {function(errors)
+        errors = errors.errors
```
waruqi, line 13:
```lua
errors = tostring(errors)
```

PR #5388, opened 2024-07-24
fixbug: config 'vs_sdkver' has no effect when generate vs project

jinke18:
config vs_sdkver

```sh
xmake f --vs_sdkver=14.0
```

but it has no effect when generate a vs project

```sh
xmake project -k vs2015
```

this pull request fix this bug~

### DIFF
--- a/xmake/plugins/project/vstudio/impl/vs201x.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x.lua
@@ -167,14 +167,16 @@ end
 function _make_targetinfo(mode, arch, target, vcxprojdir)
 
     -- init target info
-    local targetinfo = { mode = mode, arch = vsutils.vsarch(arch) }
+    local targetinfo = { mode = mode, arch = vsutils.vsarch(arch), sdkver=config.get("vs_sdkver") }
 
     -- get sdk version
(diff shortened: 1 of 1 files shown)


waruqi:
We shouldn't have modified it that way. First of all, 14.0 should be vs_toolset not vs_sdkver.

`xmake f --vs_toolset/vs_sdkver` -> `msvc toolchain/check` -> find_vstudio.lua -> get real toolset/sdkver -> generator/xmake build

so we should debug and check `find_vstudio.lua` 

https://github.com/xmake-io/xmake/blob/71ec32ea6787c1c4b215ea40e1d5ab40c1c4eedc/xmake/modules/detect/sdks/find_vstudio.lua#L125

jinke18:
I'm sorry I made a mistake when submitting the pull request message.

it should be
config vs_sdkver
```
xmake f --vs_sdkver=8.1
```

waruqi:
you still need debug find_vstudio.lua, if it works,  `vcvars.WindowsSDKVersion` will be changed to your configs after running `xmake f --vs_sdkver=xxx`

waruqi:
You can open an issue for this patch. I will look at it.

PR #5411, opened 2024-07-29
fixbug: for cases when WindowsSDKVersion variable is not available

jinke18:
find_vstudio.lua

for some cases, the variable `WindowsSDKVersion` is not available from `vcvarsall.bat`, i parse it from `WindowsSdkBinPath`, such as: `C:\\Program Files (x86)\\Windows Kits\\8.1\\bin`.

### DIFF
--- a/xmake/modules/detect/sdks/find_vstudio.lua
+++ b/xmake/modules/detect/sdks/find_vstudio.lua
@@ -191,7 +191,7 @@ function _load_vcvarsall(vcvarsall, vsver, arch, opt)
         -- then parse it from `WindowsSdkBinPath`, such as: `C:\\Program Files (x86)\\Windows Kits\\8.1\\bin`
         local WindowsSdkBinPath = variables["WindowsSdkBinPath"]
         if WindowsSdkBinPath then
-            WindowsSDKVersion = string.match(WindowsSdkBinPath, "\\(%d+%.?%d*)\\bin")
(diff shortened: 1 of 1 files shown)


xmake/modules/detect/sdks/find_vstudio.lua
```diff
@@ -186,6 +186,16 @@ function _load_vcvarsall(vcvarsall, vsver, arch, opt)
         if WindowsSDKVersion ~= "" then
             variables["WindowsSDKVersion"] = WindowsSDKVersion
         end
+    else
+        -- sometimes the variable `WindowsSDKVersion` is not available
+        -- then parse it from `WindowsSdkBinPath`, such as: `C:\\Program Files (x86)\\Windows Kits\\8.1\\bin`
+        local WindowsSdkBinPath = variables["WindowsSdkBinPath"]
+        if WindowsSdkBinPath then
+            WindowsSDKVersion = string.match(WindowsSdkBinPath, "\\(%d+%.?%d*)\\bin")
```
waruqi, line 194: I think `%d+%.%d+` will be better.

waruqi: and we should use path.directory and path.filename to get the parent directorty name, then match version.

jinke18: I don't known how to modify this pr. 
or i need to created a new one?

PR #5414, opened 2024-07-30
fixbug:nsis pack only support ascii path without blank space

jinke18:
when using xpack with nsis format, it will not work if there is path with chinese character or blank space.

for the first problem, i fix it by using ansi encoding. Since nsis is not supported in linux-like os, we do not need to consider conditions for linux?

And the second problem is fixed according to the grammar tips given by nsis.

### DIFF
--- a/xmake/plugins/pack/nsis/main.lua
+++ b/xmake/plugins/pack/nsis/main.lua
@@ -268,9 +268,7 @@ function _pack_nsis(makensis, package)
     local specvars_values = {}
     io.gsub(specfile, "(" .. pattern .. ")", function(_, name)
         table.insert(specvars_names, name)
-    end,
-        {encoding="ansi"}
-    )
+    end,{encoding="ansi"})
     for _, name in ipairs(specvars_names) do
         name = name:trim()
         if specvars_values[name] == nil then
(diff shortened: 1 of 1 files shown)


xmake/plugins/pack/nsis/main.lua
```diff
@@ -268,7 +268,9 @@ function _pack_nsis(makensis, package)
     local specvars_values = {}
     io.gsub(specfile, "(" .. pattern .. ")", function(_, name)
         table.insert(specvars_names, name)
-    end)
+    end,
+        {encoding="ansi"}
```
waruqi, line 272:
```
end, {encoding = "ansi"})
```

xmake/plugins/pack/nsis/main.lua
```diff
@@ -268,7 +268,7 @@ function _pack_nsis(makensis, package)
     local specvars_values = {}
     io.gsub(specfile, "(" .. pattern .. ")", function(_, name)
         table.insert(specvars_names, name)
-    end)
+    end,{encoding="ansi"})
```
waruqi, line 271: spaces, `, {`, `g = "`

xmake/plugins/pack/nsis/main.lua
```diff
@@ -288,7 +288,7 @@ function _pack_nsis(makensis, package)
     io.gsub(specfile, "(" .. pattern .. ")", function(_, name)
         name = name:trim()
         return specvars_values[name]
-    end)
+    end, {encoding="ansi"})
```
waruqi, line 291: spaces, ` = `

PR #5416, opened 2024-07-30
add /sdl, /Zc:inline and ExceptionHandling flag support for `xmake project -k vs`

jinke18:
i add /sdl and /Zc:inline flag support for `xmake project -k vs`

### DIFF
--- a/xmake/plugins/project/vstudio/impl/vs201x_vcxproj.lua
+++ b/xmake/plugins/project/vstudio/impl/vs201x_vcxproj.lua
@@ -517,10 +517,10 @@ function _make_source_options_cl(vcxprojfile, flags, condition)
     if flagstr:find("[%-/]EH[asc]+%-?") then
         local args = flagstr:match("[%-/]EH([asc]+%-?)")
         -- remove the last arg if flag endwith `-`
-        if args:endswith("-") then
+        if args and args:endswith("-") then
             args = args:sub(1, -2)
         end
-        if args:find("a", 1, true) then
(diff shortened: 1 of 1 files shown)


xmake/plugins/project/vstudio/impl/vs201x_vcxproj.lua
```diff
@@ -494,11 +494,50 @@ function _make_source_options_cl(vcxprojfile, flags, condition)
...
+        else
+            vcxprojfile:print("<RemoveUnreferencedCodeData%s>true</RemoveUnreferencedCodeData>", condition)
+        end
+    end
+
+    -- make ExceptionHandling flag:
+    if flagstr:find("[%-/]EH[asc]+%-?") then
+        local args = flagstr:match("[%-/]EH([asc]+%-?)")
+        -- remove the last arg if flag endwith `-`
+        if arg:sub(-1, -1) == "-" then
```
waruqi, line 520: arg? args? please use `:endswith("-")` instead of `sub`

xmake/plugins/project/vstudio/impl/vs201x_vcxproj.lua
```diff
@@ -494,11 +494,50 @@ function _make_source_options_cl(vcxprojfile, flags, condition)
...
+            args = args:sub(1, -2)
+        end
+        if args:find("a") then
```
waruqi, line 523: args may be nil, `if args and args:find("a", 1, true)`

Issue #5542, opened 2024-08-29
Empty toolchain while attempting to install packages using windows wasm


Chi-EEE:
### Xmake Version

xmake v2.9.4

### Operating System Version and Architecture

Windows 10 22H2

### Describe Bug

Empty toolchain while attempting to install packages using windows wasm.

[The toolchains list was empty here:](https://github.com/xmake-io/xmake/blob/dev/xmake/core/platform/platform.lua#L154)

### Expected Behavior

Finds the toolchain for wasm (emcc@emscripten)

### Project Configuration

Project folder where the error occurred
[platform.lua.zip](https://github.com/user-attachments/files/16801385/platform.lua.zip)

#### Steps to replicate:

1. Install fresh xmake

2. Run `xmake f -p wasm -vD` in the project

### Additional Information and Error Logs

```
checking for em++.bat ... C:\Users\admin\AppData\Local\.xmake\packages\e\emscripten\3.1.55\e632956fdb43439b94196f0dfdac70b7\upstream\emscripten\em++.bat
checking for the linker (ld) ... em++.bat
error: @programdir\core\package\package.lua:2364: cannot get program for ld
stack traceback:
    [C]: in function 'error'
    [@programdir\core\base\os.lua:973]:
    [@programdir\core\package\package.lua:2364]: in function '_generate_build_configs'
    [@programdir\core\package\package.lua:2589]: in function 'check_cxxsnippets'
    [....xmake\repositories\xmake-repo\packages\a\asap\xmake.lua:15]: in function 'script'
    [...dir\modules\private\action\require\impl\utils\filter.lua:114]: in function 'call'
    [...dir\modules\private\action\require\impl\actions\test.lua:41]:
    [...\modules\private\action\require\impl\actions\install.lua:432]:

  => install asap-fork 2023.04.21 .. failed
error: @programdir\core\main.lua:329: @programdir\modules\async\runjobs.lua:325: ...\modules\private\action\require\impl\actions\install.lua:494: install failed!
stack traceback:
    [C]: in function 'error'
    [@programdir\core\base\os.lua:973]:
    [...\modules\private\action\require\impl\actions\install.lua:494]: in function 'catch'
    [@programdir\core\sandbox\modules\try.lua:123]: in function 'try'
    [...\modules\private\action\require\impl\actions\install.lua:361]:
    [...modules\private\action\require\impl\install_packages.lua:496]: in function 'jobfunc'
    [@programdir\modules\async\runjobs.lua:241]:

[...]

PR #5543, opened 2024-08-29
improve linker for package #5542

waruqi:
https://github.com/xmake-io/xmake/issues/5542


waruqi:
this patch will fix some runtime flags for package toolchain. @Arthapz

Issue #5531, opened 2024-08-28
vulkansdk package not found with MinGW profile


yh-sb:
### Xmake Version

2.9.4

### Operating System Version and Architecture

Windows 11 23H2 22631.4037

### Describe Bug
This issue is specific only for `mingw` profile.

When I add `vulkansdk` package to my root xmake.lua, then I see the following error:
```
error: The packages(vulkansdk) not found, please run `xmake repo -u` first!
```

But I have installed VulkanSDK in my system and appropriate env variables are present:
```
> echo $env:VK_SDK_PATH
C:\VulkanSDK\1.3.290.0

> echo $env:VULKAN_SDK
C:\VulkanSDK\1.3.290.0
```

_Note: I am able to build project with vulkan using old CMake files. Just want to migrate to xmake._
I am also able to build project with msvc windows profile.

From what I found in modules/detect/sdks/find_vulkansdk.lua, neither `_find_vulkan_from_paths` nor `_find_vulkan_from_system` find vulkan.

### Expected Behavior

vulkansdk package found and project successfully built.

### Project Configuration

```
add_rules("mode.debug", "mode.release", "plugin.compile_commands.autoupdate", {outputdir = "build"})
set_defaultmode("debug")
set_targetdir("$(buildir)")
set_languages("c++20")

if is_plat("mingw") then
    add_ldflags("-static")
end

add_requires("glfw 3.4", "vulkan-hpp", "vulkansdk")

target("vulkan-app")
    add_packages("glfw", "vulkan-hpp", "vulkansdk")
    add_files(
        "main.cpp"
    )

```

### Additional Information and Error Logs

<details>
  <summary>xmake f -p mingw -vD -y</summary>

```
PS D:\dev\cpp\vulkan-app> xmake f -p mingw -vD -y
checking for architecture ... x86_64
checking for mingw directory ... C:\Program Files\mydevtools\MinGW-w64
checkinfo: cannot runv(unzip.exe -v), No such file or directory
checking for unzip ... no
checking for 7z ... C:\Program Files\mydevtools\xmake\winenv\bin\7z
checking for git ... ok
checkinfo: cannot runv(gzip.exe --version), No such file or directory
checking for gzip ... no
checking for ping ... ok
pinging the host(gitlab.com) ... 2 ms
pinging the host(github.com) ... 35 ms
pinging the host(gitee.com) ... 151 ms
git rev-parse HEAD
finding vulkan from pkgconfig ..
[...]

PR #5549, opened 2024-08-30
Fix vulkansdk searching for mingw platform

yh-sb:
#5531

Fix vulkan-1.lib searching when platform is mingw.


xmake/modules/detect/sdks/find_vulkansdk.lua
```diff
@@ -30,13 +30,14 @@ import("lib.detect.find_package")
 function _find_vulkan_from_paths(paths, opt)
     opt = opt or {}
     local arch = opt.arch or config.arch() or os.arch()
+    local plat = opt.plat or config.get("plat") or os.host()
```
waruqi, line 33: use config.plat()

PR #5634, opened 2024-09-18
Allows installdir, bindir and libdir to be configured via opt arguments

A2va:

------

### DIFF
--- a/xmake/modules/target/action/install/main.lua
+++ b/xmake/modules/target/action/install/main.lua
@@ -35,14 +35,10 @@ end
 
 function _get_target_bindir(target, opt)
     if not opt.installdir then
-        return (target:is_plat("windows", "mingw") or opt.true_bindir) and target:bindir() or target:libdir()
+        return target:bindir()
     end
-    assert(opt.libdir, "opt.libdir is missing")
     assert(opt.bindir, "opt.bindir is missing")
-
-    local bindir = path.join(opt.installdir, opt.bindir)
-    local libdir = path.join(opt.installdir, opt.libdir)
-    return (target:is_plat("windows", "mingw") or opt.true_bindir) and bindir or libdir
+    return path.join(opt.installdir, opt.bindir)
 end
 
 -- we need to get all deplibs, e.g. app -> libfoo.so -> libbar.so ...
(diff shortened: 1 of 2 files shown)


xmake/modules/target/action/install/main.lua
```diff
@@ -233,7 +243,8 @@ function _install_moduleonly(target, opt)
 end
 
 function main(target, opt)
-    local installdir = target:installdir()
+    opt = opt or {}
+    local installdir = opt.installdir and target:set("installdir", opt.installdir) or target:installdir()
```
waruqi, line 247: we should not call `target:set` to modify target installdir configuration.

A2va: I was wondering if I could save the installdir of a target, then set it to the old value at the end of the function.
The other solution that comes to mind is to add an opt argument to installdir, libdir and bindir of target.

Which solution do you prefer?

waruqi: why not use `local installdir = opt.installdir and opt.installdir or target:installdir()`?

A2va: Because when `target:installdir(opt.installdir)` is called, installdir is not set in the target (or not the right value). One solution would be to join installdir and bindir each time, but as this is already what bindir does, I haven't considered this option.

waruqi: you can add `function _get_installdir(target, opt)` to wrap it. then get them each time.

waruqi: like this 

https://github.com/xmake-io/xmake/blob/b9385560675a37623b2d4d385b6290b87b824d66/xmake/plugins/pack/batchcmds.lua#L28-L62

xmake/modules/target/action/install/main.lua
```diff
@@ -123,7 +143,7 @@ end
 
 -- install shared libraries
 function _install_shared_libraries(target, opt)
-    local bindir = target:is_plat("windows", "mingw") and target:bindir() or target:libdir()
+    local bindir = _get_target_bindir(target, opt)
```
waruqi, line 146: Please do not mix libdir and bindir.

https://github.com/xmake-io/xmake/blob/9563841bfc304874cc9d308b1e29bd6cde76631a/xmake/plugins/pack/batchcmds.lua#L247

xmake/modules/target/action/install/main.lua
```diff
@@ -158,7 +178,7 @@ function _update_install_rpath(target, opt)
     if target:is_plat("windows", "mingw") then
         return
     end
-    local bindir = target:bindir()
+    local bindir = _get_target_bindir(target, table.join(opt, {true_bindir = true}))
```
waruqi, line 181: please remove `true_bindir` argument.

PR #5733, opened 2024-10-18
Update cmake_importfiles

Yc7521:
fix: update CMake import file generation for architecture handling

- Improved TARGET_PTRBYTES calculation to support additional architectures.
- On DLL platforms, use IMPORTED_LOCATION to reference the dependent DLL file and IMPORTED_IMPLIB for the associated lib file.


xmake/modules/target/action/install/cmake_importfiles.lua
```diff
@@ -130,6 +130,16 @@ function _install_cmake_targetfile(target, installdir, filename, opt)
             local value = builtinvars[variable]
             return type(value) == "function" and value() or value
         end)
+        local libfile = path.filename(target:targetfile())
+        local postfix = is_mode("debug") and "DEBUG" or "RELEASE"
+        if target:is_shared() and (_get_libfile(target, installdir) ~= libfile) then
+            -- On DLL platforms, the import library is named differently from the target file
+            content = content:gsub("# IMPORTED_IMPLIB_" .. postfix, "IMPORTED_IMPLIB_" .. postfix)
+            content = content:gsub(
+                "IMPORTED_LOCATION_" .. postfix .. " \"${_IMPORT_PREFIX}/lib/.-\"",
```
waruqi, line 139: $ is lua pattern character.

Yc7521: If the pattern character is not at the beginning or end of the string, it matches the character literally.

xmake/modules/target/action/install/cmake_importfiles.lua
```diff
@@ -130,6 +130,16 @@ function _install_cmake_targetfile(target, installdir, filename, opt)
...
         end)
+        local libfile = path.filename(target:targetfile())
+        local postfix = is_mode("debug") and "DEBUG" or "RELEASE"
+        if target:is_shared() and (_get_libfile(target, installdir) ~= libfile) then
+            -- On DLL platforms, the import library is named differently from the target file
+            content = content:gsub("# IMPORTED_IMPLIB_" .. postfix, "IMPORTED_IMPLIB_" .. postfix)
+            content = content:gsub(
+                "IMPORTED_LOCATION_" .. postfix .. " \"%%${_IMPORT_PREFIX}/lib/.-\"",
+                "IMPORTED_LOCATION_" .. postfix .. " \"%%${_IMPORT_PREFIX}/bin/" .. libfile .. "\""
+            )
```
Yc7521, line 141: But why can't this match the string?

waruqi: `%%$` -> `%$`

```console
$ xmake l
xmake> ("${xxx}"):gsub("%${xxx}", "aaa")
< 1: "aaa"
< 2: 1
```

PR #5749, opened 2024-10-22
fix linux-driver build err

ririyeye:
new version kernel add `-Wmissing-prototypes` and `-Wall`
We should add the function prototype otherwise the compilation will fail


waruqi:
please push to dev branch next.

Issue #3855, opened 2023-06-17
Cannot link to system libraries due to ABI mismatch when modules are used


DavidTruby:
### Xmake Version

2.7.9

### Operating System Version and Architecture

Gentoo Linux

### Describe Bug

When modules are enabled, xmake unconditionally passes a flag switching g++ to use the old pre-C++11 abi: 
 https://github.com/xmake-io/xmake/blob/68153d06e61342aa2be927116c9b7d4e40125945/xmake/rules/c%2B%2B/modules/modules_support/gcc.lua#L77
This causes failures to link, and also sometimes runtime failures, due to ABI mismatch as Linux distrobutions build with the new ABI. For example if calling a library function that takes a `std::string` from a library compiled without this flag you will either get a linker failure, or worse a runtime segfault, because `std::string` has a different implementation with or without the flag.

The old ABI is broken and non-standards conformant so it shouldn't really be being used anyway. If a specific ABI needs to be forced for some reason, we should flip the flag and pass `-D_GLIBCXX_USE_CXX11_ABI=1` to force the new ABI instead.

### Expected Behavior

Successful linkage with system libraries because both are using the new ABI

### Project Configuration

xmake.lua:
```lua
add_rules("mode.debug", "mode.release")

set_languages("c++latest")

add_requires("llvm", {system = true, kind = "library"})

target("test")
    set_kind("binary")
    add_files("src/*.cpp")
    add_packages("llvm")
    set_policy("build.c++.modules", true)
```
src/main.cpp:
```cpp
import <llvm/Support/raw_ostream.h>;
import <string>;

int main() {
  std::string s = "hello world";
  llvm::outs() << s << "\n";
}
```

waruqi:
just to solve this issue. https://github.com/xmake-io/xmake/issues/2716#issuecomment-1225057760

DavidTruby:
I think that fix isn't correct: that's what is introducing the `-D_GLIBCXX_USE_CXX11_ABI=0` that forces the old ABI. The problem with that is e.g. my `libLLVM.so` on my system is built with the new ABI (as it would be on any linux distro since about 2014) so if I link to that it breaks. 
I think the fix should be `-D_GLIBCXX_USE_CXX11_ABI=1` instead to force the new ABI

DavidTruby:
I tried making that change myself and it seems to cause other issues but I'm not sure why... Forcing the old ABI will cause issues other than linkage though (for example, the old ABI is not thread safe)

[...]

Issue #2716, opened 2022-08-23
Module compilation error


Latias94:
### Xmake Version

xmake v2.6.9+dev.0864717

### Operating System Version and Architecture

WSL2 Ubuntu 22.04

### Describe Bug

Test files and build logs are as below.   
[ModuleTest.zip](https://github.com/xmake-io/xmake/files/9404087/ModuleTest.zip)

It contains two targets: 
1. module include legacy header files (build success with msvc,but failed with gcc)
2. module partition

The code is following the book <Professional C++, 5th Edition>, so maybe there are compiler issues in it, but I don't dive into it.

### Expected Behavior

Build

### Project Configuration

```lua
set_languages("c++20")
target("hello")
    set_kind("binary")
    add_files("src/*.cpp","src/*.cppm")
```

### Additional Information and Error Logs

1.

```shell
~/study/cpp/01_Person via 🌙 v5.4.4 
23:32:23 ❯ xmake -rv
checking for platform ... linux
checking for architecture ... x86_64
checking for gcc ... /usr/bin/gcc
checking for dmd ... no
checking for ldc2 ... no
checking for gdc ... no
checking for zig ... no
checking for zig ... no
checking for gcc ... /usr/bin/gcc
checking for the c++ compiler (cxx) ... gcc
checking for /usr/bin/gcc ... ok
checking for flags (-fPIC) ... ok
checking for flags (gcc_modules_ts) ... ok
checking for flags (gcc_module_mapper) ... ok
checking for flags (gcc_dep_format) ... no
checking for flags (gcc_dep_file) ... no
checking for flags (gcc_dep_output) ... no
[  0%]: generating.cxx.module.deps src/test.cpp
checking for gcc ... /usr/bin/gcc
checking for the c compiler (cc) ... gcc
[  0%]: generating.cxx.module.deps src/Person.cppm
[ 18%]: generating.cxx.headerunit.bmi iostream
[ 18%]: generating.cxx.headerunit.bmi string
checking for flags (-std=c++20) ... ok
checking for flags (-fmodules-ts) ... ok
/usr/bin/gcc -m64 -std=c++20 -fmodules-ts -fmodule-mapper=/dev/shm/.xmake1000/220823/_3385625651234B2083BF312399354B50 -c -x c++-system-header iostream
[...]

PR #5869, opened 2024-11-22
Add support for libstdc++ standard library modules

Juvwxyz:
An installed source form of module std and std.compat has been merged into gcc trunk branch. It's time to add support for them.
[ \[Bug 106852\]\(c++lib-std-module\) - Implement C++23 P2465R3 Standard Library Modules std and std.compat](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106852)

GCC provides a `libstdc++.modules.json` file alongside `libstdc++.so`, which tells the build system where the files are. We can also find this file with :
```shell
gcc -print-file-name=libstdc++.modules.json
```

Content in the json may looks like this:
```json
{
  "version": 1,
  "revision": 1,
  "modules": [
    {
      "logical-name": "std",
      "source-path": "../include/c++/15/bits/std.cc",
      "is-std-library": true
    },
    {
      "logical-name": "std.compat",
      "source-path": "../include/c++/15/bits/std.compat.cc",
      "is-std-library": true
    }
  ]
}
```

The structure of this json file is basicly the same as the one provided by llvm.

## Known issue

Currently, these two files cannot compile with `-D_GLIBCXX_USE_CXX11_ABI=0` option, so I set ths macro to `1` if a target is compile with std modules. This issue may related to #2716 and #3855 .

### DIFF
--- a/xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
+++ b/xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
@@ -19,6 +19,7 @@
 --
 
 -- imports
+import("core.base.json")
 import("core.base.semver")
 import("core.project.config")
 import("lib.detect.find_tool")
@@ -58,7 +59,9 @@ function load(target)
     -- fix cxxabi issue
     -- @see https://github.com/xmake-io/xmake/issues/2716#issuecomment-1225057760
     -- https://github.com/xmake-io/xmake/issues/3855
-    if target:policy("build.c++.gcc.modules.cxx11abi") then
+
+    -- libstdc++ std modules cannot compile with -D_GLIBCXX_USE_CXX11_ABI=0
+    if target:policy("build.c++.modules.std") or target:policy("build.c++.gcc.modules.cxx11abi") then
         target:add("cxxflags", "-D_GLIBCXX_USE_CXX11_ABI=1")
     else
         target:add("cxxflags", "-D_GLIBCXX_USE_CXX11_ABI=0")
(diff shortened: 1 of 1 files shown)


xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
```diff
@@ -132,8 +135,50 @@ function get_target_module_mapperpath(target)
...
+    if not target:policy("build.c++.modules.std") then
+        return nil
+    end
+
+    local modules_json_path = _get_std_module_manifest_path(target)
+    if not modules_json_path then
+        return nil
+    end
+
+    local modules_json = json.decode(io.readfile(modules_json_path))
```
A2va, line 169: You can use `json.loadfile`

xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
```diff
@@ -132,8 +135,50 @@ function get_target_module_mapperpath(target)
...
--- not supported atm
+function _get_std_module_manifest_path(target)
+    local compinst = target:compiler("cxx")
+
+    local modules_json_path, _ = try {
+        function()
+            return os.iorunv(compinst:program(), {"-print-file-name=libstdc++.modules.json"}, {envs = compinst:runenvs()})
+        end
+    }
+    modules_json_path = modules_json_path:trim()
```
waruqi, line 146: modules_json_path maybe nil, 

```lua
if modules_json_path then
    modules_json_path = modules_json_path:trim()
    if os.isfile(modules_json_path) then
        return modules_json_path
    end
end
```

xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
```diff
@@ -132,8 +135,50 @@ function get_target_module_mapperpath(target)
...
+    local modules_json = json.decode(io.readfile(modules_json_path))
+    if modules_json and modules_json.modules and #modules_json.modules > 0 then
+        local std_module_files = {}
+        local modules_json_dir = path.directory(modules_json_path)
+        for _, module_file in ipairs(modules_json.modules) do
+            local module_file_path = module_file["source-path"]
+            if not path.is_absolute(module_file_path) then
+                module_file_path = path.join(modules_json_dir, module_file_path)
+            end
+            table.join2(std_module_files, {module_file_path})
```
waruqi, line 178:
```lua
            table.insert(std_module_files, module_file_path)
```

xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
```diff
@@ -255,3 +310,19 @@ function get_cppversionflag(target)
     return cppversionflag or nil
 end
 
+function get_gcc_version(target)
+    local gcc_version = _g.gcc_version
+    if not gcc_version then
+        local program, toolname = target:tool("cxx")
+        if program and toolname:startswith("gcc") then
+            local gcc = find_tool(toolname, {program = program, version = true,
+                envs = os.getenvs(), cachekey = "modules_support_gcc_" .. toolname})
```
waruqi, line 319: we need not pass os.getenvs

waruqi:
it will break ci. https://github.com/xmake-io/xmake/actions/runs/12023662331

https://github.com/xmake-io/xmake/pull/5886

xmake/rules/c++/modules/modules_support/gcc/compiler_support.lua
```diff
@@ -58,7 +59,9 @@ function load(target)
     -- fix cxxabi issue
     -- @see https://github.com/xmake-io/xmake/issues/2716#issuecomment-1225057760
     -- https://github.com/xmake-io/xmake/issues/3855
-    if target:policy("build.c++.gcc.modules.cxx11abi") then
+
+    -- libstdc++ std modules cannot compile with -D_GLIBCXX_USE_CXX11_ABI=0
+    if target:policy("build.c++.modules.std") or target:policy("build.c++.gcc.modules.cxx11abi") then
```
waruqi, line 64: it will break

PR #5895, opened 2024-11-27
fix(project): Allow extension with wildcard character

24bit-xjkp:
1. fix"unknown source file" error when using extension with wildcard character in functions like add_files


waruqi:
不考虑，当初就是为了考虑性能，才不这么搞的。

PR #5897, opened 2024-11-28
Fix module parallel compile error

vrecluse:
Fix compile error when parallel compile modules with clang.


waruqi:
commit 改成英文，重新 push force 下

PR #5918, opened 2024-12-02
Pass --sdk to dependencies

### LINKED ISSUE #5917: Dependencies should inherit --sdk

If a user wants to explicitly set the sdk, this value should be passed to dependencies as well otherwise they might be compiled with an incompatible config.
This is a problem in particular when using msvc-wine because the sdk must be set.

When `--sdk <path>` is passed to xmake config, it shouldn't be thrown away when dependencies are built.

_No response_

### Additional context

_No response_

UE4SS:
This fixes #5917.
This is my first pull request for this project, please let me know if there's anything missing or if I did something wrong.

### DIFF
--- a/xmake/modules/package/tools/xmake.lua
+++ b/xmake/modules/package/tools/xmake.lua
@@ -89,6 +89,13 @@ function _get_configs_for_windows(package, configs, opt)
     if package:config("toolchains") and _is_toolchain_compatible_with_host(package) then
         _get_configs_for_host_toolchain(package, configs, opt)
     end
+
+    if not is_host("windows") then
+        local sdkdir = _get_config_from_toolchains(package, "sdkdir") or get_config("sdk")
+        if sdkdir and #sdkdir > 0 then
+            table.insert(configs, "--sdk=" .. sdkdir)
+        end
+    end
 end
 
 -- get configs for appleos


waruqi:
We don't always pass the sdk, it's only required for cross-compilation and toolchains like mingw. what's your toolchain and platform?

https://github.com/xmake-io/xmake/blob/71a5624d37cdfaaba039a874e67341ee354357b7/xmake/modules/package/tools/xmake.lua#L202

https://github.com/xmake-io/xmake/blob/71a5624d37cdfaaba039a874e67341ee354357b7/xmake/modules/package/tools/xmake.lua#L176

UE4SS:
> We don't always pass the sdk, it's only required for cross-compilation and toolchains like mingw. what's your toolchain and platform?
> 
> https://github.com/xmake-io/xmake/blob/71a5624d37cdfaaba039a874e67341ee354357b7/xmake/modules/package/tools/xmake.lua#L202
> 
> https://github.com/xmake-io/xmake/blob/71a5624d37cdfaaba039a874e67341ee354357b7/xmake/modules/package/tools/xmake.lua#L176

I'm using msvc-wine and host is Arch Linux.

waruqi:
you should add it in _get_configs_for_windows()

```lua
function _get_configs_for_windows()
    if not is_host("windows") then
        -- pass sdk
    end
end
```

UE4SS:
> you should add it in _get_configs_for_windows()
> 
> ```lua
> function _get_configs_for_windows()
>     if not is_host("windows") then
>         -- pass sdk
>     end
> end
> ```

This makes sense, I've applied the changes.

xmake/modules/package/tools/xmake.lua
```diff
@@ -89,6 +89,13 @@ function _get_configs_for_windows(package, configs, opt)
     if package:config("toolchains") and _is_toolchain_compatible_with_host(package) then
         _get_configs_for_host_toolchain(package, configs, opt)
     end
+
+    if not is_host("windows") then
+        local sdk = get_config("sdk")
```
waruqi, line 94: local sdkdir = _get_config_from_toolchains(package, "sdkdir") or get_config("sdk")

PR #5919, opened 2024-12-02
Use ccache when debug cmake package

star-hengxing:


xmake/modules/private/action/require/impl/actions/install.lua
```diff
@@ -343,6 +345,28 @@ function _enter_package_testenvs(package)
...
+        return
+    end
+
+    if not project.policy("package.build.ccache") then
+        return
+    end
+
+    local ccache = find_tool("ccache")
+    if not ccache then
+        ccache = find_tool("sccache")
```
waruqi, line 359: 可以改进这里，走全局获取 https://github.com/xmake-io/xmake/blob/dev/xmake/modules/private/tools/ccache.lua

PR #6015, opened 2024-12-28
zig cc: fix wrapper not working properly in Git Bash

Doekin:
The `zig cc` toolchain encounters issues when installing packages via `xrepo` in `Git Bash`:
1. Attempting to run cmd scripts as sh scripts.
2. Incorrectly appending the `.exe` extension to cmd scripts.


Doekin:
Thank you for highlighting this issue. It occurs when paths with spaces are not quoted, which is also a problem in the original implementation. Perhaps we should check and ensure `binpath` is a valid shell command beforehand.

PR #6023, opened 2024-12-30
[feature] [project] [cmake]  Forward compatibility way, can specify file name v…

19317362:
I'm trying to use Android Studio to debug native code using CMake bridge via xmake.  The command `xmake project -k cmake` will overwrite CMakeLists.txt.  

After adding this feature, I can use xmake to specify a filename like CMakeLists_arm64-v8a.txt, and then, in the CMakeLists.txt, to include CMakeLists_arm64-v8a.txt.

By this way, I can use AS to debug native code step by step. All code managed by xmake.lua.

I think this feature is usable for others.

### DIFF
--- a/xmake/plugins/project/xmake.lua
+++ b/xmake/plugins/project/xmake.lua
@@ -66,6 +66,7 @@ task("project")
                                                         ,   "    - xmake project -k compile_commands --lsp=clangd"
                                                         ,   values = {"clangd", "cpptools", "ccls"}}
                 ,   {nil, "outputdir", "v"  , "."       ,   "Set the output directory."                                                     }
+                ,   {nil, "outputfile", "v"  , nil       ,  "Set the output file (cmake only) ."                                                     }
                 }
             }
 
--- a/xmake/plugins/project/cmake/cmakelists.lua
+++ b/xmake/plugins/project/cmake/cmakelists.lua
@@ -19,6 +19,7 @@
 --
 
 -- imports
+import("core.base.option")
 import("core.base.colors")
 import("core.project.project")
 import("core.project.config")
@@ -1337,8 +1338,10 @@ function make(outputdir)
     -- enter project directory
     local oldir = os.cd(os.projectdir())
 
+    -- determine the filename
+    local filename = option.get("outputfile") or "CMakeLists.txt"
     -- open the cmakelists
-    local cmakelists = io.open(path.join(outputdir, "CMakeLists.txt"), "w")
+    local cmakelists = io.open(path.join(outputdir, filename), "w")
 
     -- generate cmakelists
     _generate_cmakelists(cmakelists, outputdir)


xmake/plugins/project/xmake.lua
```diff
@@ -66,6 +66,7 @@ task("project")
                                                         ,   "    - xmake project -k compile_commands --lsp=clangd"
                                                         ,   values = {"clangd", "cpptools", "ccls"}}
                 ,   {nil, "outputdir", "v"  , "."       ,   "Set the output directory."                                                     }
+                ,   {nil, "outputfile", "v"  , nil       ,  "Set the output file (cmake only) ."                                                     }
```
waruqi, line 69: If this parameter is added globally, then other generators should be required to support it.

waruqi:
> After adding this feature, I can use xmake to specify a filename like CMakeLists_arm64-v8a.txt, and then, in the CMakeLists.txt, to include CMakeLists_arm64-v8a.txt.

and why not use outputdir? you can generate it to `xmake/CMakeLists.txt` then include it in root CMakeLists.txt

xmake/plugins/project/xmake.lua
19317362: Should I make another PR for all generators?

waruqi: right

19317362:
> > After adding this feature, I can use xmake to specify a filename like CMakeLists_arm64-v8a.txt, and then, in the CMakeLists.txt, to include CMakeLists_arm64-v8a.txt.
> 
> and why do not use outputdir? you can generate it to `xmake/CMakeLists.txt` then include it in root CMakeLists.txt

I've tested, outputdir param can be used. CMake includes treats relative paths of .cpp/.h files correctly.
Before this test, I'm afraid of paths to .cpp/.h files in other dirs.

19317362:
> > After adding this feature, I can use xmake to specify a filename like CMakeLists_arm64-v8a.txt, and then, in the CMakeLists.txt, to include CMakeLists_arm64-v8a.txt.
> 
> and why do not use outputdir? you can generate it to `xmake/CMakeLists.txt` then include it in root CMakeLists.txt

I'm trying to migerate my old project (Makefile/Android.mk based) to xmake, and make it can be debugged with AS.

When using xmake-gradle-plugin and AS to debug, it has some problem with source code level debugging.


When use outputdir, after generate CMakeLists.txt  in sub-dir, it can not find source files correctly.

xmake.lua   --- to generate CMakeLists.txt for AS to debug

CMakeLists.txt    -- for build.gradle
   --- includes  build/android/arm64-v8a/CMakeLists.txt

-- build/android/arm64-v8a/CMakeLists.txt   -- generated with xmake project -k cmake
-- jni_stub/nativelib.cpp
-- dir1/*.cpp
-- dir2/*.cpp


When use --outpufile to make it like this, it works fine.

xmake.lua
CMakeLists.txt
CMakeLists_arm64-v8a.txt
-- jni_stub/nateivelib.cpp
-- dir1/*.cpp
-- dir2/*.cpp

waruqi:
why not rewrite root CMakelist.txt directly?

19317362:
> why not rewrite root CMakelist.txt directly?

The native.cpp is just a bridge between JAVA and C++.so
1. the old project now works only arm32, and will be migrated to support arm64. so, both ABI was needed.
    the CMakeLists.txt generated by xmake is only one ABI
2. before CMakeLists.txt was generated,  the operation of "Sync with Gradle file" will fail.
     I want to make all works (e.g. sync, generate, build, clean, debug) with AS

If you do not think this is useful to the main baseline, I'll keep it in my own version. 
Anyway, I've made it works with AS + Gradle + xmake + cmake, editing, debugging step by step. 

Thanks for you great works. Simplified the develop process of our OLD embedded Linux + Android project.

Issue #6010, opened 2024-12-26
Re-add actions to build xmake with cosmocc


A2va:
### Is your feature request related to a problem? Please describe.

Previously, a workflow was created to build xmake with cosmocc
https://github.com/xmake-io/xmake/blob/b657911e2b0c34674570368c73c5025eff9d7649/.github/workflows/cosmocc.yml
However, this was not successful as it didn't run, but I noticed that this build was made using macos, so I tried to build it on linux.

First, I just tested that a normal xmake build (without embed) would work
```
a2va:~/xmake/core$ xmake f -p linux --cosmocc=y
```

The executable runs fine, then I enabled embed
```
a2va:~/xmake/core$ xmake f -p linux --cosmocc=y
```

Again, no problem with the executable, then I copied the file to Windows (the build was done in WSL Ubuntu). And it starts, but with this error:

```
PS D:\\xm-tt> .\\xmake-tt.exe --version
Error: @programdir/core/base/task.lua:398: @programdir/core/base/winos.lua:238: Field '_registry_query' is not accessible (a null value)
Stack traceback:
    [@programdir/core/base/winos.lua:238]: in function 'registry_query'.
    [@programdir/core/base/cpu.lua:222]: in function '_info'.
    [@programdir/core/base/cpu.lua:286]: in function 'Vendor'
    [@programdir/core/base/cpu.lua:352]:
    [@programdir/core/base/os.lua:1443]: in function 'default_njob'.
    [@programdir/plugins/pack/xmake.lua:31]: in main chunk

```

I suspect this is because Cosmopolitan doesn't implement the Windows API for editing the registry, but those can added if needed (eg: https://github.com/jart/cosmopolitan/pull/1318), or the other solution would be to completely disable this functionality.

### Describe the solution you'd like

With this feature the xmake bundle executable which is for now windows only could be extended to other platforms.

waruqi:
I tried it before, but it still has many problems, such as it will stuck on macos ci and some windows api problems.

But I don't have time to solve them recently, so I can only delete it temporarily. If you can solve these problems, you can submit a PR.

A2va:
I managed to built xmake in the CI but one of the test fails, I don't know if this is due to cmake or not.
```
[...]

PR #6083, opened 2025-01-15
Re-add cosmocc actions

A2va:
I successfully built xmake with cosmocc, it runs on linux, so it should work on macos. On Windows there is still the bug in the first comment of issue #6010. 

This binary should also work on ARM64 but I cannot test it myself, so if anyone wants to try it out.


A2va:
What I meant by "it should work on macos" is that the same binary compiled in linux would also work

waruqi:
> What I meant by "it should work on macos" is that the same binary compiled in linux would also work on macos

I know, but we also need to test it to confirm it actually works and isn't just a guess. I ran it on macos ci before and it would get stuck.

waruqi:
> I don't see to advantages to test the compilation with cosmocc on other platforms

Not for testing compilation, but for testing running, we can only recompile it on macos. Unless we have a way to pass the cosmocc binary built on linux to macos to run.

A2va:
I don't know why the comoscc macos actions was cancelled but the same actions on my repo worked (the actions name is different)
https://github.com/A2va/xmake/actions/runs/12796569514

.github/workflows/cosmocc_macos.yml
```diff
@@ -0,0 +1,66 @@
...
+  build:
+    strategy:
+      matrix:
+        os: [macos-13]
+        arch: [x86_64]
+
+    runs-on: ${{ matrix.os }}
+
+    concurrency:
+        group: ${{ github.ref }}-${{ github.base_ref }}-${{ github.head_ref }}-Cosmocc-${{ matrix.arch }}
```
waruqi, line 18: we need to set different key for linux/macos.

PR #6097, opened 2025-01-19
Add `qt_host` Option to Enable Cross-Platform Qt Builds Using Host SDK Tools

Doekin:
This update introduces an option, `qt_host`, to facilitate cross-building of Qt desktop applications.

#### Example Usage

1. **Build a Qt Widgets app for MinGW on Linux**  
   Using an SDK downloaded by `aqt`:
   ```bash
   xmake f -c -p mingw --qt=../sdk/6.9.0/llvm-mingw_64 --qt_host=../sdk/6.9.0/gcc_64
   ```

2. **Build a Qt Widgets app for Linux on Windows**  
   Using an SDK downloaded by `aqt` and the Zig toolchain:
   ```bash
   xmake f -c -p linux --toolchain=zig --qt=../sdk/6.9.0/gcc_64 --qt_host=../sdk/6.9.0/llvm-mingw_64
   ```

#### Caution

- Ensure the host Qt version matches the target version.
- The rules `qt.deploy` and `qt.install` might cause cross-platform issues, as `windeployqt` and `macdeployqt` need to run on their native platforms.

#### Tested Configurations

- **Cross-Builds:**
  - Qt 6.9.0 for Linux built on Windows with the Zig toolchain
  - Qt 6.9.0 for MinGW built on Linux
  - Qt 6.8.0 for Android built on Windows 

- **Native Builds (Regression Tests):**
  - Qt 6.8.1 for Linux (system-installed)
  - Qt 5.15.16 for Linux (system-installed)
  - Qt 6.9.0 for Linux (downloaded via `aqt`)
  - Qt 6.9.0 for MinGW

### DIFF
--- a/xmake/modules/detect/sdks/find_qt.lua
+++ b/xmake/modules/detect/sdks/find_qt.lua
@@ -190,11 +190,21 @@ function _find_qmake(sdkdir, sdkver)
 end
 
 -- get qt environment
-function _get_qtenvs(qmake)
+function _get_qtenvs(qmake, sdkdir)
     local envs = {}
+    local run_args = {"-query"}
+    if sdkdir then
+        local conf_paths = {path.join(sdkdir, "bin", "target_qt.conf"), path.join(sdkdir, "bin", "qt.conf")}
+        for _, conf_path in ipairs(conf_paths) do
+            if os.isfile(conf_path) then
+                table.join2(run_args, {"-qtconf", conf_path})
+                break
+            end
+        end
+    end
     local results = try {
         function ()
-            return os.iorunv(qmake, {"-query"})
+            return os.iorunv(qmake, run_args)
         end,
         catch {
             function (errors)
@@ -215,24 +225,44 @@ function _get_qtenvs(qmake)
     end
 end
 
+-- Verify and correct the Qt SDK version for cross-compiling.
+-- qmake reports its own version (QT_VERSION), not the version specified in the SDK's configuration files.
+function _tryfix_sdkver_for_cross(sdkdir, sdkver)
+    local qconfig_path = sdkdir and path.join(sdkdir, "mkspecs", "qconfig.pri")
+    if not sdkver or not os.isfile(qconfig_path) then
+        return sdkver
+    end
+    -- Extract the actual SDK version from qconfig.pri
+    local qconfig = io.readfile(qconfig_path)
+    local actual_sdkver = qconfig and qconfig:match("QT_VERSION%s*=%s*(%S+)") -- Expected format: QT_VERSION = x.y.z
+    if not actual_sdkver then
+        return sdkver
+    end
+    if sdkver ~= actual_sdkver then
+        wprint("Host Qt SDK version (%s) differs from Target Qt SDK version (%s). To prevent build issues, please ensure both use the same version.", sdkver, actual_sdkver);
+    end
+    return actual_sdkver
+end
+
 -- find qt sdk toolchains
-function _find_qt(sdkdir, sdkver)
+function _find_qt(sdkdir, sdkver, sdkdir_host)
 
     -- find qmake
-    local qmake = _find_qmake(sdkdir, sdkver)
+    local qmake = _find_qmake(sdkdir_host or sdkdir, sdkver)
     if not qmake then
         return
     end
 
     -- get qt environments
-    local qtenvs = _get_qtenvs(qmake)
+    local located_sdkdir = sdkdir and _find_sdkdir(sdkdir, sdkver)
+    local qtenvs = _get_qtenvs(qmake, located_sdkdir or sdkdir)
     if not qtenvs then
         return
     end
 
     -- get qt toolchains
     sdkdir = qtenvs.QT_INSTALL_PREFIX
-    local sdkver = qtenvs.QT_VERSION
+    local sdkver = _tryfix_sdkver_for_cross(sdkdir, qtenvs.QT_VERSION)
     local bindir = qtenvs.QT_INSTALL_BINS
     local libexecdir = qtenvs.QT_INSTALL_LIBEXECS
     local qmldir = qtenvs.QT_INSTALL_QML
@@ -259,6 +289,16 @@ function _find_qt(sdkdir, sdkver)
             -- TODO
         end
     end
+
+    if sdkdir_host then
+        local located_sdkdir_host = _find_sdkdir(sdkdir_host, sdkver)
+        local qtenvs_host = _get_qtenvs(qmake, located_sdkdir_host or sdkdir_host)
+        if qtenvs_host then
+            bindir_host = qtenvs_host.QT_HOST_BINS or qtenvs_host.QT_INSTALL_BINS or bindir_host
+            libexecdir_host = qtenvs_host.QT_HOST_LIBEXECS or qtenvs_host.QT_INSTALL_LIBEXECS or libexecdir_host
+        end
+    end
+
     return {sdkdir = sdkdir, bindir = bindir, bindir_host = bindir_host, libexecdir = libexecdir, libexecdir_host = libexecdir_host, libdir = libdir, includedir = includedir, qmldir = qmldir, pluginsdir = pluginsdir, mkspecsdir = mkspecsdir, sdkver = sdkver}
 end
 
@@ -282,13 +322,16 @@ function main(sdkdir, opt)
 
     -- attempt to load cache first
     local key = "detect.sdks.find_qt"
-    local cacheinfo = detectcache:get(key) or {}
+    local cacheinfo = (sdkdir and detectcache:get2(key, sdkdir)) or detectcache:get(key) or {}
     if not opt.force and cacheinfo.qt and cacheinfo.qt.sdkdir and os.isdir(cacheinfo.qt.sdkdir) then
         return cacheinfo.qt
     end
 
     -- find qt
-    local qt = _find_qt(sdkdir or config.get("qt") or global.get("qt") or config.get("sdk"), opt.version or config.get("qt_sdkver"))
+    local sdkdir = sdkdir or config.get("qt") or global.get("qt") or config.get("sdk")
+    local sdkver = opt.version or config.get("qt_sdkver")
+    local sdkdir_host = opt.sdkdir_host or config.get("qt_host") or global.get("qt_host")
+    local qt = _find_qt(sdkdir, sdkver, sdkdir_host)
     if qt then
 
         -- save to config
@@ -314,7 +357,11 @@ function main(sdkdir, opt)
 
     -- save to cache
     cacheinfo.qt = qt or false
-    detectcache:set(key, cacheinfo)
+    if sdkdir then
+        detectcache:set2(key, sdkdir, cacheinfo)
+    else
+        detectcache:set(key, cacheinfo)
+    end
     detectcache:save()
     return qt
 end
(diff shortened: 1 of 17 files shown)


Doekin:
The `-qtconf` argument is valid for both `qmake5` and `qmake6`.

<details>
<summary> Help message for qmake-qt5 </summary>

```
$ qmake-qt5 -v
QMake version 3.1
Using Qt version 5.15.16 in /usr/lib

$ qmake-qt5 -help
Usage: qmake-qt5 [mode] [options] [files]

QMake has two modes, one mode for generating project files based on
some heuristics, and the other for generating makefiles. Normally you
shouldn't need to specify a mode, as makefile generation is the default
mode for qmake, but you may use this to test qmake on an existing project

Mode:
  -project       Put qmake into project file generation mode
                 In this mode qmake interprets [files] as files to
                 be added to the .pro file. By default, all files with
                 known source extensions are added.
                 Note: The created .pro file probably will 
                 need to be edited. For example add the QT variable to 
                 specify what modules are required.
  -makefile      Put qmake into makefile generation mode (default)
                 In this mode qmake interprets files as project files to
                 be processed, if skipped qmake will try to find a project
                 file in your current working directory

Warnings Options:
  -Wnone         Turn off all warnings; specific ones may be re-enabled by
                 later -W options
  -Wall          Turn on all warnings
  -Wparser       Turn on parser warnings
  -Wlogic        Turn on logic warnings (on by default)
  -Wdeprecated   Turn on deprecation warnings (on by default)

Options:
   * You can place any variable assignment in options and it will be *
   * processed as if it was in [files]. These assignments will be    *
   * processed before [files] by default.                            *
  -o file        Write output to file
  -d             Increase debug level
  -t templ       Overrides TEMPLATE as templ
  -tp prefix     Overrides TEMPLATE so that prefix is prefixed into the value
  -help          This help
  -v             Version information
  -early         All subsequent variable assignments will be
                 parsed right before default_pre.prf
  -before        All subsequent variable assignments will be
                 parsed right before [files] (the default)
  -after         All subsequent variable assignments will be
                 parsed after [files]
  -late          All subsequent variable assignments will be
                 parsed right after default_post.prf
  -norecursive   Don't do a recursive search
  -recursive     Do a recursive search
  -set <prop> <value> Set persistent property
  -unset <prop>  Unset persistent property
  -query <prop>  Query persistent property. Show all if <prop> is empty.
  -qtconf file   Use file instead of looking for qt.conf
  -cache file    Use file as cache           [makefile mode only]
  -spec spec     Use spec as QMAKESPEC       [makefile mode only]
  -nocache       Don't use a cache file      [makefile mode only]
  -nodepend      Don't generate dependencies [makefile mode only]
  -nomoc         Don't generate moc targets  [makefile mode only]
  -nopwd         Don't look for files in pwd [project mode only]
```

</details>

Doekin:
It's important to note that `qmake` reports its own version instead of the SDK version specified by `-qtconf`. For example, when running:

```bash
qmake -query -qtconf ../sdk/6.9.0/llvm-mingw_64/bin/qt.conf
```

The output shows:

```
QT_SYSROOT:
QT_INSTALL_PREFIX:/run/media/leemu/D8C48579C4855B20/projects/qt/sdk/6.9.0/llvm-mingw_64
...
QT_VERSION:5.15.16
```

Notice that `QT_VERSION` is 5.15.16, not 6.9.0 as expected. This discrepancy may lead to XMake incorrectly identifying the Qt SDK version:

```bash
$ xmake f -c -vD --qt=../sdk/6.9.0/llvm-mingw_64 --qt_host=../sdk/6.8.1/gcc_64 -p mingw
```

The output shows:

```
checking for Qt SDK version ... 6.8.1
```

The Qt SDK version should be 6.9.0

Doekin:
> This discrepancy may lead to XMake incorrectly identifying the Qt SDK version

It is fixed now.

waruqi:
Why is an additional qt_host needed? I remember that cross-compilation was also supported before.

It can do cross-compilation for ios/android.

https://github.com/xmake-io/xmake/blob/1ea11f3c8a23bb86f2015cbba8c41d19c1e471d4/xmake/modules/detect/sdks/find_qt.lua#L244-L262

and did you test qt5base/qt6base packages? it also supports cross-compilation without extra qt_host sdk.


https://github.com/xmake-io/xmake-repo/blob/a266c2344972fd8d7cce4301b1a02500db9afa1c/packages/q/qtbase/xmake.lua#L153

Doekin:
The current setup doesn't support building Qt applications for Linux on Windows, or vice versa. 

Regarding the `qt6base` package, I've encountered issues where it fails to function properly, with errors like:

```
checking for qt6widgets ... no
note: the following packages are unsupported on linux/x64:
  -> qt6base 6.8.0 [from:qt6core,qt6gui,qt6widgets, license:LGPL-3]
  -> qt6core 6.8.0 [toolchains:"zig", from:qt6gui,qt6widgets, license:LGPL-3]
  -> qt6gui 6.8.0 [toolchains:"zig", from:qt6widgets, license:LGPL-3]
  -> qt6widgets 6.8.0 [toolchains:"zig", license:LGPL-3]
error: @programdir\core\main.lua:329: stack traceback:
    [C]: in function 'error'
    [@programdir\core\base\os.lua:1086]:
    [...modules\private\action\require\impl\install_packages.lua:790]: in function '_install_packages'
    [...modules\private\action\require\impl\install_packages.lua:852]:
    [@programdir\modules\private\action\require\install.lua:85]:
    [@programdir\actions\config\main.lua:391]:
    [C]: in function 'xpcall'
    [@programdir\core\base\utils.lua:244]:
    [@programdir\core\base\task.lua:491]: in function 'run'
    [@programdir\core\main.lua:327]: in function 'cotask'
    [@programdir\core\base\scheduler.lua:406]:
```

Doekin:
Marked this PR as draft for further refinement

Doekin:
### Additional Notes:

The Qt SDK consists of two main components: libraries and tools (e.g., `qmake`, `moc`, `uic`, etc.), where libraries are specific to the target platform.

Since Qt SDKs are not typically configured for cross-platform use, cross-compilation requires two separate SDKs:

1. **Target-Dependent SDK**: Includes the libraries the application links against, built for the target platform. It may also contain Qt tools that are incompatible with the build machine.  
2. **Host-Dependent SDK**: Provides the tools needed for the build process, compatible with the build machine.

xmake/rules/qt/moc/xmake.lua
```diff
@@ -27,12 +27,15 @@ rule("qt.moc")
...
         local qt = assert(target:data("qt"), "Qt not found!")
-        local moc = path.join(qt.bindir, is_host("windows") and "moc.exe" or "moc")
-        if not os.isexec(moc) and qt.libexecdir then
-            moc = path.join(qt.libexecdir, is_host("windows") and "moc.exe" or "moc")
-        end
-        if not os.isexec(moc) and qt.libexecdir_host then
-            moc = path.join(qt.libexecdir_host, is_host("windows") and "moc.exe" or "moc")
+        local moc
+        local moc_name = is_host("windows") and "moc.exe" or "moc"
+        for _, dir in ipairs({qt.bindir_host, qt.libexecdir_host, qt.bindir, qt.libexecdir}) do
```
waruqi, line 32: we can use `lib.detect.find_file`

xmake/rules/qt/qmltyperegistrar/xmake.lua
```diff
@@ -27,12 +27,15 @@ rule("qt.qmltyperegistrar")
...
-            qmltyperegistrar = path.join(qt.libexecdir, is_host("windows") and "qmltyperegistrar.exe" or "qmltyperegistrar")
-        end
-        if not os.isexec(qmltyperegistrar) and qt.libexecdir_host then
-            qmltyperegistrar = path.join(qt.libexecdir_host, is_host("windows") and "qmltyperegistrar.exe" or "qmltyperegistrar")
+        local qmltyperegistrar
+        local qmltyperegistrar_name = is_host("windows") and "qmltyperegistrar.exe" or "qmltyperegistrar"
+        for _, dir in ipairs({qt.bindir_host, qt.libexecdir_host, qt.bindir, qt.libexecdir}) do
+            if dir then
+                qmltyperegistrar = path.join(dir, qmltyperegistrar_name)
+                if os.isexec(qmltyperegistrar) then
```
waruqi, line 35: and here

xmake/rules/qt/ts/xmake.lua
```diff
@@ -20,22 +20,31 @@ rule("qt.ts")
...
-            end
-            if not os.isexec(lrelease) and qt.libexecdir_host then
-                lrelease = path.join(qt.libexecdir_host, is_host("windows") and "lrelease.exe" or "lrelease")
+
+            local lupdate
+            local lupdate_name = is_host("windows") and "lupdate.exe" or "lupdate"
+            for _, dir in ipairs({qt.bindir_host, qt.libexecdir_host, qt.bindir, qt.libexecdir}) do
+                if dir then
+                    lupdate = path.join(dir, lupdate_name)
+                    if os.isexec(lupdate) then
```
waruqi, line 29: and here

Doekin:
qt5base/qt6base does not support cross-compilation well. Cross-compiling for desktop platforms is not supported, and Android targets are not enabled. 

https://github.com/xmake-io/xmake-repo/blob/4fd9216b6d1bd108f9211849a4bfd3b1b6f96b3c/packages/q/qt6base/xmake.lua#L25-L27

xmake/modules/detect/sdks/find_qt.lua
```diff
@@ -215,24 +225,43 @@ function _get_qtenvs(qmake)
...
 
+-- Verify and correct the Qt SDK version for cross-compiling.
+-- qmake reports its own version (QT_VERSION), not the version specified in the SDK's configuration files.
+function _tryfix_sdkver_for_cross(sdkdir, sdkver)
+    local qconfig_path = sdkdir and path.join(sdkdir, "mkspecs", "qconfig.pri")
+    if not sdkver or not os.isfile(qconfig_path) then
+        return sdkver
+    end
+    -- Extract the actual SDK version from qconfig.pri
+    local actual_sdkver = io.readfile(qconfig_path):match("QT_VERSION%s*=%s*(%S+)") -- Expected format: QT_VERSION = x.y.z
```
waruqi, line 236: readfile maybe return nil

xmake/rules/qt/deploy/android.lua
```diff
@@ -71,10 +71,10 @@ function main(target, opt)
     end
 
     -- get androiddeployqt
-    local androiddeployqt = path.join(qt.bindir, "androiddeployqt" .. (is_host("windows") and ".exe" or ""))
-    if not os.isexec(androiddeployqt) and qt.bindir_host then
-        androiddeployqt = path.join(qt.bindir_host, "androiddeployqt" .. (is_host("windows") and ".exe" or ""))
-    end
+    local search_dirs = {}
+    if qt.bindir_host       then table.insert(search_dirs, qt.bindir_host       ) end
```
waruqi, line 75: remove spaces, we don't need alignment

waruqi:
Thanks, can you help to update docs here https://xmake.io/#/guide/project_examples?id=supported-qt-sdks

Doekin:
OK, I'd be happy to help

star-hengxing:
https://github.com/xmake-io/xmake/discussions/6133 Old Qt (qmake) doesn't support `-conf`. I think we need to provide a fallback to execute `qmake -query` again.

PR #6124, opened 2025-01-27
Clearer extract errors

Shiffted:
Forwards and rewords errors fom 'archive.extract' to make it clearer what exactly the error is.


waruqi:
please rebase it

PR #6142, opened 2025-02-08
Improve targetname input checks

Shiffted:
Applies the help/error message for invalid targets from action run to more actions/plugins, makes it generally available and improves on it.
I replaced existing checks and added checks to actions that would try to access a nil target when given an invalid input targetname.
The levenshtein function now also takes options for costs which the `find_similar_targetnames` function makes use of.

Currently this is only applied to actions/plugins but could also be used for e.g. xpack.
I'm unsure where to put the new files or if the current location is appropriate.


xmake/modules/lib/detect/find_similar_targetnames.lua
```diff
@@ -0,0 +1,71 @@
...
+    local targetname_lower = targetname:lower()
+    local matching_targetnames = {}
+    local matching_levenshtein = {}
+
+    for _, target in ipairs(project.ordertargets()) do
+        local name = target:name()
+        if name:lower():find(targetname_lower, 1, true) then
+            table.insert(matching_targetnames, name)
+        else
+            local distance = targetname:levenshtein(name, {sub = 2, ins = 1, del = 1})
```
waruqi, line 45: > {sub = 2, ins = 1, del = 1}

Why not use the default value?

xmake/actions/run/main.lua
```diff
@@ -29,6 +29,7 @@ import("devel.debugger")
 import("async.runjobs")
 import("private.action.run.runenvs")
 import("private.service.remote_build.action", {alias = "remote_build_action"})
+import("lib.detect.check_targetname")
```
waruqi, line 32: it's private interface, please move to `private.xxx` modules.

waruqi: I am not sure how many people will use this interface in the plugin. 

At present, it is only used internally. If it is to be opened as a public interface, it must at least be mature and stable. 

As it is a newly added interface, it should be temporarily placed in the private interface, and iterated for several versions until it does not change, and then we will gradually open it.

In addition, find_similar_targetnames, as its internal implementation interface, should be moved to the private interface first.

we can move them to `private.detect.xxx` fisrt. It will not be made public until its interface name and parameters are stable after several version iterations.

PR #6149, opened 2025-02-11
feat: improve GN tool to support extra arguments

poirierlouis:
`package.tools.gn`

Allows to pass extra arguments to ninja command. It defaults to an empty table (no arguments).

Example of usage:
```lua
local gn = import("package.tools.gn")

gn.build(package, configs, {buildir = "out.gn", argv = {"v8_monolith"}})
--                                              ^
--                                              | optional key of strings
```

This is a solution to discussion: https://github.com/xmake-io/xmake/discussions/6148#discussion-7947552
I've been able to test with my local installation of xmake. fyi, it allows to reduce time to build V8 in half.

Let me know if something is wrong or need changes.

### DIFF
--- a/xmake/modules/package/tools/gn.lua
+++ b/xmake/modules/package/tools/gn.lua
@@ -124,8 +124,8 @@ function build(package, configs, opt)
 
     -- do build
     local buildir = _get_buildir(opt)
-    local argv = opt.argv or {}
-    ninja.build(package, argv, {buildir = buildir, envs = opt.envs or buildenvs(package, opt)})
+    local targets = table.wrap(opt.target)
+    ninja.build(package, targets, {buildir = buildir, envs = opt.envs or buildenvs(package, opt)})
 end
 
 -- install package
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/gn.lua
```diff
@@ -124,7 +124,8 @@ function build(package, configs, opt)
 
     -- do build
     local buildir = _get_buildir(opt)
-    ninja.build(package, {}, {buildir = buildir, envs = opt.envs or buildenvs(package, opt)})
+    local argv = opt.argv or {}
+    ninja.build(package, argv, {buildir = buildir, envs = opt.envs or buildenvs(package, opt)})
```
waruqi, line 128: we shoud use target.

see 

https://github.com/xmake-io/xmake/blob/b509baeaf00de64d57f0fa1171e4c3eaf94e0dd2/xmake/modules/package/tools/cmake.lua#L1048-L1051

```lua
gn.build(package, configs, {target = "v8_xxx"})
```

```
usage: ninja [options] [targets...]
```

waruqi: and we need also to modify gn.install

poirierlouis: Oh right. I pushed changes.
I'm not sure regarding usage of `table.wrap`. Is it a `xmake` implementation only?

waruqi: > Oh right. I pushed changes. I'm not sure regarding usage of `table.wrap`. Is it a `xmake` implementation only?

right, only for xmake

PR #6172, opened 2025-02-26
Pass more options from `package:find_tool` to `lib.detect.find_tool`

Doekin:
This update ensures that ~~all options~~ options `check`, `command` and `parse` are passed from `package:find_tool` to `lib.detect.find_tool`, allowing for customizable check methods in packages.

### DIFF
--- a/xmake/core/package/package.lua
+++ b/xmake/core/package/package.lua
@@ -1966,6 +1966,9 @@ function _instance:find_tool(name, opt)
                                   bindirs = self:get("bindirs"),
                                   version = true, -- we alway check version
                                   require_version = opt.require_version,
+                                  check = opt.check,
+                                  command = opt.command,
+                                  parse = opt.parse,
                                   norun = opt.norun,
                                   system = opt.system,
                                   force = opt.force})


xmake/core/package/package.lua
```diff
@@ -1960,15 +1960,12 @@ end
...
+    opt.version = true -- we alway check version
     self._find_tool = self._find_tool or sandbox_module.import("lib.detect.find_tool", {anonymous = true})
-    return self._find_tool(name, {cachekey = opt.cachekey or "fetch_package_system",
-                                  installdir = self:installdir({readonly = true}),
-                                  bindirs = self:get("bindirs"),
-                                  version = true, -- we alway check version
-                                  require_version = opt.require_version,
-                                  norun = opt.norun,
-                                  system = opt.system,
-                                  force = opt.force})
```
waruqi, line 1971: ? why remove these args?

> 
                                  require_version = opt.require_version,
                                  norun = opt.norun,
                                  system = opt.system,
                                  force = opt.force

Doekin: I thought those arguments were already included in the `opt` table and passed directly to `self._find_tool`. Since `opt` is being passed as an entire table, I assumed we didn't need to explicitly pass each of these individual arguments again.

waruqi: We should not use opt parameter directly, it will modify the state of opt object outside the function.

waruqi: Also, not all opt parameters should be passed in. They are not all the same. Some parameters are only used by packages and should not be passed to lib.detect.find_tool

PR #6189, opened 2025-03-04
add xxx_scan rule/target build stage

Arthapz:
Add a new rule "stage" for scanning and generation, this would improve C++ module scanning support (and maybe other rules based on parsing file)


waruqi:
If we want to add new APIs, we should submit a feature request first, and then consider whether it is necessary to implement it after careful design and evaluation. Instead of opening a PR to add it directly.

This PR involves too many new APIs and actions. I haven't had time to carefully evaluate whether they are really needed.

In addition, I plan to refactor some rule APIs later, and it is not recommended to add new APIs to the rule for the time being.

PR #6222, opened 2025-03-14
meson: Add cmake search env

SirLynix:
Meson has an option to use cmake to find dependencies (harfbuzz uses it for example)


SirLynix:
see https://github.com/xmake-io/xmake-repo/pull/6590

PR #6249, opened 2025-03-24
(C++ modules) disable std module on some tests

Arthapz:
disable std module on tests that doesn't use it to improve tests speed


waruqi:
we should modify test_base.lua to disable them. `xmake f --policies=build.c++.modules.std:n`

These tests should be kept as simple as possible to help new users get started quickly. They can also be used as template projects.

Issue #6253, opened 2025-03-26
xmake install does not copy SO of cmake-package like freetype when cross-build


19317362:
### Xmake Version

v2.9.8+HEAD.13fc392

### Operating System Version and Architecture

Ubuntu 24.04

### Describe Bug

After `rm -rf build .xmake && xmake f -y && xmake && xmake install -o build/linux/x86_64/release/ -av`, xmake will install libfreetype.so* as expected.

But when build with arm-cross, it install nothing.
```
rm -rf build/linux/armv7/release
xmake f -cy -k shared -p linux -a armv7 --runtimes=stdc++_shared --ccache=n --shflags=-Wl,--build-id=sha1 --ldflags=-Wl,--build-id=sha1 -m release --YW_PLAT_ID=NT98528 --sdk=/opt/buildtools/arm32/ -v

xmake b -y -v && xmake i -v --installdir=build/linux/armv7/release
```

### Expected Behavior

When build and install with cross, copy SO as expected.

### Project Configuration

```lua
    add_requireconfs("*", {
        configs = {
            shared = true
        }
    }) 

add_requires("freetype")
target("test-freetype")
    set_kind("binary")
    add_files("src/*.cpp")
    -- add_packages("freetype")
```

src/main.cpp
```C++
#include <iostream>
#include <freetype/freetype.h>
#include <cstdint>

int main(int argc, char** argv) {

    printf("Hello, %s %d\n",argv[0],argc);

    FT_Library library;
    FT_Face face;

    // Initialize FreeType library
    if (FT_Init_FreeType(&library)) {
        std::cerr << "Could not initialize FreeType library" << std::endl;
        return 1;
    }

    // Load a font (replace "path/to/font.ttf" with an actual font file path)
    const char* fontPath = "path/to/font.ttf";
    if (FT_New_Face(library, fontPath, 0, &face)) {
        std::cerr << "Could not load font: " << fontPath << std::endl;
        FT_Done_FreeType(library);
        return 1;
    }

    // Output some basic font information
    std::cout << "Font loaded successfully!" << std::endl;
    std::cout << "Number of glyphs: " << face->num_glyphs << std::endl;
    std::cout << "Family name: " << face->family_name << std::endl;
    std::cout << "Style name: " << face->style_name << std::endl;

    // Clean up
    FT_Done_Face(face);
    FT_Done_FreeType(library);

    return 0;
}


```

### Additional Information and Error Logs

```bash
tree build/
build/
└── linux
[...]

PR #6257, opened 2025-03-27
[fix] xmake install skips SO with version like libfreetype.so.6.20.2

19317362:
[#6253 ](https://github.com/xmake-io/xmake/issues/6253)

### DIFF
--- a/xmake/modules/utils/binary/deplibs.lua
+++ b/xmake/modules/utils/binary/deplibs.lua
@@ -81,7 +81,7 @@ function _get_all_depends_by_objdump(binaryfile, opt)
                 else
                     if line:startswith("NEEDED") then
                         local filename = line:split("%s+")[2]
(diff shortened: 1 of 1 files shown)


xmake/modules/utils/binary/deplibs.lua
```diff
@@ -81,7 +81,7 @@ function _get_all_depends_by_objdump(binaryfile, opt)
                 else
                     if line:startswith("NEEDED") then
                         local filename = line:split("%s+")[2]
-                        if filename and filename:endswith(".so") then
+                        if filename and filename:endswith(".so") or filename:match(".-%.so%.%d+") then
```
waruqi, line 84: I think we should use `filename:find("%.so%.%d+$")` instead of `filename:match(".-%.so%.%d+")`

PR #6209, opened 2025-03-11
Add build jobgraph support

waruqi:
https://github.com/xmake-io/xmake/issues/6202

PR #6281, opened 2025-04-02
feat(verilator): Support Verilator target build to shared library

### LINKED ISSUE #6280: Support Verilator target build to shared library

Similar to #5060 , however, I need build shared library.

Using the Verilator command `--lib-create <lib_name>` can automatically generate some FFI-friendly APIs, like:
```cpp
// Checks to make sure the .sv wrapper and library agree
    void top_protectlib_check_hash(int protectlib_hash__V) {
        const int expected_hash__V = 1473972723U;
        if (protectlib_hash__V != expected_hash__V) {
            fprintf(stderr, "%%Error: cannot use top library, Verliog (%u) and library (%u) hash values do not agree\n", protectlib_hash__V, expected_hash__V);
            std::exit(EXIT_FAILURE);
        }
    }
    
    // Creates an instance of the library module at initial-time
    // (one for each instance in the user's design) also evaluates
    // the library module's initial process
    void* top_protectlib_create(const char* scopep__V) {
        Vtop_container* const handlep__V = new Vtop_container{scopep__V};
        return handlep__V;
    }
```
This makes it easier to call Verilator from other languages.

_No response_

### Additional context

_No response_

TimeKeepper:
close #6280

### DIFF
--- a/tests/projects/embed/verilator/shared/src/main.v
+++ b/tests/projects/embed/verilator/shared/src/main.v
@@ -1,0 +1,6 @@
+module hello;
+  initial begin
+    $display("hello world!");
+    $finish ;
+  end
+endmodule
--- a/tests/projects/embed/verilator/shared/src/sim_main.cpp
+++ b/tests/projects/embed/verilator/shared/src/sim_main.cpp
@@ -1,0 +1,14 @@
+#include "hello.h"
+#include "verilated.h"
+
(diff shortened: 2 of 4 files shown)


waruqi:
wait for https://github.com/xmake-io/xmake/pull/6209

We need to add support for the new jobgraph to replace batchjobs

PR #6286, opened 2025-04-05
[toolchains(armclang)] fix target and cpu flags for MDK

jnshuiji:
- Always set `--target=arm-arm-none-eabi` for compiler and assembler to match MDK's armclang requirements
- Add `-masm=auto` to asflags when using armclang assembler for better compatibility
- Unify ldflags cpu format: change from space separator to `--cpu=<arch>` format (note: the `--cpu` parameter is case insensitive)
- Fix incorrect architecture target handling reported in #6283


waruqi:
thanks

PR #6288, opened 2025-04-06
Fix git clone failure

Elite-stay:
Fix git clone failure in older version #6246

### DIFF
--- a/xmake/modules/private/action/require/impl/actions/download.lua
+++ b/xmake/modules/private/action/require/impl/actions/download.lua
@@ -122,7 +122,7 @@ function _checkout(package, url, sourcedir, opt)
 
             -- clone whole history and tags
             -- @see https://github.com/xmake-io/xmake/issues/5507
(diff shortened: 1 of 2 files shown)


xmake/modules/private/action/require/impl/actions/download.lua
```diff
@@ -122,7 +122,7 @@ function _checkout(package, url, sourcedir, opt)
 
             -- clone whole history and tags
             -- @see https://github.com/xmake-io/xmake/issues/5507
-            git.clone(url, {treeless = true, checkout = false, longpaths = longpaths, outputdir = packagedir})
+            git.clone(url, {treeless = git.clone.can_treeless(), checkout = not git.clone.can_treeless(), longpaths = longpaths, outputdir = packagedir})
```
waruqi, line 125: We should only handle this in git.clone

Elite-stay: done.

Elite-stay:
close #6246

Issue #6298, opened 2025-04-08
Unable to Build CMake-based Package for WASM on Arch Linux - `emscripten not found!`


Doekin:
### Xmake Version

v2.9.9+20250408

### Operating System Version and Architecture

Arch Linux

### Describe Bug

Despite having `emscripten` installed on Arch Linux, building CMake-based packages fails with the error `emscripten not found!`. In contrast, autoconf-based packages build successfully.

### Expected Behavior

CMake-based packages should build for WASM using the system-installed `emscripten`.

### Project Configuration

Command used:
```bash
xrepo install -vD -p wasm gflags
```

### Additional Information and Error Logs

```console
$ xrepo install -vD -p wasm gflags
/usr/bin/xmake f -c --require=n -vD -p wasm
checking for architecture ... wasm32
checking for emsdk directory ... no
checking for emcc ... /usr/lib/emscripten/emcc
configure
{
    buildir = build
    network = public
    arch = wasm32
    host = linux
    ccache = true
    proxy_pac = pac.lua
    theme = default
    diagnosis = true
    plat = wasm
    ndk_stdcxx = true
    mode = release
    kind = static
    clean = true
}
/usr/bin/xmake require -v -D -j 14 --extra={system=false} gflags
checking for git ... /usr/bin/git
checking for gzip ... /usr/bin/gzip
checking for tar ... /usr/bin/tar
/usr/bin/git rev-parse HEAD
checking for gcc ... /usr/bin/gcc
checkinfo: cannot runv(nim --version), No such file or directory
checking for nim ... no
checkinfo: cannot runv(nim --version), No such file or directory
checking for nim ... no
checking for cmake ... no
checking for cmake ... /usr/bin/cmake
finding gflags from xmake ..
checking for xmake::gflags ... no
note: install or modify (m) these packages (pass -y to skip confirm)?
in myrepo:
  -> gflags v2.2.2 [license:BSD-3-Clause]
please input: y (y/n/m)

checking for ping ... /usr/bin/ping
pinging the host(github.com) ... 107 ms
/usr/bin/tar -xf /home/leemu/.xmake/cache/packages/2504/g/gflags/v2.2.2/v2.2.2.tar.gz -C source.tmp
patching /run/media/leemu/D8C48579C4855B20/projects/xmake-repo/packages/g/gflags/patches/v2.2.2/fix-cmake.patch to gflags-v2.2.2 ..
[...]

PR #6304, opened 2025-04-09
improve `emscripten` detection by including `emcc.py` search

Doekin:
The `emscripten` package in the [APT repository](https://packages.debian.org/bookworm/all/emscripten/filelist) does not include `emsdk.py`, leading to detection issues in xmake.

This PR improves detection by also searching for `emcc.py`, which aids in locating the installed `emscripten`. However, specifying the `emsdk` path may still be necessary. For example:

- On Debian: `xmake g --emsdk=/usr/share/emscripten/`
- On Arch: `xmake g --emsdk=/usr/lib/emscripten`

(Note: The `emscripten` installation path may change in the future, so the paths mentioned here might become outdated.)

https://github.com/xmake-io/xmake/issues/6298

### DIFF
--- a/xmake/modules/detect/sdks/find_emsdk.lua
+++ b/xmake/modules/detect/sdks/find_emsdk.lua
@@ -33,9 +33,16 @@ function _find_emsdkdir(sdkdir)
         table.insert(paths, sdkdir)
     end
     table.insert(paths, "$(env EMSDK)")
+    if is_host("linux") then
+        table.join2(paths, {"/usr/share/emscripten/", "/usr/lib/emscripten/"})
+    end
     local emsdk = find_file("emsdk.py", paths, {suffixes = subdirs})
     if emsdk then
         return path.directory(emsdk)
+    end
+    local emcc_py = find_file("emcc.py", paths, {suffixes = subdirs})
+    if emcc_py then
+        return path.directory(emcc_py)
     end
 end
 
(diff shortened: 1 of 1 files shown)


xmake/modules/detect/sdks/find_emsdk.lua
```diff
@@ -55,6 +59,8 @@ function _find_emsdk(sdkdir)
     local emcc = find_file("emcc", sdkdir, {suffixes = subdirs})
     if emcc then
         emscripten = path.directory(emcc)
+    elseif find_file("emcc.py", sdkdir) then
+        emscripten = sdkdir
```
waruqi, line 63:
```lua
    local emcc = find_file("emcc", sdkdir, {suffixes = subdirs})
    if not emcc then
        emcc = find_file("emcc.py", sdkdir)
    end
    if emcc then
```

waruqi:
> * On Debian: `xmake g --emsdk=/usr/share/emscripten/`
> * On Arch: `xmake g --emsdk=/usr/lib/emscripten`
> 
> (Note: The `emscripten` installation path may change in the future, so the paths mentioned here might become outdated.)

We can add some default search paths here.

```lua
if is_host("linux") then
    table.insert(paths, "/usr/share/emscripten")
end
```
https://github.com/xmake-io/xmake/blob/af49507ab63373386ffb84c899be7beb89d5fddc/xmake/modules/detect/sdks/find_emsdk.lua#L35

PR #6321, opened 2025-04-14
package(autoconf): De-duplicate `LDFLAGS`

Doekin:
This PR aims to eliminate duplicate linker flags that appear after merging `ldflags` and `shflags`. This adjustment resolves the issue where `icu4c` fails to build for Android in `Git Bash` on Windows due to redundant flags.

### DIFF
--- a/xmake/modules/package/tools/autoconf.lua
+++ b/xmake/modules/package/tools/autoconf.lua
@@ -362,7 +362,7 @@ function buildenvs(package, opt)
     end
     if ldflags or shflags then
         -- autoconf does not use SHFLAGS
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/autoconf.lua
```diff
@@ -362,7 +362,7 @@ function buildenvs(package, opt)
     end
     if ldflags or shflags then
         -- autoconf does not use SHFLAGS
-        envs.LDFLAGS   = table.concat(_translate_paths(table.join(ldflags or {}, shflags)), ' ')
+        envs.LDFLAGS   = table.concat(table.unique(_translate_paths(table.join(ldflags or {}, shflags))), ' ')
```
waruqi, line 365: use reverse_unique to reserve link flags orders.

PR #6326, opened 2025-04-16
do windeployqt for qt.shared deps, fix windeployqt bindir bug

choyy:
https://github.com/xmake-io/xmake/pull/6323

对依赖的qt.shared执行windeployqt

### DIFF
--- a/xmake/rules/qt/install/mingw.lua
+++ b/xmake/rules/qt/install/mingw.lua
@@ -26,18 +26,25 @@ import("lib.detect.find_path")
 import("lib.detect.find_file")
 import("detect.sdks.find_qt")
 
--- get install directory
-function _get_installdir(target)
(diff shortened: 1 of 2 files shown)


xmake/rules/qt/install/mingw.lua
```diff
@@ -26,18 +26,26 @@ import("lib.detect.find_path")
...
 
-    local targetfile = target:targetfile()
-    local installdir = _get_installdir(target)
-    local installfile = path.join(installdir, "bin", path.filename(targetfile))
+    local bindir = _get_bindir(target)
+    local targetfile = path.join(bindir, path.filename(target:targetfile()))
+    local installfiles = {}
+    table.insert(installfiles, targetfile)
+    local deps = target:deps()
+    for _, tgt in pairs(deps) do
```
waruqi, line 43:
```lua
for _, t in ipairs(target:orderdeps()) do
end
```

PR #6380, opened 2025-05-01
(llvm toolchain) fix cross support

Arthapz:
i'm not sure about the impact of this, so it's a draft for now

PR #6381, opened 2025-05-01
(cmake, llvm toolchain) when using --toolchain=llvm on windows, dependencies should be build with it

Arthapz:


Doekin:
With this change, building CMake packages using `--toolchain=llvm` successfully uses `clang` instead of MSBuild. Thanks for addressing this!

However, installing package fails with the following error:  
```console
finding libogg from xmake ..
checking for xmake::libogg ... libogg v1.3.5
{
  version = "v1.3.5",
  static = true,
  sysincludedirs = {
    "C:\Users\leemu\AppData\Local\.xmake\packages\l\libogg\v1.3.5\0df42afefcc14ed29c81a7b0567b9dec\include"
  }
}

error: @programdir\core\package\package.lua:2524: package(libogg): links not found!
```

In the `installdir.failed` directory, I found `lib/libogg.a`. It seems the check might not account for `lib<xxx>.a` on Windows.

Arthapz:
> With this change, building CMake packages using `--toolchain=llvm` successfully uses `clang` instead of MSBuild. Thanks for addressing this!
> 
> However, installing package fails with the following error:
> 
> ```
> finding libogg from xmake ..
> checking for xmake::libogg ... libogg v1.3.5
> {
>   version = "v1.3.5",
>   static = true,
>   sysincludedirs = {
>     "C:\Users\leemu\AppData\Local\.xmake\packages\l\libogg\v1.3.5\0df42afefcc14ed29c81a7b0567b9dec\include"
>   }
> }
> 
> error: @programdir\core\package\package.lua:2524: package(libogg): links not found!
> ```
> 
> In the `installdir.failed` directory, I found `lib/libogg.a`. It seems the check might not account for `lib<xxx>.a` on Windows.

oups wrong PR i wanted your insight [here](https://github.com/xmake-io/xmake/pull/6380) :D but it's nice this one is usefull to you

PR #6402, opened 2025-05-07
cargo: fix package cross-compilation

SirLynix:
Fixes cross-compilation with Cargo by passing the proper triple automatically
Also, the /host folder has to be passed to rustc to find some library (tested with naga)

Fixes https://github.com/xmake-io/xmake-repo/pull/7023

### DIFF
--- a/xmake/modules/package/manager/cargo/find_package.lua
+++ b/xmake/modules/package/manager/cargo/find_package.lua
@@ -136,6 +136,12 @@ function main(name, opt)
     local librarydir_host = path.join(opt.installdir, "lib", "host")
     local libfiles = os.files(path.join(librarydir, "*.rlib"))
 
+    local frameworkdirs = {}
+    table.insert(frameworkdirs, librarydir)
+    if os.exists(librarydir_host) then
+        table.insert(frameworkdirs, librarydir_host)
+    end
+
     -- @see https://github.com/xmake-io/xmake/issues/4228
     local libfiles_native
     local plat = opt.plat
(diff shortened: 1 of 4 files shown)


xmake/modules/package/manager/cargo/find_package.lua
```diff
@@ -136,6 +136,12 @@ function main(name, opt)
     local librarydir_host = path.join(opt.installdir, "lib", "host")
     local libfiles = os.files(path.join(librarydir, "*.rlib"))
 
+    local frameworkdirs = {}
+    table.insert(frameworkdirs, librarydir)
+    if os.exists(librarydir_host) then
```
waruqi, line 141: os.isdir

xmake/toolchains/rust/xmake.lua
```diff
@@ -28,58 +28,14 @@ toolchain("rust")
...
-                end
-            elseif toolchain:is_plat("bsd") then
-                target = target .. "-unknown-freebsd"
-            elseif toolchain:is_plat("wasm") then
-                target = target .. "-unknown-unknown"
-            end 
-        end
-
+ 
+        local target = import("private.tools.rust.target_triple")(toolchain:plat(), toolchain:arch(), opt)
```
waruqi, line 36: move import to head

xmake/toolchains/rust/xmake.lua
```diff
@@ -28,58 +28,16 @@ toolchain("rust")
...
-            end 
+        local opt = {}
+        if toolchain:config("appledev") == "simulator" then
+            opt.apple_sim = true
         end
-
+ 
+        local target = target_triple(toolchain:plat(), toolchain:arch(), opt)
         if target then
+            toolchain:add("rcflags", "--verbose")
```
waruqi, line 40: why? if we want to get verbose info, we should use `option.get("verbose")` and `option.get("diagnosis")`

SirLynix: It's a leftover, I'm removing it

waruqi:
It will generate incorrect arch. https://github.com/xmake-io/xmake/issues/6574

on macOS

```
xmake f -a aarch64-unknown-none
/usr/local/bin/rustc --target=aarch64-unknown-none-apple-darwin
```

PR #6409, opened 2025-05-08
(cmake, meson) append shared/static libraries by default when relevent

Arthapz:

------


xmake/modules/package/tools/cmake.lua
```diff
@@ -432,6 +432,9 @@ function _get_configs_for_generic(package, configs, opt)
     else
         table.insert(configs, "-DCMAKE_BUILD_TYPE=Release")
     end
+    if package:is_library() then
+        table.insert(configs, "-DBUILD_SHARED_LIBS=".. (package:config("shared") and "ON" or "OFF"))
```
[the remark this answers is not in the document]
Arthapz: done

xmake/modules/package/tools/cmake.lua
```diff
@@ -644,6 +655,8 @@ function _get_configs_for_cross(package, configs, opt)
     opt = opt or {}
     opt.cross                      = true
     local envs                     = {}
+    envs.CMAKE_BUILD_TYPE          = package:is_debug() and "Debug" or "Release"
+    envs.BUILD_SHARED_LIBS         = package:config("shared") and "ON" or "OFF"
```
waruqi, line 659: why? _get_configs_for_generic will add them, and we need also check flags exist

xmake/modules/package/tools/meson.lua
```diff
@@ -351,6 +351,11 @@ function _get_configs(package, configs, opt)
         table.insert(configs, "-Db_sanitize=address")
     end
 
+    -- add library kind
+    if package:is_library() then
+        table.insert(configs, "-Ddefault_library=".. (package:config("shared") and "shared" or "static"))
+    end
```
waruqi, line 357: we need also has_already_libflag

xmake/modules/package/tools/cmake.lua
Arthapz: it doesn't seem to me that _get_configs_for_generic is called for cross

here it's generic OR cross
```
    if package:is_cross() then
        _get_configs_for_cross(package, configs, opt)
    else
        _get_configs_for_generic(package, configs, opt)
    end
```
https://github.com/xmake-io/xmake/blob/accf9cdc27b899d8cbece2987408986c1e0d2378/xmake/modules/package/tools/cmake.lua#L488

and here if is_cross is true, _get_configs_for_generic is not called

```
    elseif package:is_cross() then
        _get_configs_for_cross(package, configs, opt)
    elseif package:config("toolchains") then
        -- we still need find system libraries,
        -- it just pass toolchain environments if the toolchain is compatible with host
        if _is_toolchain_compatible_with_host(package) then
        else
            _get_configs_for_cross(package, configs, opt)
        end
    else
        _get_configs_for_generic(package, configs, opt)
    end
```
or i miss something ?

waruqi: ok, but we should check _configs_str here too.

waruqi: ok

PR #6421, opened 2025-05-11
(C++ modules support) improve std module support for gcc and clang

Arthapz:
this patch default cxx11abi for gcc >= 15 it should be safe now and is a requirement for libstdc++ module
factorize msstl std module support for clang/llvm
add support for libstdc++ module for clang/llvm
add fallback for finding libstdc++ module manifest for gcc

### DIFF
--- a/xmake/rules/c++/modules/clang/support.lua
+++ b/xmake/rules/c++/modules/clang/support.lua
@@ -277,7 +277,11 @@ function get_stdmodules(target)
                     end
                 end
             elseif cpplib == "stdc++" then
-                -- libstdc++ doesn't have a std module file atm
+                -- dont be greedy and don't enable stdc++ std module support for llvm < 19
+                local clang_version = get_clang_version(target)
+                if clang_version and semver.compare(clang_version, "19.0") >= 0 then
+                    return import(".gcc.support").get_stdmodules(target, {warn = false})
+                end
             elseif cpplib == "msstl" then
                 -- msstl std module file is not compatible with llvm < 19
                 local clang_version = get_clang_version(target)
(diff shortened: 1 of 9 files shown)


xmake/rules/c++/modules/clang/support.lua
```diff
@@ -71,7 +71,7 @@ function _get_cpplibrary_name(target)
     elseif target:has_runtime("MD", "MT", "MDd", "MTd") then
         return "msstl"
     end
-    if target:is_plat("macosx") then
+    if target:is_plat("macosx") or target:is_plat("ios") or target:is_plat("tvos") then
```
waruqi, line 74: ios -> iphoneos
tvos -> appletvos

waruqi: `if target:is_plat("macosx", "iphoneos", "appletvos") then`

xmake/rules/c++/modules/gcc/support.lua
```diff
@@ -139,33 +146,47 @@ function _get_std_module_manifest_path(target)
...
         end
     end
-    -- fallback on custom detection
-    -- manifest can be found alongside libstdc++.so
-
-    -- TODO
+    local ext = ".so"
+    if target:is_plat("windows") or target:is_plat("mingw") then
+        ext = ".dll"
+    elseif target:is_plat("macos") or target:is_plat("ios") or target:is_plat("tvos") then
```
waruqi, line 152: macos -> macosx, ios, tvos is incorrect too

xmake/rules/c++/modules/msvc/support.lua
```diff
@@ -117,32 +116,34 @@ function has_two_phase_compilation_support(_)
...
-                    return {path.normalize(path.join(stdmodulesdir, "std.ixx")), path.normalize(path.join(stdmodulesdir, "std.compat.ixx"))}
-                end
+function get_stdmodules(target, opt)
+    opt = opt or {}
+    if not target:policy("build.c++.modules.std") then
+        return
+    end
+    local msvc
+    if opt.toolchain then
+        msvc = import("core.tool.toolchain", {anonymous = true}).load("msvc", {plat = opt.toolchain:plat(), arch = opt.toolchain:arch()})
```
waruqi, line 126: move import to head

PR #6425, opened 2025-05-12
(autotools) fix ld override when ld binary is not in the same directo…

Arthapz:
shared  packages installation are broken when --sdk is set with a bin directory without ld binary
e.g when using llvm toolchain with --sdk, ld binary may not exists in the bin directory and result with a non-existent path which make autotools use clang++ as linker wich prevent .so to be generated

### DIFF
--- a/xmake/modules/package/tools/autoconf.lua
+++ b/xmake/modules/package/tools/autoconf.lua
@@ -449,7 +449,14 @@ function buildenvs(package, opt)
             name = name:gsub("gcc-%d+", "ld")
             name = name:gsub("g%+%+$", "ld")
             name = name:gsub("g%+%+%-%d+", "ld")
-            envs.LD = dir and path.join(dir, name) or name
+            if os.isfile(path.join(dir, name)) then
+                envs.LD = path.join(dir, name)
+            else
+                local ld = find_tool("ld")
+                if ld and path.filename(ld.program) == name then
+                    envs.LD = ld.program
+                end
+            end
         end
         -- we need use clang++ as cxx, autoconf will use it as linker
         -- https://github.com/xmake-io/xmake/issues/2170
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/autoconf.lua
```diff
@@ -439,17 +439,10 @@ function buildenvs(package, opt)
...
-            name = name:gsub("clang%+%+$", "ld")
-            name = name:gsub("clang%+%+%-%d+", "ld")
-            name = name:gsub("clang$", "ld")
-            name = name:gsub("clang%-%d+", "ld")
-            name = name:gsub("gcc$", "ld")
-            name = name:gsub("gcc-%d+", "ld")
-            name = name:gsub("g%+%+$", "ld")
-            name = name:gsub("g%+%+%-%d+", "ld")
-            envs.LD = dir and path.join(dir, name) or name
+            local ld = find_tool("ld")
```
waruqi, line 442: Using find_tool may find the system ld instead of the ld corresponding to the current package toolchain, which may break many packages, especially for some cross-compilation toolchains.

/sdk/bin/linux-arm-clang -> /sdk/bin/linux-arm-ld

now:

/sdk/bin/linux-arm-clang -> /usr/bin/ld

Arthapz: right, maybe i can add the path of the compiler in the search directory list of find_tool, would it work ?

waruqi: no, It is difficult to keep the package toolchain consistent with the result of find_tool. I still recommend that all tool-related information be obtained from package:build_getenv -> package:tool, which can correctly obtain the sdk/bin path corresponding to clang/ld.

Therefore, I currently use the path file name replacement method to obtain ld under the corresponding path.

Arthapz: i changed it to only fallback on find_tool if , is it more acceptable ? we can't use package:build_getenv because the point here is that only ld work with autotools, when clang++ is set as ld, it doesn't generate .so for shared libraries

PR #6438, opened 2025-05-13
cmakelists: Fix CMAKE_COMPILER_ID => CMAKE_<lang>_COMPILER_ID

SirLynix:
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html

### DIFF
--- a/xmake/plugins/project/cmake/cmakelists.lua
+++ b/xmake/plugins/project/cmake/cmakelists.lua
@@ -784,7 +784,7 @@ function _add_target_compile_options(cmakelists, target, outputdir)
     for _, toolname in toolnames:keys() do
         local name = compilernames[toolname]
         if name then
-            cmakelists:print("if(CMAKE_COMPILER_ID STREQUAL \"%s\")", name)
+            cmakelists:print("if(CMAKE_C_COMPILER_ID STREQUAL \"%s\")", name)
             _add_target_compile_options_for_compiler(toolname)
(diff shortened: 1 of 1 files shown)


waruqi:
Which one is more correct? https://github.com/xmake-io/xmake/pull/6439

SirLynix:
I think it would be mine because I use CMAKE_C_COMPILER_ID when it's not sure and CMAKE_CXX_LINKER_ID, what do you think @PierreEVEN?

PierreEVEN:
> I think it would be mine because I use CMAKE_C_COMPILER_ID when it's not sure and CMAKE_CXX_LINKER_ID, what do you think @PierreEVEN?

I was wondering if "C" is better than "CXX". I saw "CXX" everywhere else so I choose to keep it consistent but you may be right.

I couldn't find any reference of CMAKE_CXX_LINKER_ID in the documentation, is it right ?
If it should be `CMAKE_CXX_COMPILER_LINKER_ID`, the value for clang is AppleClang so it won't work as expected.

PierreEVEN:
> Which one is more correct?

https://github.com/xmake-io/xmake/pull/6439 (mine) works as long as CXX is enabled. but it has the same problem as this pr if C is not enabled. I'm fixing it to handle both cases...

SirLynix:
Closing because #6439 will have a better fix

PR #6439, opened 2025-05-13
Fixed cmake generator that was using 'CMAKE_COMPILER_ID' 

PierreEVEN:
Fix previous merge that failed to fix issue with compiler detection (was not a regression, but neither a progress)

CMAKE_COMPILER_ID is not the right variable (but works in certain conditions)
The right one is [CMAKE__CXX_COMPILER_ID](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html) 

Now it should be the right one.

### DIFF
--- a/xmake/plugins/project/cmake/cmakelists.lua
+++ b/xmake/plugins/project/cmake/cmakelists.lua
@@ -355,18 +355,24 @@ function _add_project(cmakelists, outputdir)
     if _can_native_support_for_cxxmodules() then
         cmakelists:print("set(CMAKE_CXX_SCAN_FOR_MODULES ON)")
     end
+    local languages = _get_project_languages()
     if project_name then
         local project_info = ""
         local project_version = project.version()
         if project_version then
             project_info = project_info .. " VERSION " .. project_version
         end
-        local languages = _get_project_languages()
         if languages then
             cmakelists:print("project(%s%s LANGUAGES %s)", project_name, project_info, table.concat(languages, " "))
         else
             cmakelists:print("project(%s%s)", project_name, project_info)
         end
+    end
+    -- Define a language-independant global compiler_id variable
+    if (languages and #languages > 0) then
+        cmakelists:print("set(CMAKE_COMPILER_ID ${CMAKE_%s_COMPILER_ID})", _get_project_languages()[1])
+    else
+        cmakelists:print("set(CMAKE_COMPILER_ID ${CMAKE_C_COMPILER_ID})") -- C should be defined by default if not specified
     end
     cmakelists:print("")
 end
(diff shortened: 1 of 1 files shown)


waruqi:
Which one is more correct? https://github.com/xmake-io/xmake/pull/6438

xmake/plugins/project/cmake/cmakelists.lua
```diff
@@ -355,19 +355,25 @@ function _add_project(cmakelists, outputdir)
...
             cmakelists:print("project(%s%s)", project_name, project_info)
         end
     end
+    -- Define a language-independant global compiler_id variable
+    if (languages and #languages > 0) then
+        cmakelists:print("set(CMAKE_COMPILER_ID CMAKE_%s_COMPILER_ID)", _get_project_languages()[1])
+    else
+        cmakelists:print("set(CMAKE_COMPILER_ID CMAKE_C_COMPILER_ID)") -- C should be defined by default if not specified
+    end
     cmakelists:print("")
```
PierreEVEN, line 377: Whoops that's totally not a valid fix !

PierreEVEN: Fixed !

PierreEVEN:
> Which one is more correct? #6438

I've just fixed this PR which now handles any language so it's better in theory.
It should be ready for a merge now.

xmake/plugins/project/cmake/cmakelists.lua
```diff
@@ -811,7 +817,7 @@ function _add_target_values(cmakelists, target, name)
         if name:endswith("s") then
             name = name:sub(1, #name - 1)
         end
-        cmakelists:print("if(CMAKE_COMPILER_ID STREQUAL \"MSVC\")")
+        cmakelists:print("if(XMAKE_GLOBAL_COMPILER_ID STREQUAL \"MSVC\")")
```
waruqi, line 820: Where is XMAKE_GLOBAL_COMPILER_ID defined?

PierreEVEN: Crap ! My bad, I probably did something weird with my last commit as it was missing the remaining changes.
Thanks for seeing this, I should have checked twice. (now fixed : https://github.com/xmake-io/xmake/pull/6439/commits/7f0d792d890e15677057c545a7b3d5f7eb52da52)

Also if you prefer a better name for this property I'm open to suggestions :)

waruqi: CURRENT_COMPILER_ID? we should not add `XMAKE_` prefix

PierreEVEN: That's fair enough, it should be good now. :+1:

PR #6446, opened 2025-05-15
Fix unitybuild with pch/modules

waruqi:
https://github.com/xmake-io/xmake/issues/6420
https://github.com/xmake-io/xmake/pull/6440


Arthapz (approved):
Lgtm

PR #6450, opened 2025-05-15
(C++ modules support) Optimize module scanner

Arthapz:
some perf regression  was introduced in the module scanner recently, this PR fix them

[perf-ezgif.com-gif-to-webm-converter.webm](https://github.com/user-attachments/assets/b0c87064-3c4f-46cb-9fd8-9eb3099d2555)

### DIFF
--- a/xmake/rules/c++/modules/scanner.lua
+++ b/xmake/rules/c++/modules/scanner.lua
@@ -374,8 +374,6 @@ end
 -- patch sourcebatch
 function _patch_sourcebatch(target, sourcebatch)
 
-    local cachekey = target:fullname() .. ".patched_sourcebatch"
-
     -- target deps modules
     local depsmodules = _get_targetdeps_modules(target) or {}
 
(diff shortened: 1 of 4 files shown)


xmake/rules/c++/modules/scanner.lua
```diff
@@ -131,67 +133,71 @@ function _parse_dependencies_data(target, moduleinfos)
 
     local modules
     local modules_names = hashset.new()
+    local jobs = jobpool.new()
```
waruqi, line 136: please use jobgraph instead of it. jobpool will be deprecated in future.

xmake/rules/c++/modules/support.lua
```diff
@@ -236,21 +238,25 @@ end
 -- load module infos
 function load_moduleinfos(target, sourcebatch)
     local moduleinfos
+    local jobs = jobpool.new()
```
waruqi, line 241: and here

Arthapz:
now i need to rebase this to dev and it's finished

xmake/rules/c++/modules/support.lua
```diff
@@ -23,6 +23,8 @@ import("core.base.json")
 import("core.base.hashset")
 import("core.cache.memcache", {alias = "_memcache"})
 import("core.cache.localcache", {alias = "_localcache"})
+import("private.async.jobpool")
```
A2va, line 26: Nothing much, but the import is not used

waruqi:
Thanks, I will review it in the next few days, I'm a bit busy recently.

Arthapz:
> Thanks, I will review it in the next few days, I'm a bit busy recently.

no problem

Issue #6461, opened 2025-05-17
Fix Cross Compilation for Windows


mccakit:
### Xmake 版本

2.9.9

### 操作系统版本和架构

Windows 10

### 描述问题

Cross compilation doesn't works for windows host.

### 期待的结果

Cross compilation should work for windows

### 工程配置

```lua
package("glfw")
    add_deps("cmake")
    set_sourcedir("glfw-3.4")
    on_install(function (package)
        local configs = {}
        import("package.tools.cmake")
        cmake.install(package, configs)
    end)
package_end()
add_requires("glfw")

target("glad")
    set_kind("static")
    add_files("glad/src/*.c")
    add_headerfiles("glad/include/glad/*.h")
    add_headerfiles("glad/include/KHR/*.h")
    add_includedirs("glad/include",{public=true})

target("hi")
    set_kind("binary")
    add_files("src/*.cpp")
    add_deps("glad")
    add_packages("glfw")
    if is_plat("windows") then
        add_syslinks("opengl32", "gdi32", "user32", "shell32")
    elseif is_plat("linux") then
        add_syslinks("GL", "m", "dl", "X11", "pthread")
    end
```

```bash
xmake config --toolchain=llvm --plat=linux --arch=x86_64 --sdk=C:\devtools\llvm\
```

This fails, so do this
```bash
xmake config --toolchain=llvm --plat=cross --cross=x86_64-pc-linux-gnu
```
and this
```bash
xmake config --toolchain=llvm --plat=cross `
--cross=x86_64-pc-linux-gnu `
--sdk=C:/devtools/llvm `
--cxflags=C:/devtools/sysroots/deb `
--ldflags=C:/devtools/sysroots/deb
```
and this
```bash
xmake config --toolchain=llvm --plat=linux --arch=x86_64 --sdk=C:\devtools\llvm\ --runtimes=stdc++_shared
```

### 附加信息和错误日志

I could write a lot, but the main thing is standard library cannot be found anad pacakge install fails due to not finding os specifics stuff like pthread

waruqi:
you can add `-vD` to see verbose logs and debug it.

https://github.com/xmake-io/xmake/blob/2b88ba37fa0a9957b8ab24cf1bf6da59e2c39f21/xmake/toolchains/llvm/xmake.lua#L66

or try and wait this patch. https://github.com/xmake-io/xmake/pull/6380

mccakit:
```bash
scoop uninstall xmake
scoop install xmake

[...]

PR #6471, opened 2025-05-18
improve llvm cross-compilation

waruqi:
1. https://github.com/xmake-io/xmake/issues/6461

```console
> xmake f -p cross --cross=x86_64-pc-linux-gnu --toolchain=llvm --sdk=C:\xxx\llvm -cvD
```

std header has been not supported yet

2. fix target/arch on macos Arm64

### DIFF
--- a/xmake/toolchains/llvm/check.lua
+++ b/xmake/toolchains/llvm/check.lua
@@ -28,6 +28,14 @@ import("detect.sdks.find_cross_toolchain")
 -- find xcode on macos
 function _find_xcode(toolchain)
 
+    -- get apple device
+    local appledev = toolchain:config("appledev") or config.get("appledev")
+    if appledev and appledev == "simulator" then
+        appledev = "simulator"
+    elseif not toolchain:is_plat("macosx") and toolchain:is_arch("i386", "x86_64") then
+        appledev = "simulator"
+    end
+
     -- find xcode
     local xcode_sdkver = toolchain:config("xcode_sdkver") or config.get("xcode_sdkver")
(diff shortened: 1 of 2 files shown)


GoodenoughPhysicsLab:
What if I cross compiling to x86_64-w64-mingw32 target, can't I use is_plat("mingw") to detect platform?

GoodenoughPhysicsLab:
That means, can we infer target arch and target platform according to --cross?

waruqi:
> What if I cross compiling to x86_64-w64-mingw32 target, can't I use is_plat("mingw") to detect platform?

why not use llvm-mingw + mingw toolchain. not llvm toolchain

GoodenoughPhysicsLab:
> why not use llvm-mingw + mingw toolchain. not llvm toolchain

I do use llvm-mingw toolchain. The reason why I leave that comments is that clang does not distinguish between local compilation and cross compilation. Therefore, compiling with clang's default target and a sysroot clang detected automatically should be the same behavior when I specify the target and sysroot path. Last but not least, I use 24bit-xjkp's [script](https://github.com/24bit-xjkp/toolchains) so that I can use triplet+sysroot style cross compilation. If xmake use triplet+sysroot at the begging, xjkp's hundreds of lines of script should not start with supporting `--sysroot` flag during configuration. And I think that `-rtlib` and `-unwindlib` flag should also be supported by xmake.

GoodenoughPhysicsLab:
Can we use -arch and -plat to infer target triplet instead of using `-p cross` and `--cross=`? That means, xmake also not distinguish native compilation and cross compilation. I am a beginner of xmake, so, please correct me if somewhere is wrong.

PR #6475, opened 2025-05-20
parse msvc compiler version correctly

hotaery:
https://github.com/xmake-io/xmake/issues/6474


waruqi:
https://github.com/xmake-io/xmake/pull/6479

PR #6477, opened 2025-05-20
(C++ modules support) fix clang cl module support

Arthapz:
fixes https://github.com/xmake-io/xmake/issues/6472


xmake/rules/c++/modules/config.lua
```diff
@@ -67,6 +67,10 @@ function main(target)
             wprint("build.c++.modules.tryreuse.discriminate_on_defines is deprecated, please use build.c++.modules.reuse.strict")
         end
 
+        if target:get("toolchains") == "clang-cl" then
```
waruqi, line 70: we should use `target:has_tool("cxx", "clang_cl")`

Arthapz:
ok found a better way to fix it https://github.com/llvm/llvm-project/issues/140825#issuecomment-2900112555

Arthapz:
this path works with our tests cases (pch excluded) but i think we shouldn't advertise that XMake work well with clang-cl + modules as it's not well tested and it seems to cause some issues to CMake guys, see https://github.com/llvm/llvm-project/pull/121046#issuecomment-2900475812

PR #6478, opened 2025-05-21
(C++ modules support) add a policy to reduce verbosity of -v output by putting requires flags into a file

Arthapz:
this patch add the policy `build.c++.modules.hide_dependencies` to hide requires flags (e.g `-fmodule-file=foo.pcm`)
to reduce unwanted verbosity when using -v


Arthapz:
i'll reopen when https://github.com/xmake-io/xmake/pull/6477 is merged

PR #6517, opened 2025-05-31
solutiondir: new vsxmake option

Cappecasper03:

------

### DIFF
--- a/xmake/plugins/project/vsxmake/getinfo.lua
+++ b/xmake/plugins/project/vsxmake/getinfo.lua
@@ -497,7 +497,9 @@ function main(outputdir, vsinfo)
 
                 -- make target with the given mode and arch
                 targets[targetname] = targets[targetname] or {}
-                local _target = targets[targetname]                -- init target info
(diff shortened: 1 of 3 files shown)


waruqi:
? why not use outputdir?

Cappecasper03:
I want to have the .sln in the root directory to be more accessible while all the other files can be in a build folder

waruqi:
I wouldn't consider adding it, it's not a universal option.

waruqi:
It is only used for the vsxmake generator, not a general option. If a new option is to be added, it should be applicable to all project generators, and other project generators must also support it simultaneously, similar to outputdir.

waruqi:
The new option have some conflicts and functional redundancies with  outputdir , and we should always use one option to configure the output path. If the issue is merely to remove the vsxmake2022 subdirectory, it can be supported by policy + outputdir.

Cappecasper03:
Would it be possible to have a policy similar to visual studios "Place solution and project in the same directory" to solve this instead then and also implement it on all generators or something?

PR #6524, opened 2025-06-04
adds target group filtering to `show targets`

lonless9:
Supports xmake show -g mygroup -l targets --json to display targets in specific groups

#6492

### DIFF
--- a/xmake/plugins/show/lists/targets.lua
+++ b/xmake/plugins/show/lists/targets.lua
@@ -28,18 +28,15 @@ import(".showlist")
 -- show all targets (optionally filtered by group)
 function main()
     config.load()
-
-    local group_pattern = option.get("group") and ("^" .. path.pattern(option.get("group")) .. "$")
+    
     local targets = {}
-
+    local group_pattern = option.get("group")
+    if group_pattern then
+        group_pattern = "^" .. path.pattern(group_pattern) .. "$"
+    end
     for name, target in pairs(project.targets()) do
         local group = target:get("group")
-        if (target:is_default() and not group_pattern) or
-           (group_pattern and group and (function()
-                for component in group:gmatch("[^/\\\\]+") do
-                    if component:match(group_pattern) then return true end
-                end
-            end)()) then
+        if (target:is_default() and not group_pattern) or (group_pattern and group and group:match(group_pattern)) then
             table.insert(targets, name)
         end
     end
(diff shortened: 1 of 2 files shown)


xmake/plugins/show/lists/targets.lua
```diff
@@ -19,17 +19,39 @@
...
+
+        local normalized_filter = path.normalize(group_filter)
+        local filter_prefix = normalized_filter .. "/"
+
+        for name, target in pairs(project.targets()) do
+            local target_group = target:get("group")
+            if target_group then
+                local normalized_target_group = path.normalize(target_group)
+                if normalized_target_group == normalized_filter or
+                   normalized_target_group:startswith(filter_prefix) then
```
waruqi: 参考这里写 https://github.com/xmake-io/xmake/blob/6989184ed0b7e61614d3f41bdbcdac4c6075ac77/xmake/modules/private/action/build/target.lua#L731

xmake/plugins/show/lists/targets.lua
```diff
@@ -29,28 +29,21 @@ import(".showlist")
...
+    end
     local targets = {}
 
-    if not group_filter then
-        for name, _ in pairs(project.targets()) do
+    for name, target in pairs(project.targets()) do
+        local group = target:get("group")
+        if (target:is_default() and not group_pattern) or
+           (group_pattern and group and (group:match(group_pattern) or
+            table.find_first_if(group:split("[/\\]"), function(i, component)
```
waruqi: why, please always use same match way.

xmake/plugins/show/lists/targets.lua
```diff
@@ -19,17 +19,30 @@
...
+    local group_pattern = option.get("group") and ("^" .. path.pattern(option.get("group")) .. "$")
     local targets = {}
-    for name, _ in pairs(project.targets()) do
-        table.insert(targets, name)
+
+    for name, target in pairs(project.targets()) do
+        local group = target:get("group")
+        if (target:is_default() and not group_pattern) or
+           (group_pattern and group and (function()
+                for component in group:gmatch("[^/\\\\]+") do
```
waruqi: I've already said, use exactly the same matching method, don't do any extra matching.

https://github.com/xmake-io/xmake/blob/6989184ed0b7e61614d3f41bdbcdac4c6075ac77/xmake/modules/private/action/build/target.lua#L731

PR #6546, opened 2025-06-13
python.module: Feature of arg python

GoodenoughPhysicsLab:
python.module: Feature of arg python

    * Support passing python path by user
      Imagine a situation where user has installed multi python
      versions of python on their computer, find_tool function
      might identify a python version that dismatch the user's
      specified requirement.

---

some important messages in QQ group are listed as follow:

2ae4599285225b7cc3542eb0662af363

I have installed python3.13 and python3.8 (not saying venv). Cause xmake before this patch always use find_tool to detect python, xmake generated soabi not properly.

Therefore, a mechanism that allows user to specify python path is needed.

### DIFF
--- a/xmake/rules/python/module/xmake.lua
+++ b/xmake/rules/python/module/xmake.lua
@@ -27,7 +27,10 @@ rule("python.module")
         local soabi = target:extraconf("rules", "python.module", "soabi")
         if soabi == nil or soabi then
             import("lib.detect.find_tool")
-            local python = assert(find_tool("python3"), "python not found!")
+            local python = target:extraconf("rules", "python.module", "python")
+            if not python then
+                python = assert(find_tool("python3"), "python not found!")
+            end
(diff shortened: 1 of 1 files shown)


waruqi:
不考虑，指定路径不符合这边统一的设计规范。。

即使走自己安装的 python，也是统一走 find_tool ，通过 add_requires + add_packages 绑定 python 环境。python 包可以自己定制 改进 on_fetch 去找系统的。也可以自定义 find_python 去找自己的。

也可以自己设置 PATH

SekaiArendelle:
I guess what you mean is python.module can not detect python headers, python libs, and even should change PATH env var because find_tool only select the first path where.exe returned. How awkward this rule is!

waruqi:
you need use add_requires/add_package to bind python env to target/rules

https://github.com/xmake-io/xmake/blob/ace14b3911da9aeab3afeae2362bc284ca912836/tests/projects/python/cython/example/xmake.lua#L7
https://github.com/xmake-io/xmake/blob/ace14b3911da9aeab3afeae2362bc284ca912836/tests/projects/python/pybind/example_with_soabi/xmake.lua#L7

PR #6561, opened 2025-06-18
Fix: cmake package build failure on Windows when build dir is a file

A2va:
On Windows, if a file with the same name as the build directory exists (e.g., a bazel BUILD file), cmake package can fail because `os.mkdir` will fail. This commit adds a check to remove the file if it exists before creating the build directory.

### DIFF
--- a/xmake/modules/package/tools/cmake.lua
+++ b/xmake/modules/package/tools/cmake.lua
@@ -994,6 +994,11 @@ function _enter_builddir(package, opt)
     if opt.buildir then
         wprint("{buildir = } has been deprecated, please use {builddir = } in cmake.install")
     end
+    if os.is_host("windows") and os.exists(builddir) and os.isfile(builddir) then
+        -- some projects contain both a cmakelist and a bazel BUILD file, 
+        -- which prevents the creation of a directory named 'build' on windows
+        os.rm(builddir)
+    end
     os.mkdir(path.join(builddir, "install"))
     return os.cd(builddir)
 end


A2va:
I'm wondering if it would be better to do this on a per-package basis. The advantage of putting it there is that it supports any build directory. 🤔

waruqi:
I don't think this problem should occur because I have done compatibility processing for this issue.

cmake will use `build_[hash]` as build directory instead of `build`

https://github.com/xmake-io/xmake/blob/9da165f03d6967aa1c82836527eb22702745421a/xmake/core/package/package.lua#L835

waruqi:
If you still encounter this problem, you should open an issue. I will fix it.

A2va:
It is reproducible by running  `xmake l .\scripts\test.lua -vD --shallow "gperftools 83edb60836d87cf1b406e8846b9059c03031e8f5"`
in xmake-repo.

https://github.com/xmake-io/xmake/blob/9da165f03d6967aa1c82836527eb22702745421a/xmake/modules/package/tools/cmake.lua#L993
Since opt.buildir is defined it use it instead of `package:builddir()`
```
opt.buildir build
opt.builddir
package:builddir() build_b8d20ff4
local builddir build
```

Maybe a bug in a test script of xmake-repo ?

waruqi:
This is just an issue for a specific package. you can open a pr to improve this package, not xmake

https://github.com/xmake-io/xmake-repo/blob/dfdb70a4956153764a70deaccc53be58fc90a923/packages/g/gperftools/xmake.lua#L82

SirLynix:
the problem is there: <https://github.com/xmake-io/xmake-repo/blob/dev/packages/g/gperftools/xmake.lua#L82>

I opened https://github.com/xmake-io/xmake-repo/pull/7445 to fix it

PR #6619, opened 2025-07-11
(cmake) fix default flags detection on windows platform

Arthapz:
when using llvm or clang  toolchain on windows platform and using --sdk instead of having clang in PATH env var, xmake fail to get correct default c and cxx flags when installing a package using CMake, this patch fix it

### DIFF
--- a/xmake/modules/package/tools/cmake.lua
+++ b/xmake/modules/package/tools/cmake.lua
@@ -819,15 +819,22 @@ end
 function _get_default_flags(package, configs, buildtype, opt)
     -- The default flags are different for different platforms
     -- @see https://github.com/xmake-io/xmake-repo/pull/4038#issuecomment-2116489448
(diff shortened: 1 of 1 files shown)


xmake/modules/package/tools/cmake.lua
```diff
@@ -819,12 +819,16 @@ end
 function _get_default_flags(package, configs, buildtype, opt)
     -- The default flags are different for different platforms
     -- @see https://github.com/xmake-io/xmake-repo/pull/4038#issuecomment-2116489448
-    local cachekey = buildtype .. package:plat() .. package:arch()
+    local cachekey = buildtype .. package:plat() .. package:arch() .. package:build_getenv("cc") .. package:build_getenv("cxx")
```
waruqi:
```
local cc = package:build_getenv("cc")
```

Issue #6650, opened 2025-07-23
Add interface to get all dependencies of dynamic libraries


star-hengxing:
### Is your feature request related to a problem? Please describe.

- During packaging, automatically perform topological sorting of `links` to avoid link order errors caused by GNU ld.
- Some packaging tasks only need shared libraries, such as Python binding plugins. In xpack on_package, this interface can be used to get all dependent dynamic libraries. xmake install can also provide an option to only install compiled artifacts in the bin directory.

### Describe the solution you'd like

There is already an implementation available, and we can organize it into a general-purpose interface.
https://github.com/xmake-io/xmake/blob/5670477e415e24962c57eac21194c63946ee5e1e/xmake/modules/target/action/install/main.lua#L54-L122

waruqi:
like this? https://github.com/xmake-io/xmake/pull/6651

but maybe we need to resolve @rpath, @loader_path and relative paths.

star-hengxing:
> like this? [#6651](https://github.com/xmake-io/xmake/pull/6651)
> 
> but maybe we need to resolve [@rpath](https://github.com/rpath), @loader_path and relative paths.

Yes, perhaps we need to simulate the operating system's process of finding the actual path of dynamic libraries.

waruqi:
try this patch again. 

```lua
-- get the library dependencies of the give binary files
--
-- @param binaryfile the binary file
-- @param opt        the option, e.g. {recursive = false, resolve_path = true, resolve_hint_paths = {}}
--                      - plat, arch: the platform and architecture
--                      - recursive: recursively get all sub-dependencies, sorted by topology
--                      - resolve_path: try to resolve the file full path, e.g. @rpath, @loader_path, $ORIGIN, relative path ..
--                      - resolve_hint_paths: we can resolve and match path from them
--                      - resolve_search_paths: the search library paths, like: LD_LIBRARY_PATH, DYLD_LIBRARY_PATH, ...
--
function main(binaryfile, opt)
```

```lua
import("utils.binary.deplibs", {alias = "get_depend_libraries"})

local deplibs = get_depend_libraries(opt.binaryfile or target:targetfile(), {
    plat = target:plat(), arch = target:arch(),
    recursive = true, resolve_path = true, resolve_hint_paths = libfiles})
```

PR #6669, opened 2025-08-04
Add binonly option for target install

star-hengxing:
https://github.com/xmake-io/xmake/issues/6650

### DIFF
--- a/xmake/actions/install/xmake.lua
+++ b/xmake/actions/install/xmake.lua
@@ -38,6 +38,7 @@ task("install")
                                               "    xmake install --group=benchmark/*"     },
             {'a', "all",        "k",  nil   , "Install all targets."                      },
             {nil, "nopkgs",     "k",  nil   , "Only install targets without packages."    },
+            {nil, "binonly",    "k",  nil   , "Only install targets with bindir."         },
             {nil, "admin",      "k",  nil   , "Try to request administrator permission to install"},
             {},
             {nil, "target",     "v",  nil   , "The target name. It will install all default targets if this parameter is not specified.",


waruqi:
这个有啥用

SirLynix:
I have a game compiled with xmake that I want to distribute, as it's a game I only need to install its binaries.

waruqi:
> I have a game compiled with xmake that I want to distribute, as it's a game I only need to install its binaries.

Why do not remove add_headerfiles for this target?

SirLynix:
> > I have a game compiled with xmake that I want to distribute, as it's a game I only need to install its binaries.
> 
> Why do not remove add_headerfiles for this target?

Because they're useful for vs/vsxmake project generation. Also my game has libraries (it has multiple executables) and I don't need them to be installed.

Project for reference: https://github.com/DigitalPulseSoftware/ThisSpaceOfMine

star-hengxing:
- cli tools
- Plugins or language bindings, they only require shared libraries

waruqi:
We should use policy instead of adding new `--binonly` option whenever possible, which supports both cli and target configuration in xmake.lua

`xmake f --policies=install.binary_only` and `set_policy("install.binary_only", true)`

SirLynix:
A policy requires a full reconfiguration of the project where the xmake install command can just skip folders, I'm not sure it's more interesting.

Also because in some cases you may be interested in the binaries and in other you may want the headers/libs as well for the same project, for example if python/curl/openssl/etc. were to use xmake it could be interesting to be able to only get the binaries in some cases and the binaries/include/libs in other.

waruqi:
We need to consider more extensibility, otherwise there will be --libonly, --headeronly in the future.

SirLynix:
what about `--no-headers` `--no-libraries` instead?

waruqi:
> Because they're useful for vs/vsxmake project generation.

you should use `add_extrafiles` instead of `add_headerfiles`, if you do not want to install them. it's only for project generation

> what about `--no-headers` `--no-libraries` instead?

Just for personal preference, I don't like using too many `--no-xxx`, `--without-xx`, and `--with-xxx` options; it's just too cluttered.

Currently, install already has a `--nopkgs` option, but I'm not very satisfied with it. Now there are other styles of `--no-xxx` options, which makes it even more confusing.

It would be much better if a single option could be used to handle them, like `--policies`

SirLynix:
> you should use add_extrafiles instead of add_headerfiles, if you do not want to install them. it's only for project generation

This is not a solution, as it wouldn't prevent libraries generation (that would still be installed) and is mostly a hack, some rules (even user-defined) may differentiate headers and extra files.

It also doesn't fix the issue of libraries with standalone binaries where you may want only binaries in some cases and headers in other.

> It would be much better if a single option could be used to handle them, like --policies

--without-headers and --without-libs options allows for composition which is more flexible than --only-bins or equivalent, it's also possible to make an --install=bins,libs,includes filter like option.

I don't think we should use policies for this kind of problem as it's independent of project configuration, you may want to install only binaries or everything for the same project, changing policies isn't straighforward in commandline (it's not easy to enable as you have to list every other policy as well) and requires a full config step.

waruqi:
> --without-headers and --without-libs options allows for composition which is more flexible than --only-bins or equivalent, it's also possible to make an --install=bins,libs,includes filter like option.

I don't like the `--without-`, `--no-` and `--with-` prefixes.

or we can use `--headers/binaries/libraries` options, and they will be enabled by default.

if we want to only install binaries, you can pass `--headers=n` to disable it.

```sh
xmake install --headers=n
```

```sh
xmake install --headers=[y/n] --binaries=[y/n] --libraries=[y/n]
```

waruqi:
try this patch. https://github.com/xmake-io/xmake/pull/6688

PR #6718, opened 2025-08-15
ignore pch flags #6710

waruqi:
https://github.com/xmake-io/xmake/issues/6710


star-hengxing:
```console
compile_commands.lua:145: attempt to index a nil value (local 'arg')
```

waruqi:
再试试

PR #6744, opened 2025-08-25
feat(C++ modules support) use a file for requires flags when --verbose or --diagnosis is supplied

Arthapz:
this PR add a new policy `build.c++.modules.hide_dependency` (default to false)  that enable passing requires flags as a file to reduce build log noise when compiling with -v (only impact msvc and clang as gcc already use a mapper file that do the same thing)
and make -D print the mapping like for gcc

before ---------
-v:
{7174ADBE-76EC-453D-8B99-7F6B88D9F1D1}

-D
{C376C195-4DE8-401E-936A-9D5894A84237}

-vD:
{856C06DC-C87E-429F-B443-238F7340DBE4}

after ---------
-v:
{D81D94C5-6E18-4F62-B4FD-02C7EF1E4228}

-D
{0823A46A-FFD7-4400-AA20-4B6DB92B5A2D}

-vD:
{89A60786-6361-48F0-AB9F-B2198AC6808A}

### DIFF
--- a/tests/projects/c++/modules/hide_dependency_flags/src/main.cpp
+++ b/tests/projects/c++/modules/hide_dependency_flags/src/main.cpp
@@ -1,0 +1,10 @@
+#include <stdio.h>
+
+import foo;
+
+int main() {
+    printf("add(1, 2): %d\n", foo::add(1, 2));
+    printf("sub(1, 2): %d\n", foo::sub(1, 2));
+    return 0;
+}
+
--- a/tests/projects/c++/modules/hide_dependency_flags/src/main.cpp
+++ b/tests/projects/c++/modules/hide_dependency_flags/src/main.cpp
@@ -2,7 +2,7 @@
 
 import foo;
 
-int main() {
+int main(int argc, char** argv) {
     printf("add(1, 2): %d\n", foo::add(1, 2));
     printf("sub(1, 2): %d\n", foo::sub(1, 2));
     return 0;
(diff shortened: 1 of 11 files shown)


waruqi:
Does it ready now?

Arthapz:
it is, i was justs waiting for @MacroModel to know if it's what he want

xmake/rules/c++/modules/clang/builder.lua
```diff
@@ -194,15 +212,28 @@ end
...
-    for _, flag in ipairs(requiresflags) do
-        -- we need to wrap flag to support flag with space
-        if type(flag) == "string" and flag:find(" ", 1, true) then
-            table.insert(cxxflags, {flag})
+    local hide_dependencies = target:policy("build.c++.modules.hide_dependencies")
+    if #requiresflags> 0 then
+        for _, flag in ipairs(requiresflags) do
+            -- we need to wrap flag to support flag with space
+            if type(flag) == "string" and flag:find(" ", 1, true) and not hide_dependencies then
+                    table.insert(cxxflags, {flag})
```
waruqi: incorrect indent

tests/projects/c++/modules/hide_dependency_flags/src/main.cpp
```diff
@@ -0,0 +1,10 @@
+#include <stdio.h>
+
+import foo;
+
+int main() {
```
waruqi:
```c
int main(int argc, char** argv) {
}
```

PR #6796, opened 2025-09-11
feat(wdk rules) enable support of incomplete wdk

Arthapz:
this PR allow the user setting the path of WDK directory and only assert for the missing binaries when  they are needed (for exemple nuget WDK doesn't ship all of  these binary)

### DIFF
--- a/xmake/rules/wdk/inf/xmake.lua
+++ b/xmake/rules/wdk/inf/xmake.lua
@@ -41,7 +41,6 @@ rule("wdk.inf")
 
         -- get stampinf
         local stampinf = path.join(wdk.bindir, wdk.sdkver, arch, is_host("windows") and "stampinf.exe" or "stampinf")
-        assert(stampinf and os.isexec(stampinf), "stampinf not found!")
 
         -- save uic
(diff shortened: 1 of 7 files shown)


waruqi:
> xmake don't currently have a way to set the WDK rootdir

why do not use `xmake f --wdk=rootdir`? https://github.com/xmake-io/xmake/blob/e641d139eef96d83222719a5f4f349edd87f1c9a/xmake/modules/detect/sdks/find_wdk.lua#L162

Using set_values ​​to set hard-coded paths is not recommended.

and we can also use `set_config("wdk", "rootdir")` to set it.

Arthapz:
> > xmake don't currently have a way to set the WDK rootdir
> 
> why do not use `xmake f --wdk=rootdir`?
> 
> https://github.com/xmake-io/xmake/blob/e641d139eef96d83222719a5f4f349edd87f1c9a/xmake/modules/detect/sdks/find_wdk.lua#L162
> 
> Using set_values ​​to set hard-coded paths is not recommended.
> 
> and we can also use `set_config("wdk", "rootdir")` to set it.

my bad didn't saw it :D, i removed bits about wdk directory

PR #6801, opened 2025-09-13
Change compilers priority for gcc and clang

bugsnotabunny:
Before this change xmake preferred to use clang and gcc (instead of clang++ and g++ respectively) for c++ code which caused build to fail with clang in my current setup

https://github.com/xmake-io/xmake/issues/6803


waruqi:
If the build fails, please open an issue and provide error logs first.

waruqi:
and llvm, mingw toolchain ...

bugsnotabunny:
@waruqi review pls.

Fixed clang toolchain as you have asked. But I was unable to find where the error is in mingw toolchain. It's contents seem okay to me. Am missing something?

waruqi:
> @waruqi review pls.
> 
> Fixed clang toolchain as you have asked. But I was unable to find where the error is in mingw toolchain. It's contents seem okay to me. Am missing something?

mingw has been modified. we need not modify it.

PR #6818, opened 2025-09-16
Fix Nix Package Detection in nix-shell Environment

ZZBaron:
Problem
The Nix package finder in xmake was incorrectly identifying packages within nix-shell environments, which led to linking failures.  The finder would match the first store path containing a /lib directory, regardless of whether that path actually included the requested package. As a result, it produced false positives and failed to locate the correct libraries. Additionally, there was no validation step to confirm that the matched store path contained the necessary libraries for the intended package and environment variables such as NIX_LDFLAGS and NIX_CFLAGS_COMPILE were not being adequately parsed to extract valid Nix store paths.

Fix
This update introduces a more robust method for resolving packages in nix-shell and flake-based development environments. It improves how the package finder scans and validates paths, ensuring it correctly identifies the presence of actual library files before resolving a match. It also enhances environment parsing logic to accurately extract Nix store paths from key environment variables. xmake can now reliably detect and link against dependencies defined via shell.nix or flake.nix.

Testing
The fix was tested using a C++ project that depends on Google Test (gtest) within a nix-shell environment. Both pure and impure shell modes were evaluated. Prior to the fix, the finder incorrectly matched the bash-interactive package solely because it contained a /lib directory, even though it had no relation to gtest. The correct gtest library path, located at /nix/store/1ispb16svjqwv4l2amy4jfygy5k8pkdr-gtest-1.17.0/lib, was present in NIX_LDFLAGS but was previously ignored due to the flawed matching logic. With the updated implementation, xmake now correctly identifies and uses the proper path, resolving the linking issue.

### DIFF
--- a/xmake/modules/package/manager/nix/find_package.lua
+++ b/xmake/modules/package/manager/nix/find_package.lua
@@ -24,13 +24,114 @@ import("lib.detect.find_tool")
 import("private.core.base.is_cross")
 import("package.manager.pkgconfig.find_package", {alias = "find_package_from_pkgconfig"})
 
+-- check if we're in a nix-shell environment
+function _in_nix_shell()
+    local in_nix_shell = os.getenv("IN_NIX_SHELL")
+    return in_nix_shell == "pure" or in_nix_shell == "impure"
+end
+
+-- extract store paths from nix environment variables with better filtering
+function _extract_nix_store_paths(env_var_name, env_var_value)
+    local paths = {}
+    local seen = {}
+    
+    if env_var_value == "" then
+        return paths
+    end
+    
+    -- Handle different environment variable formats
+    local separators = {
+        PATH = ":",
+        PKG_CONFIG_PATH = ":",
+        LIBRARY_PATH = ":",
+        LD_LIBRARY_PATH = ":",
+        C_INCLUDE_PATH = ":",
+        CPLUS_INCLUDE_PATH = ":",
+        NIX_LDFLAGS = "%s",  -- space separated, may contain -L flags
+        NIX_CFLAGS_COMPILE = "%s"  -- space separated, may contain -I flags
+    }
+    
+    local separator = separators[env_var_name] or ":"
+    local pattern = separator == ":" and "[^:]+" or "[^%s]+"
+    
+    for item in env_var_value:gmatch(pattern) do
+        local clean_item = item
+        
+        -- Remove flag prefixes for compiler/linker flags
+        if env_var_name == "NIX_LDFLAGS" then
+            clean_item = item:gsub("^%-L", "")
+        elseif env_var_name == "NIX_CFLAGS_COMPILE" then
+            clean_item = item:gsub("^%-[iI]system%s*", ""):gsub("^%-I", "")
+        end
+        
+        if clean_item:startswith("/nix/store/") then
+            local store_path = clean_item:match("(/nix/store/[^/]+)")
+            if store_path and not seen[store_path] then
+                seen[store_path] = true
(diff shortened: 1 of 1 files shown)


xmake/modules/package/manager/nix/find_package.lua
```diff
@@ -24,13 +24,114 @@ import("lib.detect.find_tool")
 import("private.core.base.is_cross")
 import("package.manager.pkgconfig.find_package", {alias = "find_package_from_pkgconfig"})
 
+-- check if we're in a nix-shell environment
+function _in_nix_shell()
+    local in_nix_shell = os.getenv("IN_NIX_SHELL")
```
waruqi, line 29: you can use os.shell and improve here to detect nix shell

https://github.com/xmake-io/xmake/blob/d4da0944ead398d2e7eba985928adeba197557b7/xmake/core/base/tty.lua#L153

xmake/modules/package/manager/nix/find_package.lua
```diff
@@ -24,13 +24,114 @@ import("lib.detect.find_tool")
...
+-- get current shell buildInputs from environment with improved detection
+function _get_shell_build_inputs()
+    local all_paths = {}
+    local seen = {}
+    
+    if not _in_nix_shell() then
+        return all_paths
+    end
+    
+
```
waruqi: remove empty line

xmake/modules/package/manager/nix/find_package.lua
```diff
@@ -81,16 +183,64 @@ function _get_available_nix_paths()
...
+-- check if a store path actually contains the requested package
+function _validate_package_in_store_path(store_path, name)
+    
+    -- Check if the store path name contains the package name
+    local store_name = path.basename(store_path):lower()
+    local search_name = name:lower()
+    
+    -- Look for exact match, or package name in the store path
+    local name_match = store_name:find(search_name, 1, true) or 
+                      store_name:find(search_name:gsub("%-", "%%-")) -- handle hyphens
```
waruqi: gsub will return two values, `content, count`.

count will be passed find too. it's incorrect. `find(content, count)`

you need add a `()` . e.g. `store_name:find((search_name:gsub("%-", "%%-")))`

or use `local content = search_name:gsub("%-", "%%-")` to ignore count value.

xmake/modules/package/manager/nix/find_package.lua
```diff
@@ -118,8 +268,10 @@ function _find_in_store_path(store_path, name)
                            filename:match("^lib(.+)%.dylib")
             
             if linkname then
-                table.insert(result.links, linkname)
-                table.insert(result.libfiles, libfile)
+                if linkname == name or linkname:find(name) then
```
waruqi: `linkname:find(name, 1, true)`

xmake/modules/package/manager/nix/find_package.lua
ZZBaron: Sorry if I misunderstood, but in this case I don't believe it makes sense to add a "nix-shell" entry to the tty.shell() function because nix-shell is really a shell environment not a new shell like fish. You can use bash, sh, fish, zsh, etc. as the shell and still be inside a nix-shell (environment).

xmake/modules/package/manager/nix/find_package.lua
ZZBaron: ah thanks

xmake/modules/package/manager/nix/find_package.lua
waruqi: got it

Issue #6820, opened 2025-09-17
XMAKE_BINARY_REPO not effective in docker


qiuwei:
### Xmake 版本

v3.0.2+HEAD.993b5c3

### 操作系统版本和架构

ubuntu 22.04 inside docker

### 描述问题

XMAKE_BINARY_REPO not effective in docker. xmake still tries to pull from github. it works on bare metal ubuntu 22.04 though

XMAKE_BINARY_REPO=https://example.com/build-artifacts.git xrepo update-repo -vvv                                                                                                   
/home/rnd/.local/bin/xmake repo --update -v
updating repositories ..
cloning repository(build-artifacts): https://github.com/xmake-mirror/build-artifacts.git to /home/rnd/.xmake/repositories/build-artifacts ..
Cloning into '/home/rnd/.xmake/repositories/build-artifacts'...

### 期待的结果

should work in docker.

### 工程配置

None

### 附加信息和错误日志

XMAKE_BINARY_REPO=https://example.com/build-artifacts.git xrepo update-repo -vvv                                                                                                   
/home/rnd/.local/bin/xmake repo --update -v
updating repositories ..
cloning repository(build-artifacts): https://github.com/xmake-mirror/build-artifacts.git to /home/rnd/.xmake/repositories/build-artifacts ..
Cloning into '/home/rnd/.xmake/repositories/build-artifacts'...

qiuwei:
I trace into the repository.lua and find out there is some outdated cache in /tmp/.xmake1000/250917/xrepo/working/.xmake/linux/x86_64/cache/repository.

how is this cache supposed to be cleared using built-in command?
I have tried
xmake g -c
but it does not work

waruqi:
https://github.com/xmake-io/xmake/pull/6822

PR #6822, opened 2025-09-17
Prioritize envs over repo cache

Shiffted:
#6820


Shiffted:
Alternatively we can could clear the `repository` cache when running `xmake f -c`/`xmake f --check` etc.

PR #6841, opened 2025-09-21
Fix clang-xx toolchain detect

star-hengxing:
https://github.com/xmake-io/xmake/pull/6824#issuecomment-3316027626


waruqi:
额，我刚已经直接 commit 修了

PR #6847, opened 2025-09-23
get.sh: support dnf

black-desk:
Signed-off-by: Chen Linxuan <me@black-desk.cn>


waruqi:
Does it ready now?

black-desk:
> Does it ready now?

It seems to work on my machine.

But I am not sure about whether we actually install all dependencies we need here or not.

waruqi:
> > Does it ready now?
> 
> It seems to work on my machine.
> 
> But I am not sure about whether we actually install all dependencies we need here or not.

git, @development-tools is required, readline-devel is optional and is only needed for REPL, and bzip2 is also optional and is needed for package management.

Issue #6842, opened 2025-09-21
Introduce diagnostic for calling `set_kind()` in `package()`


sveljko:
### Xmake Version

3.0.2

### Operating System Version and Architecture

AlmaLinux 9

### Describe Bug

In my own `xmake-repo`, I introduced a simple package definition for a small static library. Because of copy-paste, I left `set_kind("static")` in it, like:
```lua
package("mylib")
    set_kind("static")
    set_urls("<mylibURL>")
    on_install(function (pkg)
        import("package.tools.xmake").install(pkg, {})
    end)
```
The static library itself is very simple, nothing interesting to report.
```lua
add_rules("mode.debug", "mode.release")
set_languages("gnuxx20")
target("mylib")
    set_kind("static")
    add_files("source/*.cpp")
    add_includedirs("include", {public = true})
    add_headerfiles("include/*.hpp")
 ```

When building my project which `add_requires("mylib", {configs={"local"}})` and for the target `add_packages("mylib", ...<other libs>)`, the package installed fine and, with `-v` verbose output, I saw that the table for the installed library contains `sysincludedirs`. But the cache (`.xmake/linux/x86_64/cache/package`) did not have `sysincludedirs` for the `mylib` package, thus the build failed.

It took a while for me to notice the `set_kind()` "inside" `package()` and when I removed it, it worked fine.

### Expected Behavior

Obviously the `set_kind()` "inside" `package()` wreaks havoc. Now, maybe some of this is intended behavior, but even if it is, it's strange. So, either:
- report an error when calling `set_kind()` "inside" `package()`
- report a warning about it having potentially unwanted effect, with some way of suppressing said warning for those who are fine with it

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")
add_requires("mylib main", {configs = {"local"}})
set_languages "gnuxx20"
target "myapp"
    set_kind "binary"
    add_files(<stuff>)
    add_packages("mylib", ...<other packages>)
```

### Additional Information and Error Logs

nothing to add

star-hengxing:
package:set_kind values is library/binary/toolchain
Please reference
https://github.com/xmake-io/xmake-repo/blob/43c6b0306ee6fe70f53b3749d85952fe1a69a46e/packages/y/yyjson/xmake.lua#L47

sveljko:
[...]

PR #6850, opened 2025-09-23
add package api checking

Shiffted:
https://github.com/xmake-io/xmake/issues/6842

### DIFF
--- a/xmake/modules/private/check/checker.lua
+++ b/xmake/modules/private/check/checker.lua
@@ -26,6 +26,8 @@ function checkers()
     local checkers = _g._CHECKERS
     if not checkers then
         checkers = {
+            -- package api checkers
+            ["api.package.kind"]         = {description = "Check kind configuration in package.", load = true},
             -- target api checkers
             ["api.target.version"]       = {description = "Check version configuration in target."},
             ["api.target.kind"]          = {description = "Check kind configuration in target.", build = true},
--- a/xmake/modules/private/check/checkers/api/package/kind.lua
+++ b/xmake/modules/private/check/checkers/api/package/kind.lua
@@ -1,0 +1,42 @@
+--!A cross-platform build utility based on Lua
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- Copyright (C) 2015-present, Xmake Open Source Community.
+--
+-- @author      Shiffted
+-- @file        kind.lua
+--
+
+import(".api_checker")
+
+function main(opt)
+    opt = opt or {}
+    api_checker.check_packages("kind", table.join(opt, {check = function(package, value)
+        if value == "library" then
+            local extraconf = package:extraconf("kind", "library")
+            if extraconf then
+                if extraconf.headeronly and extraconf.moduleonly then
+                    return false, "a library package cannot be set as both 'headeronly' and 'moduleonly'"
+                end
+                for key, _ in pairs(extraconf) do
+                    if key ~= "headeronly" and key ~= "moduleonly" then
+                        return false, string.format("unknown kind configuration '%s'", key)
+                    end
+                end
+            end
+            return true
+        end
+        return value == "binary" or value == "toolchain" or value == "template"
+    end}))
+end


xmake/plugins/check/main.lua
```diff
@@ -73,6 +75,12 @@ function _check(group_or_name, arguments)
     end
     assert(#checked_checkers > 0, "checker(%s) not found!", group_or_name)
 
+    local needs_packages = group_or_name == "api" or group_or_name:startswith("api.package")
+    if needs_packages then
+        arguments = arguments or {}
+        arguments.packages = package.load_packages(get_requires(), {check_api = false})
+    end
+
```
Shiffted: Should this be handled in the api checker with a cached function?

xmake/modules/private/action/require/impl/check_api.lua
```diff
@@ -0,0 +1,32 @@
...
+-- Copyright (C) 2015-present, Xmake Open Source Community.
+--
+-- @author      Shiffted
+-- @file        check_api.lua
+--
+
+import("private.check.checker")
+import("private.check.show")
+
+function main(package)
```
waruqi: package -> instance

xmake/plugins/check/main.lua
waruqi: Please do not modify this file, it's generic logic.

waruqi: you should load and packages in check_packages()

waruqi: In addition, we should only use project.packages() to detect the package configuration defined in the project xmake.lua. We should not check the package configuration in xmake-repo here

We only use opt.package in package.load to detect packages in xmake-repo.

xmake/modules/private/check/checkers/api/api_checker.lua
```diff
@@ -163,11 +163,21 @@ function check_targets(apiname, opt)
...
 end
+
+-- check api configuration in targets
+function check_targets(apiname, opt)
+    check_instances(apiname, project.targets, opt)
+end
+
+-- check api configuration in packages
+function check_packages(apiname, opt)
+    check_instances(apiname, function() return opt and opt.packages end, opt)
```
waruqi: like project.targets, we only check project.packages

waruqi: We only use opt.package in package.load to detect packages in xmake-repo.

xmake/modules/private/action/require/impl/package.lua
```diff
@@ -1167,7 +1172,8 @@ function _load_packages(requires, opt)
                                                         parentinfo = requireinfo,
                                                         nodeps = opt.nodeps,
                                                         resolvedinfo = opt.resolvedinfo,
-                                                        system = false})
+                                                        system = false,
+                                                        check_api = opt.check_api})
```
waruqi: remove this arg

xmake/modules/private/action/require/impl/package.lua
```diff
@@ -1128,6 +1129,10 @@ function _load_package(packagename, requireinfo, opt)
     -- load environments from the manifest to enable the environments of on_install()
     package:envs_load()
 
+    if opt.check_api ~= false then
```
waruqi: remove opt.check_api

xmake/modules/private/check/checker.lua
```diff
@@ -26,6 +26,8 @@ function checkers()
     local checkers = _g._CHECKERS
     if not checkers then
         checkers = {
+            -- package api checkers
+            ["api.package.kind"]         = {description = "Check kind configuration in package.", package_load = true},
```
waruqi: package_load -> load

xmake/modules/private/check/show.lua
```diff
@@ -0,0 +1,36 @@
...
+-- @author      Shiffted
+-- @file        show.lua
+--
+
+function wshow(str, opt)
+    opt = opt or {}
+    _g.showed = _g.showed or {}
+    local showed = _g.showed
+
+    local message = str or string.format("unknown %s value '%s'", opt.apiname, opt.value)
```
waruqi: The location can be moved, but the implementation should not be modified.

```lua
    if str and opt.sourcetips then
        infostr = string.format("%s${clear}: %s", opt.sourcetips, str)
    elseif opt.sourcetips and opt.apiname and opt.value ~= nil then
        infostr = string.format("%s${clear}: unknown %s value '%s'", opt.sourcetips, opt.apiname, opt.value)
    elseif str then
        infostr = string.format("${clear}: %s", str)
    end
```

is missing

xmake/plugins/check/main.lua
Shiffted: How can we access project.packages() from here? It doesn't get exported in the project module, should I add it?

waruqi: Maybe we can't check project.packages , they are not package instances.

we can only use opt.package to check it in package/load.

Shiffted: Hmm, we could still check some things with them, e.g., kind/versions/versionfiles. Other checkers could check for `opt.package` maybe.

waruqi:
Does it ready now?

Shiffted:
> Does it ready now?

I would still like to add `project.packges`, even if they aren't package instances we can still check some apis with them. Otherwise, I would add another `checkers()` function so the package checkers are not shown in the check plugin.

Shiffted:
@waruqi Can I make it public or is there some other way to access it here?

waruqi:
> I would still like to add , even if they aren't package instances we can still check some apis with them. Otherwise, I would add another function so the package checkers are not shown in the check plugin.`project.packges``checkers()`

core.project.project/required_packages + core.package.package/load_from_project

PR #6890, opened 2025-10-04
Update cmake_importfiles.lua

luadebug:
Despite we fixed issues with cross auto-generated CMakeConfigs, we have acquired new issue with wasm platform.
https://github.com/xmake-io/xmake/pull/6880

https://github.com/xmake-io/xmake-repo/pull/8312 (imgui)

### DIFF
--- a/xmake/modules/target/action/install/cmake_importfiles.lua
+++ b/xmake/modules/target/action/install/cmake_importfiles.lua
@@ -41,8 +41,10 @@ function _get_builtinvars(target, installdir)
     local target_ptrbytes
     if target:is_plat("wasm") then
         target_ptrbytes = is_arch("wasm32") and "4" or "8"
-    else
+    elseif target:is_plat("cross") then
         target_ptrbytes = target:check_sizeof("void*")
+    else
+        target_ptrbytes = target:is_arch64() and "8" or "4"
     end
     return {TARGETNAME      = target:name(),
             PROJECTNAME     = project.name() or target:name(),
--- a/xmake/modules/target/action/install/cmake_importfiles.lua
+++ b/xmake/modules/target/action/install/cmake_importfiles.lua
@@ -39,9 +39,7 @@ end
 -- get the builtin variables
 function _get_builtinvars(target, installdir)
     local target_ptrbytes
-    if target:is_plat("wasm") then
-        target_ptrbytes = target:is_arch("wasm32") and "4" or "8"
-    elseif target:is_plat("cross") then
+    if target:is_plat("cross") then
         target_ptrbytes = target:check_sizeof("void*")
     else
         target_ptrbytes = target:is_arch64() and "8" or "4"
--- a/xmake/modules/target/action/install/cmake_importfiles.lua
+++ b/xmake/modules/target/action/install/cmake_importfiles.lua
@@ -38,12 +38,18 @@ end
 
 -- get the builtin variables
 function _get_builtinvars(target, installdir)
+    local target_ptrbytes
+    if target:is_plat("wasm") then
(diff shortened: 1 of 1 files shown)


xmake/modules/target/action/install/cmake_importfiles.lua
```diff
@@ -38,12 +38,18 @@ end
 
 -- get the builtin variables
 function _get_builtinvars(target, installdir)
+    local target_ptrbytes
+    if target:is_plat("wasm") then
+        target_ptrbytes = target:is_arch("armeabi", "armeabi-v7a", "x86", "mips", "i386", "armv7", "armv7s", "mipsel", "wasm32") and "4" or "8"
+    else
+        target_ptrbytes = target:check_sizeof("void*")
+    end
```
luadebug, line 46: We could use if `target:is_plat("cross") then` and swap `target_ptrbytes = ` but that looks like not very good idea.

xmake/modules/target/action/install/cmake_importfiles.lua
```diff
@@ -38,12 +38,18 @@ end
 
 -- get the builtin variables
 function _get_builtinvars(target, installdir)
+    local target_ptrbytes
+    if target:is_plat("wasm") then
+        target_ptrbytes = target:is_arch("armeabi", "armeabi-v7a", "x86", "mips", "i386", "armv7", "armv7s", "mipsel", "wasm32") and "4" or "8"
```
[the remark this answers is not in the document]
luadebug: Will it work more accurate, especially behind WASM environment?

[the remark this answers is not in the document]
waruqi: It should be better. I think judging arch64 is more general than judging arch32.

The architecture listing commands for arch32 are too cluttered.

xmake/modules/target/action/install/cmake_importfiles.lua
```diff
@@ -38,12 +38,18 @@ end
 
 -- get the builtin variables
 function _get_builtinvars(target, installdir)
+    local target_ptrbytes
+    if target:is_plat("wasm") then
```
waruqi: only for wasm? only wasm32/wasm64 archs exist.

xmake/modules/target/action/install/cmake_importfiles.lua
waruqi: For known platforms, we can detect it by arch.

like this?

```lua
if target:is_plat("cross") then
    target_ptrbytes = target:check_sizeof("void*")
else
    target_ptrbytes = target:is_arch64() and "8" or "4"
end
```

luadebug: Yes, like that maybe is best way to do this?

xmake/modules/target/action/install/cmake_importfiles.lua
luadebug: I expected `target_ptrbytes = target:check_sizeof("void*")` to work for wasm as well, but it looks like did not return nor 4 nor 8 for wasm build, so I placed problematic platforms here.

waruqi: or we should fix check_sizeof to support wasm build

xmake/modules/target/action/install/cmake_importfiles.lua
waruqi: is_arch64 will detect wasm64 arch.

It is equivalent to

```lua
target_ptrbytes = target:is_arch("wasm64") and "8" or "4"
```

=

```lua
target_ptrbytes = target:is_arch64() and "8" or "4"
```

Therefore we don't need to handle wasm archs specially

PR #6899, opened 2025-10-07
improve(toolchain): enhance gcc/clang support for cross-compilation

romeoxbm:
This PR patches the built-in gcc and clang toolchain files to properly support cross-compilation by prepending the appropriate target prefix to the toolset. This avoids the need for custom toolchain definitions in many common scenarios and allows xmake to correctly pick up the expected cross tools.

After implementing this, I noticed that xmake already provides a dedicated `cross` toolchain. However, that implementation seems to mainly support gcc and may not cover more advanced setups (e.g., clang-based cross builds). It would be good to get maintainer input here, as my changes might overlap or partially conflict with that toolchain.

### DIFF
--- a/xmake/toolchains/gcc/xmake.lua
+++ b/xmake/toolchains/gcc/xmake.lua
@@ -47,6 +47,7 @@ toolchain("gcc" .. suffix)
     end)
 
     on_load(function (toolchain)
+        import("core.base.option")
 
         -- add march flags
         local march
@@ -61,6 +62,36 @@ toolchain("gcc" .. suffix)
             toolchain:add("asflags", march)
             toolchain:add("ldflags", march)
             toolchain:add("shflags", march)
+        end
+
+        local host_arch = os.arch()
+        local target_arch = toolchain:arch()
+
+        if host_arch == target_arch then
+            -- Early exit: prevents further configuration of this toolchain
+            return
+        elseif option.get("verbose") then
+            cprint("${bright yellow}cross compiling from %s to %s", host_arch, target_arch)
+        end
+
+        local target
+        if toolchain:is_arch("x86_64", "x64") then
+            target = "x86_64"
+        elseif toolchain:is_arch("i386", "x86", "i686") then
+            target = "i686"
+        elseif toolchain:is_arch("arm64", "aarch64") then
+            target = "aarch64"
+        elseif toolchain:is_arch("arm") then
+            target = "armv7"
+        end
+
+        -- TODO: Add support for more platforms, such as mingw.
+        if target and toolchain:is_plat("linux") then
+            target = target .. "-linux-gnu-"
+            toolchain:set("toolset", "cc", target .. "gcc" .. suffix)
+            toolchain:set("toolset", "cxx", target .. "g++" .. suffix, "gcc" .. suffix)
+            toolchain:set("toolset", "ld", target .. "g++" .. suffix, "gcc" .. suffix)
+            toolchain:set("toolset", "sh", target .. "g++" .. suffix, "gcc" .. suffix)
         end
     end)
 end
--- a/xmake/toolchains/clang/load.lua
+++ b/xmake/toolchains/clang/load.lua
@@ -51,13 +51,46 @@ function _add_vsenv(toolchain, name, curenvs)
 end
 
 function main(toolchain, suffix)
+    import("core.base.option")
+    import("core.project.project")
+
+    if project.policy("build.optimization.lto") then
+        toolchain:set("toolset", "ar",  "llvm-ar" .. suffix)
+        toolchain:set("toolset", "ranlib",  "llvm-ranlib" .. suffix)
+    end
+
+    local march
+    if toolchain:is_arch("x86_64", "x64", "arm64") then
+        march = "-m64"
+    elseif toolchain:is_arch("i386", "x86", "i686") then
+        march = "-m32"
+    end
+    if march then
+        toolchain:add("cxflags", march)
+        toolchain:add("mxflags", march)
+        toolchain:add("asflags", march)
+        toolchain:add("ldflags", march)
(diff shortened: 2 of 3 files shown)


xmake/toolchains/gcc/xmake.lua
```diff
@@ -62,6 +63,36 @@ toolchain("gcc" .. suffix)
...
+            target = "i686"
+        elseif toolchain:is_arch("arm64", "aarch64") then
+            target = "aarch64"
+        elseif toolchain:is_arch("arm") then
+            target = "armv7"
+        end
+
+        -- TODO: Add support for more platforms, such as mingw.
+        if target and toolchain:is_plat("linux") then
+            target = target .. "-linux-gnu-"
```
waruqi, line 90: The gcc toolchain is only designed to support the host platform. Please use the cross toolchain.

it will break many things. e.g. packages, toolchain cross state, ...

https://github.com/xmake-io/xmake/blob/888f5c3b5452b2b8c6c47e89028c4449a238e31f/xmake/modules/private/utils/toolchain.lua#L28

https://github.com/xmake-io/xmake/blob/888f5c3b5452b2b8c6c47e89028c4449a238e31f/xmake/modules/package/tools/cmake.lua#L99

waruqi:
The clang/gcc toolchain is only designed to support the host platform. Please use the cross toolchain.

it will break many things. e.g. packages, toolchain cross state, ...

https://github.com/xmake-io/xmake/blob/888f5c3b5452b2b8c6c47e89028c4449a238e31f/xmake/modules/private/utils/toolchain.lua#L28

https://github.com/xmake-io/xmake/blob/888f5c3b5452b2b8c6c47e89028c4449a238e31f/xmake/modules/package/tools/cmake.lua#L99

and other places.

waruqi:
In addition, the cross toolchain is bound to the cross platform (xmake f -p cross), and many cross-compilation logics rely on the judgment of `if is_plat("cross")`.

e.g. install all packages (cross-compilation)

So, please specify `xmake f -p cross` to do cross-compilation.

romeoxbm:
Thanks for the clarification!

The idea that something might be amiss with the current cross-compilation behavior originated from examining the Clang toolchain. In the `load.lua` file, I noticed the use of the `--target` option, but that file was only imported on Windows and MinGW. That seemed unusual to me, which is why I initially thought extending that logic more generally might be the right approach.

That said, I completely understand your point about the cross toolchain being the intended way to handle these cases, and why modifying the built-in gcc/clang toolchains could break existing assumptions around `is_plat("cross")` and package/toolchain states.

One thing I did notice, though — unless I’m missing something — is that the cross toolchain seems to rely primarily on GCC. In my case, and I suspect for others as well, using GCC is not always an option since I also need to cross-compile with other compilers such as Clang or MSVC.

I’ll take a closer look at the cross toolchain implementation and the files you pointed out to better understand how xmake handles cross-compilation internally. I definitely don’t want to push changes that could disrupt existing workflows.

waruqi:
If you use gcc to cross-compile, you can use the cross toolchain. No need to modify the gcc toolchain.

msvc also supports cross-compilation, just need to modify the arch. e.g. build arm64 on x86_64

If you want to improve the clang toolchain to support building other arch products, you can do so, but the premise is that you don't break the existing logic. It may need to be supported in many other places, such as the ones I just mentioned.

Cross toolchain currently mainly supports gcc, because most embedded cross-compilation toolchain SDKs are based on gcc

In addition, mingw toolchain is also cross-compiled, but it is handled by mingw platform.

waruqi:
Therefore, I think you should just improve the clang toolchain and modify some of the clang handling in package.tools and so on.

Perhaps we also need to improve toolchain:is_cross()

https://github.com/xmake-io/xmake/blob/dev/xmake/core/base/private/is_cross.lua

waruqi:
And llvm toolchain. https://github.com/xmake-io/xmake/blob/dev/xmake/toolchains/llvm/xmake.lua

- clang toolchain:  clang on system, e.g. /usr/bin/clang, /usr/local/bin/clang-17, clang-18, clang-xx with system ar.
- llvm toolchain: full llvm tools, use llvmsdk/bin/clang + llvmsdk/bin/llvm-ar (first) , we need `xmake f --sdk=/llvmsdk` to detect it.

Therefore, I think if we want to make clang support cross-compilation, it might be better to improve the llvm toolchain to add other platform targets to support it instead of clang toolchain.

```console
xmake f --toolchain=llvm -arch xxx --sdk=/xxx
```

romeoxbm:
Thanks for the detailed clarification — it's very helpful to understand the current expectations around cross-compilation support.
In my case, I need cross-compilation to work with MSVC, GCC, and Clang. What feels confusing to me is that the way xmake handles cross-compilation seems to differ quite a bit depending on the toolchain:

* With MSVC, simply setting the architecture is enough. For example (assuming I'm building on x64):
```
xmake f -p windows -a arm64 --toolchain=msvc
```
works out of the box without needing to use the `cross` platform or toolchain.

* For GCC, from what I understand, the expected approach is to explicitly specify `-p cross --toolchain=cross`.

* For Clang, it seems to require yet another approach, which adds to the feeling of inconsistency. Each compiler currently needs a different set of command-line options to achieve what is conceptually the same goal (cross-compilation), which makes the UX harder to reason about.

I also have some doubts about the role of the `cross` platform and toolchain in this whole picture, since their usage doesn’t seem uniform across toolchains.

Regarding improvements: I'll definitely take a closer look at the LLVM toolchain, as well as the other places you pointed out, to better understand how xmake currently handles cross-compilation internally before making any further changes.

My main goal here would be to help make cross-compilation behavior more consistent across toolchains, so that specifying `-a <arch>` would ideally work similarly whether the compiler is MSVC, Clang, or GCC (or any other supported).

PR #6946, opened 2025-10-20
Enhance Clang toolchain with LLD linker support

star-hengxing:
Fix package:on_test when enable lto

### DIFF
--- a/xmake/toolchains/clang/load.lua
+++ b/xmake/toolchains/clang/load.lua
@@ -81,12 +81,7 @@ function main(toolchain, suffix)
             end
         end
 
-        if project.policy("build.optimization.lto") then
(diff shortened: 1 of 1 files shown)


xmake/toolchains/clang/load.lua
```diff
@@ -79,6 +81,11 @@ function main(toolchain, suffix)
             end
         end
 
+        if project.policy("build.optimization.lto") then
+            toolchain:add("ldflags", "-fuse-ld=lld-link")
+            toolchain:add("shflags", "-fuse-ld=lld-link")
+        end
+
         if is_host("linux") then
```
waruqi:
```lua
if is_host("linux") or project.policy("build.optimization.lto") then
```

PR #7036, opened 2025-11-18
Fix debug logging in cmake find_package

MeanSquaredError:
Currently the cmake find_package function uses `cprint()` to display its debug output with different color (dimmed). Internally `cprint()` calls `colors.translate()`, which interprets the various `${color}` tags in the printed text. However these color tags look exactly like the CMake variable expansion tags `${VARIABLE}`, which causes `cprint()`, to print `VARIABLE` instead of the correct `${VARIABLE}`.

This is an example of the current (incorrect) debug output
```
finding it from the generated CMakeLists.txt:
cmake_minimum_required(VERSION 4.2.0-rc1)
project(find_package)
find_package(Boost REQUIRED )
if(Boost_FOUND)
   add_executable(test_Boost test.cpp)
   target_include_directories(test_Boost PRIVATE Boost_INCLUDE_DIR Boost_INCLUDE_DIRS BOOST_INCLUDE_DIR BOOST_INCLUDE_DIRS)
   target_include_directories(test_Boost PRIVATE Boost_CXX_INCLUDE_DIRS)
   target_link_libraries(test_Boost PRIVATE Boost::headers)
endif(Boost_FOUND)
```

The correct one should be
```
finding it from the generated CMakeLists.txt:
cmake_minimum_required(VERSION 4.2.0-rc1)
project(find_package)
find_package(Boost REQUIRED )
if(Boost_FOUND)
   add_executable(test_Boost test.cpp)
   target_include_directories(test_Boost PRIVATE ${Boost_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${BOOST_INCLUDE_DIR} ${BOOST_INCLUDE_DIRS})
   target_include_directories(test_Boost PRIVATE ${Boost_CXX_INCLUDE_DIRS})
   target_link_libraries(test_Boost PRIVATE Boost::headers)
endif(Boost_FOUND)
```

This PR replaces calls to `cprint()` with `print()` in the cmake `find_package()` function. Ideally there would be a way to print the debug output in dimmed color while not interpreting any color tags after the initial `{dim}` tag. However the `colors` module that does seem to allow that so the only feasible approach at the moment is just to not interpret any color tags altogether.

### DIFF
--- a/xmake/modules/package/manager/cmake/find_package.lua
+++ b/xmake/modules/package/manager/cmake/find_package.lua
@@ -131,7 +131,8 @@ function _find_package(cmake, name, opt)
     cmakefile:close()
     if option.get("diagnosis") then
         local cmakedata = io.readfile(filepath)
-        cprint("finding it from the generated CMakeLists.txt:\n${dim}%s", cmakedata)
+        cprint("finding it from the generated CMakeLists.txt:")
+        io.write(cmakedata .. "\n")
     end
 
     -- run cmake
@@ -151,7 +152,8 @@ function _find_package(cmake, name, opt)
         local flagsdata = io.readfile(flagsfile)
         if flagsdata then
             if option.get("diagnosis") then
-                cprint("finding includes from %s\n${dim}%s", flagsfile, flagsdata)
+                cprint("finding includes from %s", flagsfile)
+                io.write(flagsdata .. "\n")
             end
             for _, line in ipairs(flagsdata:split("\n", {plain = true})) do
                 if line:find("CXX_INCLUDES =", 1, true) then
@@ -191,7 +193,8 @@ function _find_package(cmake, name, opt)
         local linkdata = io.readfile(linkfile)
         if linkdata then
             if option.get("diagnosis") then
-                cprint("finding links from %s\n${dim}%s", linkfile, linkdata)
+                cprint("finding links from %s", linkfile)
+                io.write(linkdata .. "\n")
             end
             for _, line in ipairs(os.argv(linkdata)) do
                 local is_ldflags = false


MeanSquaredError:
Converted it to draft in order to apply the changes suggested by gemini-code-assist.

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -131,7 +131,7 @@ function _find_package(cmake, name, opt)
     cmakefile:close()
     if option.get("diagnosis") then
         local cmakedata = io.readfile(filepath)
-        cprint("finding it from the generated CMakeLists.txt:\n${dim}%s", cmakedata)
+        print("finding it from the generated CMakeLists.txt:\n%s", cmakedata)
```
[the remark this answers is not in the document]
MeanSquaredError: Splitting the output into multiple calls to io.write() doesn't work because any color tags like `${dim}` only work during for the duration of the same call to `io.write()` and have no effect on the next call to `io.write()`.

Additionally the call to `io.write()` skips any file logging additionally performed by `cprint()` or `print()`

waruqi: print will also parse `${var}`. you can try

```lua
cprint("finding it from the generated CMakeLists.txt:")
io.write(cmakedata .. "\n")

Comment (MeanSquaredError) [reply xmake/modules/package/manager/cmake/find_package.lua]: I tried the modified version, which I submitted as a PR and it prints the debug information without modifying CMake `${VARIABLES}` expansions.

Looking at the `cprint()` definition it looks like:
```
-- print format string and colors with newline
function utils.cprint(format, ...)
    assert(format)
    local message = string.tryformat(format, ...)
    utils._print(colors.translate(message))
    if log:file() then
        log:printv(colors.ignore(message))
    end
end
```
It is the call to `colors.translate()` which modifies the CMake variable expansions.

On the other hand the `print()` definition is
```

-- print format string with newline
function utils.print(format, ...)
    assert(format)
    local message = string.tryformat(format, ...)
    utils._print(message)
    log:printv(message)
end
```
It doesn't call `colors.translate()`, so I that's why it doesn't modify the variable expansions (or at least it doesn't modify them on my test machine).

So it seems to me that calling `print()` is safe. What do you think?

MeanSquaredError:
I don't see an easy way to improve the PR. The solution proposed by the AI assistant doesn't work. So I am removing the draft status and making the PR ready for review and suggestions.

xmake/modules/package/manager/cmake/find_package.lua
MeanSquaredError: BTW `cprint()` and `print()` write to stdout, but also optionally write a copy to a debug logfile (or at least by looking at the code it seems that they can optionally write to a debug logfile).

`io.write()` on the other hand only writes to stdout and doesn't write a copy to the debug logfile. So that's why I wanted to avoid using io.write directly and would like to use `print()` instead if possible.

waruqi: vformat will parse `${var}`

https://github.com/xmake-io/xmake/blob/f28284fbeb96285ae796bd2389e94d79a70d9c25/xmake/core/sandbox/modules/utils.lua#L66

MeanSquaredError: You are right that `print()` also does `${var}` expansion.

I assumed incorrectly that find_package uses `utils.print()` instead of `sandbox_utils.print()` and that's why I thought that `print()` it doesn't expand variables, while in fact it does.

I changed the code to use `io.write()` as per your suggestion. It works correctly now, except that it doesn't do the optional logging to a file. So now the PR is ready for review again.

On a side note: I think that sandbox_utils needs another version of `print()` that writes strings to stdout and does debug logging just like the current `print()` does, but the new function wouldn't expand any variables . While useful such a new function is outside of the scope of this PR though, so I have not added it.

PR #7044, opened 2025-11-20
Fix fallback scanner creating circular dependencies out of thin air.

RoqueDeicide:
When using `c++` or `cpp` extension for C++ module interface unit files, fallback scanner was mistakenly treating them as implementation units instead, creating circular dependencies in the `dep` file.

I've changed the scanner to detect implementation units by checking the following conditions:
 1. The file doesn't contain `export module`. This excludes module and partition interface units.
 2. The name of the module doesn't include a `:`. This excludes partition implementation units.

This behavior is consistent with clang's and cl's dependency scanners.


Arthapz:
Lgtm

PR #7046, opened 2025-11-20
Optionally allow CMake system packages without include directories or linked libraries

MeanSquaredError:
Currently the CMake find_package module only accepts packages that have at least one include directory or linked library. This causes problems when trying to find header-only libraries that are installed under one of the Linux system include directories (e.g. in `/usr/include/` or `/usr/local/include/`.

For example adding the following line to xmake.lua will cause the build to fail:
```
add_requires ("cmake::Boost", {system = true, configs = {link_libraries = {"Boost::headers"}}})
```

The reason for this failure is that the `Boost::headers` target adds the header-only part of the Boost library. 

It doesn't add any linked libraries because it is header-only, but it actually adds one include directory pointing either to `/usr/include` or to `/usr/local/include`, depending on where Boost is installed. However that directory (`/usr/include` or `/usr/local/include`) is already configured in GCC as a system include directory, so CMake doesn't add it to the generated file `.../CMakeFiles/test_Boost.dir/flags.make` and when xmake's cmake module parses that file, it sees no include directories at all.

That's fine (we don't want to explicitly add include directories that are already part of the compiles system include path), except that now there are no includes, no linked libraries for the `Boost::headers` target and the `add_requires()` call fails.

So this PR adds a boolean configuration option called `allow_empty_package`, which allows finding CMake system packages that don't add any includes or linked libraries. After the PR is applied, the following call will succeed
```
add_requires ("cmake::Boost", {system = true, configs = {allow_empty_package = true, link_libraries = {"Boost::headers"}}})
```

I am not sure whether `allow_empty_package` is the best possible name for that option, so I can change it if anyone comes up with a better name.

### DIFF
--- a/xmake/modules/package/manager/cmake/find_package.lua
+++ b/xmake/modules/package/manager/cmake/find_package.lua
@@ -104,8 +104,7 @@ function _find_package(cmake, name, opt)
     end
     local testname = "test_" .. name
     cmakefile:print("find_package(%s REQUIRED %s)", requirestr, componentstr)
-    cmakefile:print("if(%s_FOUND)", name)
-    cmakefile:print("   add_executable(%s test.cpp)", testname)
+    cmakefile:print("add_executable(%s test.cpp)", testname)
     -- setup include directories
     local includedirs = ""
     if configs.include_directories then
@@ -114,9 +113,9 @@ function _find_package(cmake, name, opt)
         includedirs = ("${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS}"):format(name, name)
         includedirs = includedirs .. (" ${%s_INCLUDE_DIR} ${%s_INCLUDE_DIRS}"):format(name:upper(), name:upper())
     end
-    cmakefile:print("   target_include_directories(%s PRIVATE %s)", testname, includedirs)
+    cmakefile:print("target_include_directories(%s PRIVATE %s)", testname, includedirs)
     -- reserved for backword compatibility
-    cmakefile:print("   target_include_directories(%s PRIVATE ${%s_CXX_INCLUDE_DIRS})",
+    cmakefile:print("target_include_directories(%s PRIVATE ${%s_CXX_INCLUDE_DIRS})",
         testname, name)
     -- setup link library/target
     local linklibs = ""
@@ -126,8 +125,7 @@ function _find_package(cmake, name, opt)
         linklibs = ("${%s_LIBRARY} ${%s_LIBRARIES} ${%s_LIBS}"):format(name, name, name)
         linklibs = linklibs .. (" ${%s_LIBRARY} ${%s_LIBRARIES} ${%s_LIBS}"):format(name:upper(), name:upper(), name:upper())
     end
-    cmakefile:print("   target_link_libraries(%s PRIVATE %s)", testname, linklibs)
-    cmakefile:print("endif(%s_FOUND)", name)
+    cmakefile:print("target_link_libraries(%s PRIVATE %s)", testname, linklibs)
     cmakefile:close()
     if option.get("diagnosis") then
         local cmakedata = io.readfile(filepath)


xmake/modules/package/manager/cmake/configurations.lua
```diff
@@ -29,6 +29,7 @@ function main()
         moduledirs          = {description = "Set the cmake modules directories."},
         presets             = {description = "Set the preset values, e.g. {Boost_USE_STATIC_LIB = true}"},
         envs                = {description = "Set the run environments of cmake, e.g. {CMAKE_PREFIX_PATH = \"xxx\"}"},
+        allow_empty_package = {description = "Accept package even if it doesn't have any include directories or linked libraries"},
```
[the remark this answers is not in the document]
MeanSquaredError: Line 31 already has a trailing comma.

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -301,7 +301,7 @@ function _find_package(cmake, name, opt)
     os.tryrm(workdir)
 
     -- get results
-    if links or includedirs then
+    if configs.allow_empty_package or links or includedirs then
```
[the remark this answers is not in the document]
MeanSquaredError: The old code already allows in some cases `links`, `includedirs` and the other variables to be empty, so the nil values in these variables are already handled gracefully.

waruqi:
But you haven't judged whether the package is successfully found when using CMake's find_package. How can we determine if it's an empty package or if the package isn't found at all?

MeanSquaredError:
OK, I will make the PR a draft until I fix that.

MeanSquaredError:
OK, I changed the PR so that now cmake find_package() returns an error when a package is not found even with `allow_empty_package = true`.

While implementing the change, I changed slightly `sandbox_os.runv()` and `sandbox_os.vrunv()`, so that now they accept an `opt.try` parameter that allows them to return an `ok, errors` result, instead of raising an error. The idea for the `opt.try` option was taken from `sandbox_os.execv()`.

waruqi:
Please do not modify the behavior of os.run/vrunv, as these two APIs do not have a return value by convention. You should use try/catch blocks instead.

MeanSquaredError:
OK, I changed the PR. Now it doesn't modify the sandbox_os.runv() and sandbox_os.vrunv() interfaces, but uses a try block instead.

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -138,9 +136,17 @@ function _find_package(cmake, name, opt)
     -- run cmake
     local envs = configs.envs or opt.envs or {}
     envs.CMAKE_BUILD_TYPE = envs.CMAKE_BUILD_TYPE or _cmake_mode(opt.mode or "release")
-    try {function() return os.vrunv(cmake.program, {workdir}, {curdir = workdir, envs = envs}) end}
+    -- If the generated CMakeLists.txt fails to find the REQUIRED package, CMake will exit
+    -- with code 1, os.vrunv will raise an error and the try{} block will return nil.
+    ok = try {function()
```
waruqi: local ok

xmake/modules/package/manager/cmake/find_package.lua
```diff
@@ -138,9 +136,17 @@ function _find_package(cmake, name, opt)
...
+        os.vrunv(cmake.program, {workdir}, {try = true, curdir = workdir, envs = envs})
```
waruqi: remove `try = true`

xmake/modules/package/manager/cmake/find_package.lua
MeanSquaredError: Oops, I somehow missed that. Fixed it.

xmake/modules/package/manager/cmake/find_package.lua
MeanSquaredError: Fixed that too.

PR #7069, opened 2025-11-27
Save QT rules for llvm-toolchain

MaxwellGengYF:
These cxx-flags is only for cl.exe and clang_cl.exe


xmake/rules/qt/load.lua
```diff
@@ -225,8 +225,8 @@ function main(target, opt)
     if qt_sdkver:ge("6.0") then
         -- @see https://github.com/xmake-io/xmake/issues/2071
         if target:is_plat("windows") then
```
waruqi: 直接判断就行了。

```lua
if target:is_plat("windows") and target:has_tool("cxx", "clang_cl", "cl") then
```

PR #7084, opened 2025-12-01
fix pacman find package

choyy:


xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -238,6 +236,11 @@ function main(name, opt)
         if result.links then
             result.links = table.reverse_unique(result.links)
         end
+
+        -- We should get version from pacman, because if a pacman package contains multiples .pc we may get a wrong version
+        local version = try { function() return os.iorunv(pacman.program, {"-Q", name}) end }
```
waruqi: check version is nil. and use semver.match

https://github.com/xmake-io/xmake/blob/25c3526637322cdd36a0bc78b3cd7716dcc62382/xmake/modules/package/manager/nimble/find_package.lua#L40

xmake/modules/package/manager/pacman/find_package.lua
```diff
@@ -238,6 +237,15 @@ function main(name, opt)
         if result.links then
             result.links = table.reverse_unique(result.links)
         end
+
+        -- We should get version from pacman, because if a pacman package contains multiples .pc we may get a wrong version
+        local verstr = try { function() return os.iorunv(pacman.program, {"-Q", name}) end }
+        if verstr then
+            local version = semver.match(verstr)
+            if version then
+                result.version = version:rawstr():split('-')[1]
```
waruqi, line 246: Why do not use version:rawstr()? it's already a valid semver.

choyy: pacman version is `pkgver-pkgrel`, `split('-')[1]` is to remove pkgrel to get the original version.

PR #7104, opened 2025-12-06
Fix prepare for project generators

### LINKED ISSUE #6544: 依赖 C++ modules 的单元测试无法生成 compile_commands
### Xmake 版本

xmake v3.0.0+dev.ace14b3

### 操作系统版本和架构

Ubuntu 20.04

### 描述问题

可复现的例子已上传。

其中 `xmake.lua` 有这么一部分：

```lua
target("test")
do
    set_kind("binary")
    set_default(false)
    add_deps("lib")
    add_files("tests/test.cpp")
    add_tests("default")
end
```

依赖项 `lib` 是 C++ module。

此时运行 `xmake project -k compile_commands build` 会报错。

如果把 `set_default(false)` 去掉则能成功。

如果把 `add_deps("lib")` 去掉，则不论 `set_default` 是 true 或 false 都能成功

### 期待的结果

依赖 C++ modules 时，即使 `set_default(false)` 也能为测试样例生成 compile_commands

### 工程配置

[reproduce.tgz](https://github.com/user-attachments/files/20721461/reproduce.tgz)

### 附加信息和错误日志

```console
❗ error: @programdir/core/main.lua:329: @programdir/rules/c++/modules/scanner.lua:899: no modules!
stack traceback:
    [C]: in function 'error'
    [@programdir/core/base/os.lua:1075]: in function 'raiselevel'
    [@programdir/core/sandbox/modules/utils.lua:143]: in function 'assert'
    [@programdir/rules/c++/modules/scanner.lua:899]: in function 'get_modules'
    [@programdir/rules/c++/modules/gcc/../gcc/../builder.lua:665]:
    [@programdir/modules/private/action/build/target.lua:470]: in function 'run'
    [@programdir/plugins/project/utils/target_cmds.lua:72]: in function 'get_target_buildcmds'
    [@programdir/plugins/project/clang/compile_commands.lua:246]: in function '_add_target_commands'
    [@programdir/plugins/project/clang/compile_commands.lua:274]: in function '_add_target'
    [@programdir/plugins/project/clang/compile_commands.lua:287]: in function '_add_targets'
    [@programdir/plugins/project/clang/compile_commands.lua:312]: in function '?'
    [@programdir/plugins/project/main.lua:75]: in function '_make'
    [@programdir/plugins/project/main.lua:95]:
    [C]: in function 'xpcall'
    [@programdir/core/base/utils.lua:246]:
    [@programdir/core/base/task.lua:504]: in function 'run'
    [@programdir/core/main.lua:327]: in function 'cotask'
    [@programdir/core/base/scheduler.lua:406]:

stack traceback:
        [C]: in function 'error'
        @programdir/core/base/os.lua:1075: in function 'os.raiselevel'
        (...tail calls...)
        @programdir/core/main.lua:329: in upvalue 'cotask'
        @programdir/core/base/scheduler.lua:406: in function <@programdir/core/base/scheduler.lua:399>
```

Shiffted:
fixes #6544

### DIFF
--- a/xmake/rules/c++/precompiled_header/xmake.lua
+++ b/xmake/rules/c++/precompiled_header/xmake.lua
@@ -23,7 +23,9 @@ rule("c.build.pcheader")
         import("private.action.build.pcheader").config(target, "c", opt)
     end)
     before_prepare(function (target, jobgraph, opt)
-        import("private.action.build.pcheader").build(target, jobgraph, "c", opt)
+        if not os.getenv("XMAKE_IN_COMPILE_COMMANDS_PROJECT_GENERATOR") then
+            import("private.action.build.pcheader").build(target, jobgraph, "c", opt)
+        end
     end, {jobgraph = true})
 
 rule("c++.build.pcheader")
@@ -32,6 +34,8 @@ rule("c++.build.pcheader")
         import("private.action.build.pcheader").config(target, "cxx", opt)
     end)
     before_prepare(function (target, jobgraph, opt)
-        import("private.action.build.pcheader").build(target, jobgraph, "cxx", opt)
+        if not os.getenv("XMAKE_IN_COMPILE_COMMANDS_PROJECT_GENERATOR") then
+            import("private.action.build.pcheader").build(target, jobgraph, "cxx", opt)
+        end
     end, {jobgraph = true})
 


Shiffted:
The `target_cmds.prepare_targets()` got removed here https://github.com/xmake-io/xmake/pull/6718 but this breaks module support since the scanner never gets called.

Shiffted:
https://github.com/xmake-io/xmake/blob/40d790cf416c8677c51a8e969d421e85c25a019b/xmake/plugins/project/clang/compile_commands.lua#L204

There also seems to be a problem with clang, it will add modules twice with different arguments, so the above check will not remove the duplicate. MSVC and GCC seem to work correctly.

Shiffted:
> There also seems to be a problem with clang, it will add modules twice with different arguments, so the above check will not remove the duplicate. MSVC and GCC seem to work correctly.

@Arthapz would it be correct here to disable two phase compilation when in the compile commands generator? It seems to break clang-tidy.

xmake/plugins/project/clang/compile_commands.lua
```diff
@@ -308,6 +308,7 @@ function make(outputdir)
     local oldir = os.cd(os.projectdir())
     local jsonfile = io.open(path.join(outputdir, "compile_commands.json"), "w")
     os.setenv("XMAKE_IN_COMPILE_COMMANDS_PROJECT_GENERATOR", "true")
+    target_cmds.prepare_targets()
```
waruqi, line 311: This would cause the compilation of compile_commands to also precompile pch, which was completely unnecessary. After many users reported various problems, I specifically removed it.

Shiffted: We can check for it in the pcheader rules. What about the other project generators that use `prepare_targets()`, does any  one of them need the pchs?

waruqi: Some project generators require a pch, but some do not, such as the vsxmake generator.

Shiffted:
> would it be correct here to disable two phase compilation when in the compile commands generator? It seems to break clang-tidy.

I will leave this for now as this was also the case before module support broke for compile commands. clangd seems to work with this and clang-tidy will output an error but still report the warnings.

PR #7123, opened 2025-12-11
Add rule for android native app(NDK)

keosu:
- on_load: Add glue file to target and set_kind to shared library
- on_package: Generate android apk package
- on_install: install the app with adb
- on_run: run the app with adb (have trouble to overwrite on_run, use after_install for workaround now)

Test project path: https://github.com/keosu/xmake/tree/dev/tests/projects/c%2B%2B/ndk_raylib

A standalone example repo: [https://github.com/keosu/raycpp_android_xmake/tree/pr](https://github.com/keosu/raycpp_android_xmake/tree/pr) Andriod CI pass

PR #7139, opened 2025-12-16
Add android native app support

waruqi:
- [x] Improve https://github.com/xmake-io/xmake/pull/7123
- [x] Add more examples (raylib, lvgl, ...) 

```console
$ xmake f -p android
$ xmake
$ xmake run
```

```lua

add_requires("raylib 5.5.0")

target("raydemo_basic")
    set_kind("binary")
    set_languages("c++17")
    add_files("src/main.cpp")
    add_syslinks("log")
    add_packages("raylib")
    add_rules("android.native_app", {
        android_sdk_version = "35",
        android_manifest = "android/AndroidManifest.xml",
        android_res = "android/res",
        keystore = "android/debug.jks",
        keystore_pass = "123456",
        package_name = "com.raylib.basic",
        logcat_filters = {"raydemo_basic", "raylib"}
    })

```

image


xmake/rules/platform/android/load.lua
```diff
@@ -0,0 +1,44 @@
...
+    local ndk_root = toolchain_ndk:config("ndk")
+    if not ndk_root then
+        raise("NDK path not set! Please set NDK path properly.")
+    end
+
+    -- set target kind
+    target:set("kind", "shared")
+
+    -- add glue file to target
+    local native_app_glue_file = path.join(ndk_root, "sources", "android", "native_app_glue", "android_native_app_glue.c")
```
SirLynix, line 40: I don't think this should be automatically added when working on Android, you may want to have your own glue using the NDK ([as I did](https://github.com/NazaraEngine/NazaraEngine/pull/413/changes#diff-46d36e99c83f0d39ca6957c1cdd0454ffbb322c7ac765836551319d9ba32d870)).

Maybe add a way to disable the native glue?

Parumezan: True, the [GLFM library,](https://github.com/brackeen/glfm/blob/main/src/glfm_android.c) for example doesn't need it (unless I didn't notice it). I'm working on a project where I used it before switching to GLFM, so being able to disable it would be welcome.

waruqi: try this patch. https://github.com/xmake-io/xmake/pull/7141

Issue #7234, opened 2026-01-16
Lowercase non-English characters via `xmake l string.lower`?


luadebug:
### Xmake Version

xmake v3.0.6+20260104

### Operating System Version and Architecture

Arch Linux

### Describe Bug

```
❯ xmake l string.lower "Звезда Хэнсин"
"Звезда Хэнсин"
```
It should be lowercased, I assume it just skips and do nothing?

### Expected Behavior

It handled non-English characters correctly.
```
❯ xmake l string.lower "Звезда Хэнсин"
"звезда хэнсин"
```

### Project Configuration

Just cli

### Additional Information and Error Logs

Nothing special. Would be perfect if we could resolve `string.upper` for this situation as well.
https://github.com/xmake-io/xmake/issues/7235

Seedking:
Bro, we need utr8proc or to write our own custom solution, which would be simpler and help us avoid platform dependencies

waruqi:
> Bro, we need utr8proc or to write our own custom solution, which would be simpler and help us avoid platform dependencies

xmake doesn't plan to introduce too many third-party dependencies, and may gradually remove some existing third-party library dependencies in the future. tbox has built-in encoding handling and cross-platform support, and we should improve tbox as much as possible to improve and implement it.

Too many dependency libraries can also increase the complexity of package maintenance for Linux distributions such as Debian.

PR #7235, opened 2026-01-16
add string case conversion functions: lower and upper

luadebug:
Resolves: https://github.com/xmake-io/xmake/issues/7234

### DIFF
--- a/core/src/xmake/string/lower.c
+++ b/core/src/xmake/string/lower.c
@@ -52,7 +52,7 @@ tb_int_t xm_string_lower(lua_State *lua) {
     }
 
     // copy string to buffer
-    tb_char_t* buffer = tb_malloc_bytes(size + 1);
+    tb_char_t* buffer = (tb_char_t*)tb_malloc_bytes(size + 1);
     if (buffer) {
         tb_memcpy(buffer, cstr, size);
         buffer[size] = '\0';
--- a/core/src/xmake/string/upper.c
+++ b/core/src/xmake/string/upper.c
@@ -52,7 +52,7 @@ tb_int_t xm_string_upper(lua_State *lua) {
     }
 
     // copy string to buffer
-    tb_char_t* buffer = tb_malloc_bytes(size + 1);
+    tb_char_t* buffer = (tb_char_t*)tb_malloc_bytes(size + 1);
     if (buffer) {
         tb_memcpy(buffer, cstr, size);
         buffer[size] = '\0';


waruqi:
Please do not include third-party dependencies. xmake strictly controls the inclusion of third-party libraries and may gradually remove some existing third-party library dependencies in the future. tbox has built-in encoding handling and cross-platform support; we should improve tbox as much as possible to improve and implement these features.

luadebug:
> Please do not include third-party dependencies. xmake strictly controls the inclusion of third-party libraries and may gradually remove some existing third-party library dependencies in the future. tbox has built-in encoding handling and cross-platform support; we should improve tbox as much as possible to improve and implement these features.

OK. What is a solution for this situation?

waruqi:
Supporting full character set conversion seems like a huge undertaking, potentially increasing the xmake binary by over 1-2MB.

https://github.com/JuliaStrings/utf8proc/blob/e5e799221b45bbb90f5fdc5c69b6b8dfbf017e78/utf8proc_data.c (2.25MB)

We should improve the `charset` module of `tbox` to provide `tolower`/`toupper` interfaces.

Then, for some system platforms, we can use existing system APIs to support it (e.g., Windows `CharUpperBuffW`/`CharLowerBuffW`, Apple `NSString uppercaseString`/`lowercaseString`, `CFStringLowercase`). If no existing system API support is available, we can fall back to our own simplified implementation, only providing conversions for some commonly used language character sets.

This might be quite complex and time-consuming. If you can't implement it in a short time, you can leave it for now. I'll try to support it when I have more time later.

luadebug:
> Supporting full character set conversion seems like a huge undertaking, potentially increasing the xmake binary by over 1-2MB.
> 
> https://github.com/JuliaStrings/utf8proc/blob/e5e799221b45bbb90f5fdc5c69b6b8dfbf017e78/utf8proc_data.c (2.25MB)
> 
> We should improve the `charset` module of `tbox` to provide `tolower`/`toupper` interfaces.
> 
> Then, for some system platforms, we can use existing system APIs to support it (e.g., Windows `CharUpperBuffW`/`CharLowerBuffW`, Apple `NSString uppercaseString`/`lowercaseString`, `CFStringLowercase`). If no existing system API support is available, we can fall back to our own simplified implementation, only providing conversions for some commonly used language character sets.
> 
> This might be quite complex and time-consuming. If you can't implement it in a short time, you can leave it for now. I'll try to support it when I have more time later.

Well I think that would require to run some DragonflyBSD/Fedora maybe even another CIs to validate that things works as expected, as currently it triggered just Windows/Linux/MacOS, but skipped another OS checks.

luadebug:
Awww just using **setlocale with mutex** slowed down our MacOS I guess... :/

```
>> [149/183]: testing tests/benchmarks/config_targets ...
>>     running test_config ...
config targets/1k: xmake: 4701 ms
config targets/1k: cmake/default: 1332 ms
>>     test failed: require true but got false
>>       function test_config tests/benchmarks/config_targets/test.lua:50
error: @programdir/core/main.lua:274: ./tests/runner.lua:62: ./tests/test_utils/print_error.lua:46: aborting because of failed assertion ...
stack traceback:
    [C]: in function 'error'
    [@programdir/core/base/os.lua:1125]:
    [./tests/test_utils/print_error.lua:46]: in function 'print_error'
    [./tests/test_utils/test_assert.lua:9]: in function 'require'
    [./tests/benchmarks/config_targets/test.lua:50]:

stack traceback:
	[C]: in function 'error'
	@programdir/core/base/os.lua:1125: in function 'os.raiselevel'
	(...tail calls...)
	@programdir/core/main.lua:274: in upvalue 'cotask'
	@programdir/core/base/scheduler.lua:514: in function <@programdir/core/base/scheduler.lua:507>
warning: Appleclang isn't shipped with clang-scan-deps, disabling modules tests
warning: llvm toolchain not found!
warning: llvm not found, skipping tests
Error: Process completed with exit code 255.
```

waruqi:
wait for this patch. https://github.com/tboox/tbox/pull/300

then we can use tb_charset_utf8_tolower/tb_charset_utf8_toupper, it will call tb_wcslwr/tb_wcsupr to do it.

It makes some optimizations. Since 90% of the time the strings are ASCII strings, executing the Unicode tolower/toupper completely every time would significantly impact performance, so we don't need to do that.

```c
tb_long_t tb_charset_utf8_tolower(tb_char_t* s, tb_size_t n)
{
    // check
    tb_assert_and_check_return_val(s, -1);

    // try ascii tolower first
    tb_char_t* p = s;
    tb_char_t* e = s + n;
    while (p < e && *p)
    {
        if ((*p) & 0x80) break; 
        *p = tb_tolower(*p);
        p++;
    }
    if (p == e || !*p) return p - s;

    // convert the suffix to wchar_t
    tb_long_t   r = -1;
    tb_size_t   wn = e - p + 1;
    tb_wchar_t  wb[256];
    tb_wchar_t* w = (wn <= 256)? wb : (tb_wchar_t*)tb_malloc(wn * sizeof(tb_wchar_t));
    if (w)
    {
        // to lower
        if (tb_mbstowcs(w, p, wn) != -1)
        {
            tb_wcslwr(w);
            r = tb_wcstombs(p, w, wn);
            if (r != -1) r += (p - s);
        }

        // free it
        if (w != wb) tb_free(w);
    }
    return r;
}
```

luadebug:
> wait for this patch. [tboox/tbox#300](https://github.com/tboox/tbox/pull/300)
> 
> then we can use tb_charset_utf8_tolower/tb_charset_utf8_toupper, it will call tb_wcslwr/tb_wcsupr to do it.
> 
> It makes some optimizations. Since 90% of the time the strings are ASCII strings, executing the Unicode tolower/toupper completely every time would significantly impact performance, so we don't need to do that.
> 
> ```c
> tb_long_t tb_charset_utf8_tolower(tb_char_t* s, tb_size_t n)
> {
>     // check
>     tb_assert_and_check_return_val(s, -1);
> 
>     // try ascii tolower first
>     tb_char_t* p = s;
>     tb_char_t* e = s + n;
>     while (p < e && *p)
>     {
>         if ((*p) & 0x80) break; 
>         *p = tb_tolower(*p);
>         p++;
>     }
>     if (p == e || !*p) return p - s;
> 
>     // convert the suffix to wchar_t
>     tb_long_t   r = -1;
>     tb_size_t   wn = e - p + 1;
>     tb_wchar_t  wb[256];
>     tb_wchar_t* w = (wn <= 256)? wb : (tb_wchar_t*)tb_malloc(wn * sizeof(tb_wchar_t));
>     if (w)
>     {
>         // to lower
>         if (tb_mbstowcs(w, p, wn) != -1)
>         {
>             tb_wcslwr(w);
>             r = tb_wcstombs(p, w, wn);
>             if (r != -1) r += (p - s);
>         }
> 
>         // free it
>         if (w != wb) tb_free(w);
>     }
>     return r;
> }
> ```

Looks cool, would be nice to have it clean without windows.h or some appleframework?

SirLynix:
I don't know if tb_wchar_t maps to wchar_t but beware on Windows wchar_t is UTF-16 and has surrogate pairs, which means a single wchar_t may be half a character.

waruqi:
> I don't know if tb_wchar_t maps to wchar_t but beware on Windows wchar_t is UTF-16 and has surrogate pairs, which means a single wchar_t may be half a character.

Therefore, I would prioritize using the CharUpperW API to implement complete string conversion, as is the case on Apple's platform. A while-loop + towupper is the fallback solution. @luadebug

luadebug:
> > I don't know if tb_wchar_t maps to wchar_t but beware on Windows wchar_t is UTF-16 and has surrogate pairs, which means a single wchar_t may be half a character.
> 
> Therefore, I would prioritize using the CharUpperW API to implement complete string conversion, as is the case on Apple's platform. A while-loop + towupper is the fallback solution. @luadebug

Well sounds good as we have them included at any case.

waruqi:
We also need to update tbox.config.h for each platform to add detection results such as wcsupr and towlower. Usually only towlower and towupper exist.

waruqi:
wait for this patch https://github.com/xmake-io/xmake/pull/7237

waruqi:
rebase dev and try it again.

core/src/xmake/string/case.c
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // allocate buffer for optimistic ascii conversion
+    tb_char_t* buf = (tb_char_t*)tb_malloc_bytes(size);
+    if (!buf) return 0;
+
+    tb_size_t i;
+    for (i = 0; i < size; i++) {
+        // Stop at first non-ascii char
+        if ((tb_byte_t)str[i] >= 0x80) break;
```
waruqi: we can use tb_charset_utf8_toupper directly https://github.com/tboox/tbox/blob/73028bf9165aece5ec9863124fda67500a005e24/src/tbox/charset/utf8.c#L258

core/src/xmake/string/case.c
```diff
@@ -0,0 +1,178 @@
...
+        lua_pop(lua, 1); 
+    }
+    lua_pushlstring(lua, str, size);
+    return 1;
+}
+
+/* //////////////////////////////////////////////////////////////////////////////////////
+ * implementation
+ */
+tb_int_t xm_string_lower(lua_State* lua) {
```
waruqi: please move to lower.c and upper.c file

waruqi:
like this

```c
/* string.upper
 *
 * @param str       the string
 *
 * @code
 *      local result = string.upper(str)
 * @endcode
 */
tb_int_t xm_string_toupper(lua_State *lua) {
    // check
    tb_assert_and_check_return_val(lua, 0);

    // get string
    size_t           size = 0;
    tb_char_t const* cstr = luaL_checklstring(lua, 1, &size);
    tb_check_return_val(cstr, 0);

    // empty?
    if (!size) {
        lua_pushstring(lua, "");
        return 1;
    }

    // copy string to buffer
    tb_char_t* buffer = tb_malloc_bytes(size + 1);
    if (buffer) {
        tb_memcpy(buffer, cstr, size);
        buffer[size] = '\0';

        // to upper
        tb_long_t real_size = tb_charset_utf8_toupper(buffer, size);
        
        // push result
        if (real_size >= 0) {
            lua_pushlstring(lua, buffer, real_size);
        } else {
            lua_pushlstring(lua, cstr, size);
        }
        tb_free(buffer);
    } else {
        lua_pushnil(lua);
    }
    return 1;
}
```

Issue #7238, opened 2026-01-18
string.sub non-English characters via xmake l string.sub?


luadebug:
### Xmake Version

xmake v3.0.6+20260104

### Operating System Version and Architecture

Arch Linux

### Describe Bug

Non-English takes a byte instead of two bytes...
```
❯ xmake l string.sub "ЗВЕЗДА" 1
"ЗВЕЗДА"

❯ xmake l string.sub "ЗВЕЗДА" 2
"�ВЕЗДА"

❯ xmake l string.sub "ЗВЕЗДА" 3
"ВЕЗДА"
```

```
❯ xmake l string.sub "star" 1
"star"

❯ xmake l string.sub "star" 2
"tar"

❯ xmake l string.sub "star" 3
"ar"
```

### Expected Behavior

```
xmake l string.utf8_sub "ЗВЕЗДА" 1
"ЗВЕЗДА"

xmake l string.utf8_sub "ЗВЕЗДА" 2
"ВЕЗДА"

xmake l string.utf8_sub "ЗВЕЗДА" 3
"ЕЗДА"

### Project Configuration

Just cli

### Additional Information and Error Logs

Nothing special. Would be perfect if we could resolve that by shadowing/overriding existing .sub as we just for `.lower()`/`.upper()` or either define it as string.utf8_sub instead of string.sub so we would not hide original .sub

**Comment by waruqi** [comment]
lua string use utf8 string now. We cannot completely separate characters by Unicode. A Unicode character can also be 4 bytes.

**Comment by waruqi** [comment]
`string.sub` should always be split by bytes; this is its default behavior. We should decide for ourselves the length of the Unicode characters and from which byte to extract the substring.

like this:

```lua
xmake> string.sub("ЗВЕЗДА", #("ЗВ") + 1)
< "ЕЗДА"
```

waruqi:
If we want to split the string completely according to the length of the Unicode characters, this is not something that Lua string should do. It is equivalent to mbs strings. We should add a separate Lua module called wstring to directly encapsulate and process wcs string.

SirLynix:
Please call it utf32 or unicode but not wide characters, wide characters are plagued by Windows using them for UTF-16 (which means characters outside of the BMP takes two wide characters)

waruqi:
So we can add a new ustring.lua (unicode string) to handle it.It provides the exact same list of APIs as Lua strings, and offers mutual conversion between strings and Lua strings. 

[...]

Issue #7184, opened 2026-01-04
Improve package to support multiple different installation schemes


waruqi:
### Is your feature request related to a problem? Please describe.

While the current package installation configuration supports multiple URL mirrors, the source code packages for these URLs must be completely identical.

If a package has multiple completely different URL package files and completely different installation methods, then it cannot be supported.

However, sometimes, some packages may need to try multiple completely different URL installation methods at the same time, such as: 
1. binary URL installation
2. source package URL installation.

like this:

```lua
for _, url, ipairs({binary_url, sourcecode_url, other_urls, ...}) do
    if not on_download(package, url) then
        continue
    end
    
    if not on_install(package) then
        continue
    end
    
    if on_test(package) then
        -- install success
        break
    else
        continue
    end
end
```

We need to define a new install scheme concept to organize multiple installation schemes. This will allow a package to try multiple URL installation schemes simultaneously.

like this:

```lua
package("cmake")

    -- add urls for binary scheme
    add_urls("https://cmake.org/files/v$(version)-x86_64.tar.gz", {scheme = "binary"})
    add_urls("https://github.com/Kitware/CMake/releases/download/v$(version)-x86_64.tar.gz", {scheme = "binary"})
    add_versions("4.2.1",  "c059bff1e97a2b6b5b0c0872263627486345ad0ed083298cb21cff2eda883980", {scheme = "binary"})

    -- add urls for source scheme
    add_urls("https://github.com/Kitware/CMake/releases/download/v$(version)/cmake-$(version).tar.gz", {scheme = "source"})
    add_versions("4.2.1",  "414aacfac54ba0e78e64a018720b64ed6bfca14b587047b8b3489f407a14a070", {scheme = "source"})
    
    -- add multiple install schemes,
    -- it will try installing package from binary urls first, then fallback to install it from source url
    add_installschemes("binary", "source")
        
    on_install(function (package)
        -- if install package from source url
        if package:scheme() == "source" then
            import("core.base.option")
            os.vrunv("sh", {"./bootstrap", "--parallel=" .. (option.get("jobs") or tostring(os.default_njob())),
                "--prefix=" .. package:installdir(), "--", "-DCMAKE_USE_OPENSSL=OFF"})
            import("package.tools.make").install(package)
        else
            -- install package from binary url
            os.cp("bin", package:installdir())
            os.cp("share", package:installdir())
        end
    end)

    on_test(function (package)
        -- If the binary package fails to execute due to some Linux distributions,
[...]

PR #7241, opened 2026-01-18
implement utfsub/utflen/utffind/utfreverse/utfbyte/utfchar/utflastof functions for substring extraction in UTF-8 strings

luadebug:
Resolves: https://github.com/xmake-io/xmake/issues/7238

Achieve similiar behavior as python3 slicing...

```
>>> "🎆🎉🌮"[:-2]
'🎆'
>>> "star"[:-2]
'st'
>>> "ЗВЕЗДА"[:-2]
'ЗВЕЗ'
>>> "源文件"[:-2]
'源'
```

Perhaps maybe we need more functions to implement? Let me know...

Also we should decide what should be shadowed... As we discussed string.sub should not be shadowed that is why we going to have string.sub and string.utfsub, instead of just string.sub (with implementation of string.utfsub)... What about utflen/utffind/utfreverse... should they shadow len/find/reverse?

### DIFF
--- a/core/src/xmake/engine.c
+++ b/core/src/xmake/engine.c
@@ -294,6 +294,7 @@ tb_int_t xm_string_utffind(lua_State* lua);
 tb_int_t xm_string_utfreverse(lua_State* lua);
 tb_int_t xm_string_utfbyte(lua_State* lua);
 tb_int_t xm_string_utfchar(lua_State* lua);
+tb_int_t xm_string_utflastof(lua_State* lua);
 
 // the process functions
 tb_int_t xm_process_open(lua_State *lua);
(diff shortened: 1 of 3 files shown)


waruqi:
There are many APIs that overlap with `string`, only differing in whether or not UTF-8 support is provided. This leads to confusion in current `string` usage, making it difficult for users to distinguish when to use `string.sub lastof` and when to use `utfsub utflastof`. Furthermore, implementing these directly based on UTF-8 is extremely inefficient.

We should provide better support through a new `ustring` module. This can be discussed in detail in a new issue regarding its design and implementation.

Additionally, any new API additions must first be created as a new feature request in an issue, undergo detailed API design, and be approved before a pull request is opened for implementation.  

like this https://github.com/xmake-io/xmake/issues/7184#issuecomment-3729515195

waruqi:
https://github.com/xmake-io/xmake/issues/7242

waruqi:
https://github.com/xmake-io/xmake/pull/7246

Issue #7261, opened 2026-01-23
使用 set_pcxxheader() 导致无法增量编译


terrywh:
### Xmake 版本

xmake v3.0.6+20260122

### 操作系统版本和架构

Linux 308946ae8192 6.6.80-29.tl4.x86_64 

### 描述问题

使用 `set_pcxxheader("core/vendor.ipp")` 后，每次编译都会全部重新开始，猜测由于每次都重新生成了一个独立的 include 文件导致；

<img width="1512" height="156" alt="Image" src="https://github.com/user-attachments/assets/c40669f9-2fa7-4555-b9cc-8cd451e71d9f" />

<img width="564" height="184" alt="Image" src="https://github.com/user-attachments/assets/b268d1b7-05a4-4eb1-8dc6-35633114d354" />

取消此行设置后回复正常：

<img width="1458" height="118" alt="Image" src="https://github.com/user-attachments/assets/1082e73a-de45-4d67-be87-a86fc10aa5d9" />

### 期待的结果

能够正常使用 set_pcxxheader() 功能，进行增量编译；

### 工程配置

target("xxxx")
    set_pcxxheader("core/vendor.ipp")
    add_files("*.cpp")

### 附加信息和错误日志

使用 set_pcxxheader() 时 g++ 会输出一个警告信息：
```
make build
[  1%]: cache compiling.debug core/vendor.ipp
g++: warning: core/vendor.ipp: linker input file unused because linking not done
```

waruqi:
我这里没法复现，请提供完整可复现工程。

```console
ruki@095ba1d12c28:/mnt/xmake/tests/projects/c++/precompiled_header$ touch src/test.cpp 
ruki@095ba1d12c28:/mnt/xmake/tests/projects/c++/precompiled_header$ xmake
[ 14%]: cache compiling.release src/test.cpp
[ 80%]: linking.release main
[100%]: build ok, spent 0.458s
```

terrywh:
@waruqi 

<img width="1326" height="1612" alt="Image" src="https://github.com/user-attachments/assets/bfa1768e-6971-4e31-9dca-2e5852dd82f2" />

[example.tgz](https://github.com/user-attachments/files/24851839/example.tgz)

i think there's something to do with the extension of the "vendor.ipp" file.  (change it to "vendor.hpp" then it will not happen)

luadebug:
Can you acquire similar issue during clang toolchain run? As I think not just gcc affected...
[...]

PR #7264, opened 2026-01-26
add support for .ipp precompiled headers in C++ project

luadebug:
#7261

### DIFF
--- a/xmake/modules/private/action/build/pcheader.lua
+++ b/xmake/modules/private/action/build/pcheader.lua
@@ -95,9 +95,6 @@ function build(target, jobgraph, langkind, opt)
             if target:has_tool(sourcekind, "gcc", "gxx", "clang", "clang++") then
                 configs.force = configs.force or {}
                 configs.force.cxflags = (sourcekind == "cxx" and "-x c++-header" or "-x c-header")
-            elseif target:has_tool(sourcekind, "cl", "clang_cl") then
-                configs.force = configs.force or {}
-                configs.force.cxflags = (sourcekind == "cxx" and "/TP" or "/TC")
             end
         end
         local compflags = compinst:compflags({target = target, sourcefile = sourcefile, configs = configs})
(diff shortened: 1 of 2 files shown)


waruqi:
We should not modify the pch rule, as there are many incorrect fixes in it, and some logic has already been handled in gcc/cl.lua.

Furthermore, we don't need to add an extra test project just for the .ipp header file extension. There are many header file extensions, and simply adding .ipp won't cover all of them.

waruqi:
see https://github.com/xmake-io/xmake/pull/7266

xmake/modules/private/action/build/pcheader.lua
```diff
@@ -59,13 +61,93 @@ end
...
+            [".hh"] = true,
+            [".hpp"] = true,
+            [".hxx"] = true,
+            [".h++"] = true,
+            [".tcc"] = true,
+            [".inl"] = true,
+            [".ii"] = true,
+            [".ixx"] = true,
+            [".cppm"] = true,
+            [".mpp"] = true,
```
waruqi, line 92: Many of these are not header file extensions, so we don't need to treat them as precompiled header files.

xmake/modules/private/action/build/pcheader.lua
```diff
@@ -59,13 +61,93 @@ end
...
+
+        -- filter flags
+        local compflags_new = {}
+        local skip = 0
+        for i, flag in ipairs(compflags) do
+            if skip > 0 then
+                skip = skip - 1
+            elseif flag == "-include-pch" then
+                skip = 1
+            elseif flag == "-include" then
```
waruqi, line 110: Different compilers have different flags, and we should not hardcode them globally. MSVC does not have these flags, which would lead to unnecessary for-loops and affect performance.

waruqi: This has already been handled, and it's specifically for gcc/clang. (https://github.com/xmake-io/xmake/blob/7f68201e903054c42a5820c3d9ede6851d9d8821/xmake/modules/core/tools/gcc.lua#L344)

xmake/modules/private/action/build/pcheader.lua
```diff
@@ -59,13 +61,93 @@ end
...
+
+            -- do compile
+            dependinfo.files = {}
+            assert(compinst:compile(sourcefile, objectfile, {dependinfo = dependinfo, compflags = compflags, pcheader = true}))
+
+            -- update files and values to the dependent file
+            dependinfo.values = depvalues
+            table.insert(dependinfo.files, sourcefile)
+            depend.save(dependinfo, dependfile)
+
```
waruqi, line 150: The previous implementation supported incremental compilation, which I was unable to reproduce, and I don't think it should be modified.

PR #7266, opened 2026-01-26
fix pch header extension

waruqi:
https://github.com/xmake-io/xmake/issues/7261#issuecomment-3797709229


luadebug:
Thanks :D Would that be convenient to add that project as test as I did in my PR?

waruqi:
> Thanks :D Would that be convenient to add that project as test as I did in my PR?

I don't think it's necessary to add a separate test project. We can add multiple targets to the existing pch test example to test different extensions.

Issue #6488, opened 2025-05-22
Improve Nim language support as `target`


griffith1deady:
### Xmake Version

xmake v2.9.9+dev.a6af349ad, A cross-platform build utility based on Lua

### Operating System Version and Architecture

Windows 11 IoT 24H2

### Describe Bug

The essence of the problem: if you use a single-file project, everything is fine. If you make a project with any structure that includes for example:

main.nim:

```nim
import something
echo "Hello, world!
```

something.nim:

```nim
echo "Hello from something!"
```

At the first compilation everything will be fine. but if you only make changes to the files on which `add_files("src/main.nim")` depends, the target will not rebuilded. Also, `add_files("src/*.nim")` does not solve the problem, because each file will be built as a separate application, which is unacceptable.
I love xmake and use it extensively with nim, but now I have to use `task` which will run `os.run("nim c ...")` because there is no support for anything more complex than a single-file project, but I love the work done on xrepo. also, passing arguments to the native linker or native compiler doesn't work properly, causing the nim compiler to think we're trying to pass arguments to a running application and display an error.
also, `importc` paired with add_deps/add_packages won't properly link `static`/`shared` dependencies for some strange reason.

### Expected Behavior

- [ ] Correctly rebuild target when project files changed, not only main target file;
- [ ] Correctly link `static`/`shared` dependencies (also can come from xrepo);
- [ ] Correctly pass linker/compiler argument's as `passL:"..."` and `passC:"..."` to nim compiler

### Project Configuration

target("sellage")
    set_kind("binary")
    add_files("src/main.nim")

see $describe bug for project example

### Additional Information and Error Logs

None. I don't remember all aspect's that not work with static/shared dependencies, but i can to deal with it later if the problem is not on the surface.

waruqi:
Incremental compilation is not currently supported because it is not known how to analyze and obtain their dependencies.

griffith1deady:
> Incremental compilation is not currently supported because it is not known how to analyze and obtain their dependencies.

I'm not talking about incremental builds, I'm talking about rebuilding the target if the project files have changed? That's different, no?

waruqi:
If we just rebuild the entire project, it works for me.

```console
ruki:console ruki$ xmake
[ 63%]: linking.release test
[100%]: build ok, spent 0.359s
ruki:console ruki$ touch src/main.nim
ruki:console ruki$ xmake -v
[ 63%]: linking.release test
[...]

PR #7268, opened 2026-01-26
add dependency file generation for Nim source files

luadebug:
https://github.com/xmake-io/xmake/issues/6488

Lets assume we have got project like:

main.nim
```nim
import something
echo "Hello, world!"
```

something.nim
```nim
echo "Hello from something!"
```

xmake.lua
```xmake
set_project("import_module")

target("import_module")
    set_kind("binary")
    add_files("main.nim")
```

Once we go for `xmake run` it works, but once we edit `something.nim` file manually with vscode editor for example and save it to

something.nim
```nim
echo "Hello from something!"

import std/[times]
const buildTime = getTime()
echo "Project builded at: ", buildTime
```

and use `xmake b import_module && xmake run` it would skip our changes and executable would not execute those 3 new lines.

### DIFF
--- a/xmake/rules/nim/build/target.lua
+++ b/xmake/rules/nim/build/target.lua
@@ -25,6 +25,45 @@ import("core.tool.compiler")
 import("core.project.depend")
 import("utils.progress")
 
+-- generate dependency info
+function _generate_dependinfo(compinst, compflags, sourcefiles, dependinfo)
+
+    -- generate dependency file (.deps)
+    local flags = {}
+    local nimcache
+    for _, flag in ipairs(compflags) do
+        table.insert(flags, flag)
+        if flag:startswith("--nimcache:") then
+            nimcache = flag:sub(12)
+        end
+    end
+    table.insert(flags, "--genScript")
+
+    -- run nim --genScript to generate .deps file
+    local program = compinst:program()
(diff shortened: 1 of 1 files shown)


xmake/rules/nim/build/target.lua
```diff
@@ -66,6 +66,41 @@ function build_sourcefiles(target, sourcebatch, opt)
...
+        if flag:startswith("--nimcache:") then
+            nimcache = flag:sub(12)
+        end
+    end
+    table.insert(flags, "--genScript")
+    
+    -- run nim --genScript to generate .deps file
+    local program = compinst:program()
+    local argv = table.join("c", flags, sourcefiles)
+    os.runv(program, argv, {envs = compinst:_tool():runenvs()})
```
waruqi: compinst:runenvs()

xmake/rules/nim/build/target.lua
```diff
@@ -66,6 +66,41 @@ function build_sourcefiles(target, sourcebatch, opt)
     dependinfo.files = {}
     assert(compinst:build(sourcefiles, targetfile, {target = target, dependinfo = dependinfo, compflags = compflags}))
 
+    -- generate dependency file (.deps)
+    local flags = {}
```
waruqi: move to the seperate function

Issue #7180, opened 2026-01-01
When using zig for cross-compilation, the host tool chain cannot be correctly determined.


Redbeanw44602:
### Xmake Version

dev

### Operating System Version and Architecture

Archlinux

### Describe Bug

When using zig, xmake failed to pass in the correct compilation parameters when compiling the `~host` package.

```
[22/49] /tmp/.xmake1000/260101/zigcc/c++ -DENABLE_DARTS -DLOCALEDIR=\"/home/neko/.xmake/packages/o/opencc/1.1.9/8122471a83e545cfba0149a3418b377e/share/locale\" -DOpencc_BUILT_AS_STATIC -DPACKAGE_NAME=\"opencc\" -DPKGDATADIR=\"/home/neko/.xmake/packages/o/opencc/1.1.9/8122471a83e545cfba0149a3418b377e/share/opencc\" -DVERSION=\"1.1.9\" -I"/home/neko/.xmake/cache/packages/2601/o/opencc~host/1.1.9/source/build_d28d31d0/src" -I"/home/neko/.xmake/cache/packages/2601/o/opencc~host/1.1.9/source/src" -I"/home/neko/.xmake/cache/packages/2601/o/opencc~host/1.1.9/source" -isystem /home/neko/.xmake/packages/m/marisa/v0.2.6/61baee0fb4f149009cc4dd74483dd75d/lib/pkgconfig/../../include -isystem /home/neko/.xmake/packages/r/rapidjson/2025.02.05/94dc0246f1a54b63a255bff5f0bd27a8/include -isystem /home/neko/.xmake/packages/t/tclap/v1.2.5/f5fdc882170b4e8ba690441b4cda28d9/lib/pkgconfig/../../include -isystem /home/neko/.xmake/packages/d/darts/0.32/71099d8dd1ab4d198ffd8c09282818fe/lib/pkgconfig/../../include -target aarch64-linux-gnu.2.27 -O3 -DNDEBUG -fPIE   -std=c++14 -Wall -MD -MT src/tools/CMakeFiles/opencc_dict.dir/DictConverter.cpp.o -MF src/tools/CMakeFiles/opencc_dict.dir/DictConverter.cpp.o.d -o src/tools/CMakeFiles/opencc_dict.dir/DictConverter.cpp.o -c '/home/neko/.xmake/cache/packages/2601/o/opencc~host/1.1.9/source/src/tools/DictConverter.cpp'
[...]

PR #7269, opened 2026-01-27
add target architecture validation for cross-compilation in zig toolchain

luadebug:
https://github.com/xmake-io/xmake/issues/7180

hello.cpp
```cpp
#include <iostream>
#include <opencc/Config.hpp>

int main() {
    opencc::Config config;
    std::cout << "Hello, World!" << std::endl;
    return 0;
}
```

xmake.lua
```xmake
set_project("zig_cross_compilation_host_aarch64")

add_requires("opencc", {system = false, configs = {shared = false}}) 
add_packages("opencc")

target("zig_cross_compilation_host_aarch64")
    set_kind("binary")
    add_files("hello.cpp")
```

Lets assume we cross-compile from amd64 to aarch64... Run `xmake f -vD -p linux -m debug -a arm64-v8a --toolchain=zig --cross=aarch64-linux-gnu && xmake`

So issue is that it tries to execute aarch64 executable on my amd64 machine... We need to make sure ~host postfix packages are being built to host processor not to target processor.
As `/home/lin/Downloads/packages/2601/o/opencc~host/1.1.9/source/build_cc7a577c/src/tools/opencc_dict` is very good example of arm64 executable file according to `file /home/lin/Downloads/packages/2601/o/opencc~host/1.1.9/source/build_cc7a577c/src/tools/opencc_dict`

### DIFF
--- a/xmake/toolchains/zig/xmake.lua
+++ b/xmake/toolchains/zig/xmake.lua
@@ -78,6 +78,7 @@ toolchain("zig")
 
     on_load(function (toolchain)
         import("core.base.semver")
+        import("private.core.base.is_cross")
 
         -- set toolset
         -- we patch target to `zig cc` to fix has_flags. see https://github.com/xmake-io/xmake/issues/955#issuecomment-766929692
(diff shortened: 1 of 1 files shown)


xmake/toolchains/zig/xmake.lua
```diff
@@ -135,6 +135,16 @@ toolchain("zig")
...
         -- init target
         local target = toolchain:cross()
+
+        if target then
+            local tarch = target:match("^(.-)-") or target
+            if arch == "x86_64" and tarch ~= "x86_64" then
+                target = nil
+            elseif arch == "aarch64" and tarch ~= "aarch64" then
+                target = nil
+            end
```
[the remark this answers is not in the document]
waruqi: try simplify it, and add some comments and issue link

xmake/toolchains/zig/xmake.lua
```diff
@@ -135,6 +135,14 @@ toolchain("zig")
 
         -- init target
         local target = toolchain:cross()
+
+        if target then
+            local tarch = target:match("^(.-)-") or target
```
waruqi: I think we should use `target:split("%-")[1]`

waruqi: Additionally, some cross strings may have different names but are essentially the same. e.g.

cross: `i686-xxx` -> arch: i386
cross: `aarch64-xxx` -> arch: arm64

luadebug: %_% I am now confused exactly how this should look like ;D

Redbeanw44602:
```
[37/49] cd /home/neko/.xmake/cache/packages/2601/o/opencc/1.1.9/source/build_2a9b6ed8/data && /home/neko/.xmake/packages/o/opencc/1.1.9/8122471a83e545cfba0149a3418b377e/bin/opencc_dict --input /home/neko/.xmake/cache/packages/2601/o/opencc/1.1.9/source/data/dictionary/TWVariantsRevPhrases.txt --output TWVariantsRevPhrases.ocd2 --from text --to ocd2
FAILED: [code=127] data/TWVariantsRevPhrases.ocd2 /home/neko/.xmake/cache/packages/2601/o/opencc/1.1.9/source/build_2a9b6ed8/data/TWVariantsRevPhrases.ocd2 
cd /home/neko/.xmake/cache/packages/2601/o/opencc/1.1.9/source/build_2a9b6ed8/data && /home/neko/.xmake/packages/o/opencc/1.1.9/8122471a83e545cfba0149a3418b377e/bin/opencc_dict --input /home/neko/.xmake/cache/packages/2601/o/opencc/1.1.9/source/data/dictionary/TWVariantsRevPhrases.txt --output TWVariantsRevPhrases.ocd2 --from text --to ocd2
/bin/sh: line 1: /home/neko/.xmake/packages/o/opencc/1.1.9/8122471a83e545cfba0149a3418b377e/bin/opencc_dict: No such file or directory
```
It's a bit strange; the `opencc~host` build doesn't seem to be triggered.

```
$ xmake --version
xmake v3.0.6+dev.e1d3f3555, ...
```

luadebug:
If you remove `rm -rf /home/neko/.xmake/packages/o/opencc/` I think nothing would change?
Well as for me this works without issues :(

Redbeanw44602:
I just tested it with actions, and everything works perfectly. It's mysterious xmake on my pc always has some issues that others can't reproduce...

Redbeanw44602:
The problem is that I have the opencc installed on my Archlinux system, and xmake works correctly if I use a clean container.

opencc's `system = false` setting is not passed to `opencc~host` (but it's reasonable?),

Considering that any build using zigcc should use `add_requireconfs('**', {system = false})`.

PR #7274, opened 2026-01-28
Add os.access function for file access checking

luadebug:
It looks like `os.isexec` could be improved?

### DIFF
--- a/core/src/xmake/engine.c
+++ b/core/src/xmake/engine.c
@@ -483,7 +483,7 @@ static luaL_Reg const g_winos_functions[] = {
     { "registry_keys", xm_winos_registry_keys },
     { "registry_values", xm_winos_registry_values },
     { "short_path", xm_winos_short_path },
-    { "is_pefile", xm_winos_is_pefile },
+    { "is_pefile", xm_winos_ispefile },
     { tb_null, tb_null },
 };
 #endif


xmake/core/base/os.lua
```diff
@@ -1155,15 +1156,9 @@ function os.isexec(filepath)
 
     -- check executable program exist
     if os.isfile(filepath) then
-        return true
-    end
-    if os.host() == "windows" then
-        for _, suffix in ipairs({".exe", ".cmd", ".bat"}) do
```
waruqi, line 1161: tb_file_access currently does not support checking exec on Windows now, as this will break many things.

xmake/core/base/os.lua
```diff
@@ -1155,7 +1156,11 @@ function os.isexec(filepath)
 
     -- check executable program exist
     if os.isfile(filepath) then
-        return true
+        if os.host() == "windows" then
```
waruqi: Files without the .exe or .bat extensions, even if they exist, are not necessarily executable.

luadebug: I am a bit confused as maybe we miss some `.ps1` extension...

luadebug: Maybe there is some mechanism that exists for Detect It Easy program required... Like duh implementing PE header parser with xmake l io.parsepe

waruqi: I think that on Windows, simply checking the file extension should be sufficient, and we can add any missing extensions.That's enough, at least for now.

luadebug: OK. I am just confused how to finish this PR so.

xmake/core/sandbox/modules/io.lua
```diff
@@ -372,6 +372,13 @@ function sandbox_io.tail(filepath, linecount, opt)
     io.tail(filepath, linecount, opt)
 end
 
+-- parse pe
+function sandbox_io.parse_pe(filepath)
+    assert(filepath)
+    filepath = vformat(filepath)
+    return io.parse_pe(filepath)
```
waruqi: we need not this api, please remove it

xmake/core/base/os.lua
```diff
@@ -1155,15 +1156,23 @@ function os.isexec(filepath)
...
+            local ext = path.extension(filepath):lower()
+            if ext == ".exe" or ext == ".cmd" or ext == ".bat" or ext == ".ps1" then
```
waruqi: I thought about it again, and on Windows, simply checking if the file exists should be sufficient.

Using only the file extension isn't accurate; it might overlook some executable files, such as .sh scripts on msys2, and shell scripts might not even have a file extension.

waruqi: The current rules for filtering file extensions may break some things when built on msys2.

e.g. 7z.sh on msys2

xmake/core/base/os.lua
```diff
@@ -1158,15 +1158,15 @@ function os.isexec(filepath)
     if os.isfile(filepath) then
         if os.host() == "windows" then
             local ext = path.extension(filepath):lower()
-            if ext == ".exe" or ext == ".cmd" or ext == ".bat" or ext == ".ps1" then
+            if ext == ".exe" or ext == ".cmd" or ext == ".bat" or ext == ".ps1" or ext == ".sh" then
```
waruqi: Simply adding .sh is not enough; many files without extensions are shell script files.

Therefore, we can only use os.isfile

Image

luadebug: >Therefore, we can only use os.isfile

If we could sort of use PIPE to capture output of streams would that help to filter out files without extensions shell script files? Such as core.base.pipe.lua thing. As PE file check would not work over shell scripts but might work over binary files without extension or with wrong extension, as for Windows OS example with PE file check we can figure out arch of executable and minimum Windows OS support (Windows XP) for example.

>The current rules for filtering file extensions may break some things when built on msys2. e.g. 7z.sh on msys2

That is probably same issue with shell script files without extensions I guess...

waruqi: try this 

```lua
function os.isexec(filepath)
    assert(filepath)

    if os.host() == "windows" then
        local exts = {".exe", ".cmd", ".bat", ".ps1", ".sh"}
        if os.isfile(filepath) then
            -- detect file extension first
            local extension = path.extension(filepath):lower()
            if extension then
                if table.contains(extension, exts) then
                    return true
                end
            end

            -- detect file header
            if winos.is_pefile(filepath) then
                return true
            end

            -- detect shebang scripts
            local file = io.open(filepath, "rb")
            if file then
                local header = file:read(2)
                file:close()
                if header == "#!" then
                    return true
                end
            end
        end
        for _, suffix in ipairs(exts) do
            if os.isfile(filepath .. suffix) then
                return true
            end
        end
    elseif os.isfile(filepath) then
        return os._access(filepath, "x")
    end
    return false
end
```

Then we can improve the previous parsepe native interface and change it to the winos.is_pefile interface implementation.

like this

```py
def is_pe_file(file_path):
    try:
        with open(file_path, 'rb') as f:
            # Step 1: Check the DOS header "MZ"
            mz_signature = f.read(2)
            if mz_signature != b'MZ':
                return False

            # Step 2: Seek to the PE header
            f.seek(60)  # DOS header e_lfanew offset location
            pe_offset = int.from_bytes(f.read(4), 'little')

            # Step 3: Check the PE signature "PE\0\0"
            f.seek(pe_offset)
            pe_signature = f.read(4)
            if pe_signature == b'PE\0\0':
                return True
    except Exception as e:
        print(f"Error reading file: {e}")
    return False

# Example usage
print(is_pe_file("example.exe"))
```

core/src/xmake/os/access.c
```diff
@@ -0,0 +1,25 @@
+#include "prefix.h"
```
waruqi, line 31: license comment header is missing. Please ensure that this coding style is consistent.

core/src/xmake/engine.c
```diff
@@ -480,6 +483,7 @@ static luaL_Reg const g_winos_functions[] = {
     { "registry_keys", xm_winos_registry_keys },
     { "registry_values", xm_winos_registry_values },
     { "short_path", xm_winos_short_path },
+    { "is_pe_file", xm_winos_is_pefile },
```
waruqi: is_pefile

waruqi:
I need to refactor binutils and add the binutils.format interface to replace is_pefile. 

please wait https://github.com/xmake-io/xmake/pull/7278

Issue #6927, opened 2025-10-15
The return value of os.shell() is not reliable


Willaaaaaaa:
### Xmake Version

3.0.3

### Operating System Version and Architecture

Windows 11 24H2

### Describe Bug

When you use powershell by **cmd -> vscode -> powershell**, `os.shell()` will tell you it's cmd. During this process, the environment variable `PROMPT` will be inherited, causing a wrong result from environment variable detection. Directly opening a powershell works as expected.
https://github.com/xmake-io/xmake/blob/2fe0526ccda961b178f8e46dc179edc387c3c95e/xmake/core/base/tty.lua#L161-L164

---

Opening a powershell terminal after opening vscode via cmd:
```pwsh
PS D:\Dev\auto-blue> xmake l
xmake> os.shell()
< "cmd"
```
Or, you can reproduce it by opening a cmd:
```cmd
C:\Users\Willaaaaaaa>powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

Install the latest PowerShell for new features and improvements! https://aka.ms/PSWindows

PS C:\Users\Willaaaaaaa> xmake l
xmake> os.shell()
< "cmd"
```

### Expected Behavior

```pwsh
PS C:\Users\Willaaaaaaa> xmake l
xmake> os.shell()
< "powershell"
```

### Project Configuration

None

### Additional Information and Error Logs

None

waruqi:
There is no reliable way to detect it right now, if you know it, you can directly improve it.

waruqi:
https://github.com/xmake-io/xmake/pull/7285

PR #7285, opened 2026-01-31
Improve Windows OS shell detection cmd/pwsh...

luadebug:
https://github.com/xmake-io/xmake/issues/6927
Adds process list winos function.
Sort of attempt to acquire shell from parent process ID.

### DIFF
--- a/core/src/xmake/os/processes.c
+++ b/core/src/xmake/os/processes.c
@@ -56,7 +56,7 @@ tb_int_t xm_winos_processes(lua_State* lua) {
 
                 // name
                 tb_char_t name[MAX_PATH * 4];
-                tb_size_t size = tb_charset_conv_data(TB_CHARSET_TYPE_UTF16 | TB_CHARSET_TYPE_LE, TB_CHARSET_TYPE_UTF8, (tb_byte_t const*)pe32.szExeFile, tb_wcslen(pe32.szExeFile) * sizeof(tb_wchar_t), (tb_byte_t*)name, sizeof(name));
+                tb_size_t size = tb_wtoa(name, pe32.szExeFile, sizeof(name));
                 if (size != -1) {
                     lua_pushlstring(lua, name, size);
                 } else {
@@ -70,7 +70,7 @@ tb_int_t xm_winos_processes(lua_State* lua) {
 
                 // ppid
                 lua_pushinteger(lua, (tb_int_t)pe32.th32ParentProcessID);
-                lua_setfield(lua, -2, "ppid");
+                lua_setfield(lua, -2, "parent_pid");
 
                 // result[i++] = entry
                 lua_rawseti(lua, -2, i++);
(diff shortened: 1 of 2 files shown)


core/src/xmake/os/processes.c
```diff
@@ -0,0 +1,86 @@
...
+
+        if (Process32FirstW(hSnapshot, &pe32)) {
+            tb_int_t i = 1;
+            do {
+                // new process entry table
+                lua_newtable(lua);
+
+                // name
+                tb_char_t name[MAX_PATH * 4];
+                tb_size_t size = tb_charset_conv_data(TB_CHARSET_TYPE_UTF16 | TB_CHARSET_TYPE_LE, TB_CHARSET_TYPE_UTF8, (tb_byte_t const*)pe32.szExeFile, tb_wcslen(pe32.szExeFile) * sizeof(tb_wchar_t), (tb_byte_t*)name, sizeof(name));
```
waruqi: use tb_wtoa, see other winos native apis.

core/src/xmake/os/processes.c
```diff
@@ -0,0 +1,86 @@
...
+                }
+                lua_setfield(lua, -2, "name");
+
+                // pid
+                lua_pushinteger(lua, (tb_int_t)pe32.th32ProcessID);
+                lua_setfield(lua, -2, "pid");
+
+                // ppid
+                lua_pushinteger(lua, (tb_int_t)pe32.th32ParentProcessID);
+                lua_setfield(lua, -2, "ppid");
```
waruqi: rename to parent_pid,

xmake/core/base/tty.lua
```diff
@@ -237,11 +237,53 @@ end
 
 -- find the shell from the parent process (linux)
 function tty._find_shell_from_parent()
+    local shell
+    if os.host() == "windows" then
```
waruqi: Encapsulate it into a separate function, for example: 

_find_shell_from_parent:
  -> _find_shell_from_parent_on_linux
  -> _find_shell_from_parent_on_windows

xmake/core/base/tty.lua
```diff
@@ -237,11 +237,53 @@ end
 
 -- find the shell from the parent process (linux)
```
waruqi: update comment , remove `(linux)`

xmake/core/base/tty.lua
```diff
@@ -237,11 +237,53 @@ end
...
+                        end
+                    end
+                    if shell then
+                        break
+                    end
+                    pid = process.ppid
+                end
+            end
+        end
+    end
```
waruqi, line 275: if winos.processes == nil then

fallback to 

```lua
        if not shell then
            local subhost = xmake._SUBHOST
            if subhost == "windows" then
                if os.getenv("PROMPT") then
                    shell = "cmd"
                else
                    local ok, result = os.iorun("pwsh -v")
                    if ok then
                        shell = "pwsh"
                    else
                        shell = "powershell"
                    end
                end
            end
        end
```

because some user will call `xmake update -s dev`to only update lua scripts without latest core binary.

PR #7293, opened 2026-02-05
feat: add support for running wasm target in browser

jinke18:
add support for running wasm target in browser

```
xmake f -p wasm
xmake b
xmake r
```

image

### DIFF
--- a/xmake/actions/run/main.lua
+++ b/xmake/actions/run/main.lua
@@ -40,16 +40,18 @@ function _run_wasm_target_in_browser(targetfile, opt)
     local setenvs = opt.setenvs
     local emrun = find_tool("emrun")
     if emrun then
-        os.execv(emrun.program, {targetfile}, {curdir = rundir, detach = option.get("detach"), addenvs = addenvs, setenvs = setenvs})
+        os.execv(emrun.program, {targetfile}, {
+            curdir = rundir, detach = option.get("detach"), addenvs = addenvs, setenvs = setenvs})
     else
         local python = find_tool("python3")
         if not python then
             raise("emrun or python not found, which is required for running wasm target in browser!")
         end
-        local url = "http://localhost:8000/" .. path.relative(targetfile, rundir):gsub("\\", "/")
+        local url = "http://localhost:8000/" .. path.unix(path.relative(targetfile, rundir))
         print("please open the url in browser")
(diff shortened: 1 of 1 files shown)


SirLynix:
Hello,

It would be a better idea to rely on the official emscripten test webserver `emrun` instead of Python which may not be available.

Plus it opens the browser automatically

jinke18:
emrun is used to execute WebAssembly targets, with Python serving as a fallback. @SirLynix

xmake/actions/run/main.lua
```diff
@@ -59,6 +59,26 @@ function _do_run_target(target)
         targetfile = wine.program
     end
 
+    -- run wasm target in browser
+    if target:is_plat("wasm") then
+        -- try to run with emrun
+        local emrun = find_tool("emrun")
```
waruqi: I tried it, it is not found on windows.

It could be `emrun.bat`, `emrun.ps1`, `emrun.py`, or the emrun shell script. Different platforms should use different script files.

Additionally, it is located in the upstream/emscripten subdirectory.

waruqi:
please rebase to dev

xmake/actions/run/main.lua
```diff
@@ -59,6 +59,26 @@ function _do_run_target(target)
         targetfile = wine.program
     end
 
+    -- run wasm target in browser
+    if target:is_plat("wasm") then
```
waruqi, line 85: Please encapsulate this into a separate function.

xmake/actions/run/main.lua
jinke18: 在不同系统下检测`emrun`的逻辑在`find_emrun.lua`里面处理了，与`find_emcc.lua`逻辑一致，应该问题不大。
如果需要改进，是不是同时需要更改`find_emcc.lua`和`find_emar.lua`?

The logic for detecting `emrun` in different systems has been processed in `find_ emrun. lua`, which is consistent with the logic of `find_ emcc. lua`, so there should be no problem.
If improvement is needed, do we need to change both `find_ emcc. lua` and `find_ emar. lua` at the same time?

waruqi: 查找问题是 find emsdk 导致，已经修了，得 rebase 下 

https://github.com/xmake-io/xmake/pull/7296

PR #7298, opened 2026-02-05
add initial implementation of foo and main for Windows DLL example

luadebug:
```
    [ 52%]: linking.release test_mingw_dll.exe
    "C:\\Program Files (x86)\\Microsoft Visual Studio\\2022\\BuildTools\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX64\\x64\\link.exe" -nologo -dynamicbase -nxcompat -machine:x64 -libpath:build\windows\x64\release /opt:ref /opt:icf foo.lib psapi.lib -out:build\windows\x64\release\test_mingw_dll.exe build\.objs\test_mingw_dll\windows\x64\release\src\main.c.obj /implib:build\windows\x64\release\test_mingw_dll.lib
    error: execv(C:\Users\lin\xmake-luadebug\projects\test_windows_links\build\windows\x64\release\test_mingw_dll.exe ) failed(-1073741515): system error 0xC0000135 (STATUS_DLL_NOT_FOUND).
    The application failed to start because the following DLLs were not found:
      - foo.dll
    Please check your PATH environment variable or copy the missing DLLs to the executable directory.
```

https://github.com/xmake-io/xmake/issues/7176

I am unsure my approach is correct.
Idea was printing them in console instead of using original handler that is up to MessageBox windows GUI.

If policy is turned on so we would acquire MessageBox with error of missing link, otherwise this MessageBox with error of missing link would not appear.

### DIFF
--- a/tests/projects/windows/windows_links/xmake.lua
+++ b/tests/projects/windows/windows_links/xmake.lua
@@ -6,7 +6,7 @@ target("foo")
     set_kind("shared")
     add_files("src/foo.c")
 
-target("test_mingw_dll")
+target("test_foo_dll_presence")
     set_kind("binary")
     add_files("src/main.c")
     add_deps("foo")
--- a/xmake/core/sandbox/modules/os.lua
+++ b/xmake/core/sandbox/modules/os.lua
@@ -340,34 +340,36 @@ function sandbox_os.exec(cmd, ...)
     end
 end
 
--- execute command with arguments list
 -- get missing dlls
-local function _get_missing_dlls(program)
-    local missing = {}
-    local program_dir = path.directory(program)
+function sandbox_os._get_missing_dlls(program)
+
+    -- check
+    assert(program)
+
+    -- get paths
     local pathenv = os.getenv("PATH") or ""
     local paths = path.splitenv(pathenv)
-    table.insert(paths, 1, program_dir)
+    table.insert(paths, 1, path.directory(program))
 
-    local function _find_dll(name)
+    -- find missing dlls
+    local missing = {}
+    local imports = binutils.deplibs(program) or {}
+    for _, dll in ipairs(imports) do
+        local found = false
         for _, p in ipairs(paths) do
-            if os.isfile(path.join(p, name)) then
-                return true
+            if os.isfile(path.join(p, dll)) then
+                found = true
+                break
             end
         end
-        return false
-    end
-
-    local imports = binutils.deplibs(program) or {}
-
-    for _, dll in ipairs(imports) do
-        if not _find_dll(dll) then
+        if not found then
              table.insert(missing, dll)
         end
     end
     return missing
 end
 
+-- execute command with arguments list
 function sandbox_os.execv(program, argv, opt)
 
     -- make program
@@ -418,7 +420,7 @@ function sandbox_os.execv(program, argv, opt)
         -- get errors
         if ok ~= nil then
             if ok == -1073741515 then -- 0xC0000135
-                local missing_dlls = _get_missing_dlls(program)
+                local missing_dlls = sandbox_os._get_missing_dlls(program)
                 if #missing_dlls > 0 then
                     errors = string.format("execv(%s) failed(%d): system error 0xC0000135 (STATUS_DLL_NOT_FOUND).\nThe application failed to start because the following DLLs were not found:\n  - %s\nPlease check your PATH environment variable or copy the missing DLLs to the executable directory.", cmd, ok, table.concat(missing_dlls, "\n  - "))
                 else


tests/projects/windows/windows_links/xmake.lua
```diff
@@ -0,0 +1,19 @@
+add_rules("mode.debug", "mode.release")
```
waruqi, line 1: please move to`tests/projects/windows/`

core/src/xmake/winos/set_error_mode.c
```diff
@@ -0,0 +1,7 @@
+#include "prefix.h"
```
waruqi, line 31: license header is missing

core/src/xmake/process/openv.c
```diff
@@ -130,6 +139,17 @@ tb_int_t xm_process_openv(lua_State *lua) {
...
         lua_pop(lua, 1);
 
+#ifdef TB_CONFIG_OS_WINDOWS
+        // save winos error mode?
+        lua_pushstring(lua, "winos_error_mode_gui");
+        lua_gettable(lua, 3);
+        if (lua_toboolean(lua, -1)) {
+             old_error_mode = SetErrorMode(0);
+             set_error_mode = tb_true;
+        }
```
waruqi: Do not modify os.execv and process.openv; they are generic interfaces, and we should not add platform-specific parameters.

waruqi: Additionally, you've already set the error mode in os.execv, so there's no need to add it again.

xmake/core/sandbox/modules/os.lua
```diff
@@ -339,6 +340,139 @@ function sandbox_os.exec(cmd, ...)
 end
 
 -- execute command with arguments list
+-- get missing dlls
+local function _get_missing_dlls(program)
```
waruqi: We shouldn't parse PE files within the OS; all binary file parsing should be handled within binutils.

Furthermore, the binutils.deplibs interface is already available.

tests/projects/windows/windows_links/xmake.lua
```diff
@@ -0,0 +1,19 @@
+add_rules("mode.debug", "mode.release")
+
```
waruqi, line 2: test_windows_links -> windows_links

tests/projects/windows/windows_links/xmake.lua
```diff
@@ -0,0 +1,19 @@
+add_rules("mode.debug", "mode.release")
+
+set_policy("run.gui_error_dialogs", true)
+
+target("foo")
+    set_kind("shared")
+    add_files("src/foo.c")
+
+target("test_mingw_dll")
```
waruqi: mingw?

tests/projects/windows/windows_links/src/main.c
```diff
@@ -0,0 +1,24 @@
+#include <windows.h>
+#include <psapi.h>
+#include <stdio.h>
+
+#ifdef _WIN32
```
waruqi: this example is only for windows now.

xmake/core/sandbox/modules/os.lua
```diff
@@ -339,6 +341,33 @@ function sandbox_os.exec(cmd, ...)
...
```
waruqi: Use the same code definition style as other OS-private functions.

xmake/core/sandbox/modules/os.lua
```diff
@@ -360,7 +389,24 @@ function sandbox_os.execv(program, argv, opt)
...
+        if winos.seterrormode then
+            old_mode = winos.seterrormode(0)
+        end
+    end
+
     local ok, errors = os.execv(program, argv, opt)
+
+    -- restore error mode
+    if old_mode then
+        winos.seterrormode(old_mode)
```
waruqi, line 409: I'm unsure whether the error mode will still be followed when the process exits later.

tests/projects/windows/windows_links/xmake.lua
luadebug: Well I am confused here as 0xC0000135 error code is probably MSVC only behavior... I am unsure such error MessageBox would appear on MinGW...

luadebug: image
Alrighty so this also works for MinGW on Windows host... I see error MessageBox present as well...

xmake/core/sandbox/modules/os.lua
luadebug: If we place `print(old_mode)` before this if statement it would say `old_mode = 3`. So it's still worth of doing I guess.

xmake/core/sandbox/modules/os.lua
luadebug: Done I guess? %_% https://github.com/xmake-io/xmake/pull/7298/commits/6007059c47a51d5847772fb68e357e4869ba85b3

waruqi:
https://github.com/xmake-io/xmake/pull/7302

Issue #7307, opened 2026-02-09
Support for preserving package source code


star-hengxing:
### Is your feature request related to a problem? Please describe.

In the early days of xmake, package source code was preserved by default after building. However, because some users felt it took up too much disk space, we stopped retaining it. After various discussions, currently only `debug` packages will keep the source code:
```lua
add_requires("zlib", {debug = true})
```

- In practice, many senior developers want to inspect library source code to look up APIs and their implementations; currently, they can only do this via a web browser or by cloning the repository manually.
- In the era of LLMs, preserving source code is also beneficial for AI agents to perform queries.

### Describe the solution you'd like

We can introduce a new policy, such as `package.keep.source`, to give users the option to preserve library source code and have build systems that support `compile_commands.json` generate it by default.

For example, when this policy is enabled, we could pass `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` along with the Ninja/Make generators to CMake.

If an unsupported build system or generator is used, we could consider integrating this project in the future: https://github.com/clice-io/catter

waruqi:
https://github.com/xmake-io/xmake/pull/7309

PR #7309, opened 2026-02-10
keep package source

waruqi:
https://github.com/xmake-io/xmake/issues/7307


luadebug:
Is the idea of issue is to keep source next to project directory but not next to Downloads folder?

waruqi:
> Is the idea of issue is to keep source next to project directory but not next to Downloads folder?

keep `~/.xmake/cache/packages/2602/s/xxx/version/source/`

PR #7322, opened 2026-02-14
hide configfile print behind verbose flag

qudix:
- side note: there should probably be a `vcprint` function like `vprint` but for color

### DIFF
--- a/xmake/actions/config/configfiles.lua
+++ b/xmake/actions/config/configfiles.lua
@@ -362,7 +362,9 @@ function _generate_configfile(srcfile, dstfile, fileinfo, targets, preprocessors
     end
 
     -- trace
-    cprint("generating %s ... ${color.success}${text.success}", srcfile)
+    if option.get("verbose") then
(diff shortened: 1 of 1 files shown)


waruqi:
Why? we shouldn't hide it.

qudix:
it prints every time, even when nothing about the config files or the configuration has changed. also the first cprint at the beginning is already hidden, and as this is just the confirmation that that one succeeded, it would make sense to have them together

qudix:
I do see now that the whole function is gated behind a `on_depends`, but I'm not sure why it thinks the source file is changing every time I run any `xmake` command, every when spamming it. i don't think I'm doing anything special?

image

waruqi:
you can debug https://github.com/xmake-io/xmake/blob/821545f6a78188e733b0abe84cdfcb6d07c9e082/xmake/modules/core/project/depend.lua#L96

qudix:
i did some further testing and it only stopped after i reboot my machine, haven't been able to reproduce since...

i'll close this for now and re-open if it's not just a fluke

PR #7326, opened 2026-02-15
add winos.file_signature function to retrieve file signature information

luadebug:
TODO that still does not work.

Moved here from https://github.com/tboox/tbox/pull/308

Resolves: https://github.com/xmake-io/xmake/issues/5629

Idea is that we can check for signature owner name GIGA-BYTE ... and if that is it and executable name is GCC.exe we should skip/filter out that .exe from find_tool("gcc")/find_gcc function at that case.

Currently I see:
image
Certainly mine wrapper over find_gcc/find_tool might be not correct... :/

image

image

Yet they should return void or either another GCC.exe. Maybe this would work or maybe not.

### DIFF
--- a/.github/workflows/haiku.yml
+++ b/.github/workflows/haiku.yml
@@ -48,8 +48,8 @@ jobs:
           RUN_PROBABILITY: ${{ vars.HAIKU_RUN_PROBABILITY || '0.2' }}
 
   build:
-    #needs: check
-    #if: needs.check.outputs.should-run == 'true'
+    needs: check
+    if: needs.check.outputs.should-run == 'true'
     runs-on: ubuntu-latest
 
     concurrency:
--- a/xmake/core/base/os.lua
+++ b/xmake/core/base/os.lua
@@ -202,6 +202,12 @@ end
 
 -- if tmpdir_root is a symbolic link, os.tmpdir() may return a path that differs
 -- from the path style returned by os.curdir() (e.g. on Haiku).
+--
+-- Using a consistent root path can avoid errors in relative path resolution.
+--
+-- e.g.
+-- tmpdir: /tmp/.xmake0/260217/ -> /boot/system/cache/tmp/.xmake0/260217
+-- curdir: /boot/system/cache/tmp/.xmake0/260217
 function os._resolve_tmpdir(tmpdir_root)
     if os.islink(tmpdir_root) then
         tmpdir_root = os.readlink(tmpdir_root) or tmpdir_root


core/src/cli/xmake.lua
```diff
@@ -29,13 +29,14 @@ target("cli")
...
     if is_plat("windows") then
-        add_syslinks("ws2_32", "advapi32", "shell32")
+        add_syslinks("ws2_32", "advapi32", "shell32", "wintrust", "crypt32")
         add_ldflags("/export:malloc", "/export:free", "/export:memmove")
     elseif is_plat("android") then
         add_syslinks("m", "c")
     elseif is_plat("macosx") and is_config("runtime", "luajit") then
         add_ldflags("-all_load", "-pagezero_size 10000", "-image_base 100000000")
     elseif is_plat("mingw") then
+        add_syslinks("wintrust", "crypt32")
```
luadebug:
```suggestion
```

luadebug:
OK I think it is ready to review.

luadebug:
CC: @waruqi 
1. Test, GCC.exe will be fetched first from PATH according to Windows OS behavior. But here it fetches second gcc.exe without GIGA-BYTE signature.
image

2. Test there is only GCC.exe with GIGA-BYTE signature so it should return nil.
image

luadebug:
Arch Linux fail is false positive here.

core/src/xmake/winos/file_signature.c
```diff
@@ -0,0 +1,244 @@
...
+#include <wintrust.h>
+#include <softpub.h>
+#include <wincrypt.h>
+
+/* //////////////////////////////////////////////////////////////////////////////////////
+ * types
+ */
+/// the file signature info type
+typedef struct __tb_file_signature_info_t
+{
```
waruqi: modify code style

```c
func {
  if () {
  } else {
  } 
}

File: xmake/core/sandbox/modules/import/lib/detect/find_program.lua [hunk 3]
```diff
@@ -40,6 +40,34 @@ local scheduler   = require("sandbox/modules/import/core/base/scheduler")
 -- do check
 function sandbox_lib_detect_find_program._do_check(program, opt)
 
+    -- avoid gcc.exe signed by GIGA-BYTE ref: https://github.com/xmake-io/xmake/issues/5629
+    if winos.file_signature and program:lower():endswith("gcc.exe") then
```
waruqi: please move to find_gcc.lua

luadebug: > please move to find_gcc.lua

How to do that...?

core/src/xmake/winos/file_signature.c
luadebug: > modify code style
> 
> ```c
> func {
>   if () {
>   } else {
>   } 
> }
> ```

Done?

core/src/xmake/winos/file_signature.c
```diff
@@ -0,0 +1,227 @@
...
+#include "prefix.h"
+#include <windows.h>
+#include <wintrust.h>
+#include <softpub.h>
+#include <wincrypt.h>
+
+/* //////////////////////////////////////////////////////////////////////////////////////
+ * types
+ */
+/// the file signature info type
```
waruqi: `///` -> `//`

core/src/xmake/winos/file_signature.c
```diff
@@ -0,0 +1,227 @@
...
+typedef struct __tb_file_signature_info_t {
+    /// is the file digitally signed?
+    tb_bool_t           is_signed;
+
+    /// is the signature valid and trusted by the OS?
+    tb_bool_t           is_trusted;
+
+    /// the name of the signer (e.g., "Microsoft Corporation")
```
waruqi:
```c
/*
 */
```

core/src/xmake/winos/file_signature.c
```diff
@@ -0,0 +1,227 @@
...
+static tb_bool_t tb_file_get_signature_info(tb_char_t const* filepath, tb_file_signature_info_t* info) {
+    // check
+    tb_assert_and_check_return_val(filepath && info, tb_false);
+
+    // init info
+    tb_memset(info, 0, sizeof(tb_file_signature_info_t));
+
+    // convert path
+    tb_wchar_t wide_path[TB_PATH_MAXN];
+    if (!tb_path_to_wchar(filepath, wide_path, TB_PATH_MAXN)) return tb_false;
```
waruqi: https://github.com/tboox/tbox/blob/e5236a8aa18b06789cf89859080a3c3392505301/src/tbox/platform/windows/file.c#L381

use tb_path_absolute_w, it will also handle long path.

core/src/xmake/winos/file_signature.c
```diff
@@ -0,0 +1,227 @@
...
+
+                        if (pCertContext) {
+                            tb_wchar_t wName[256] = {0};
+                            if (CertGetNameStringW(pCertContext,
+                                                   CERT_NAME_SIMPLE_DISPLAY_TYPE,
+                                                   0,
+                                                   NULL,
+                                                   wName,
+                                                   256)) {
+                                WideCharToMultiByte(CP_UTF8, 0, wName, -1, info->signer_name, sizeof(info->signer_name), NULL, NULL);
```
waruqi: tb_wtoa

xmake/core/sandbox/modules/import/lib/detect/find_program.lua
```diff
@@ -43,7 +43,7 @@ function sandbox_lib_detect_find_program._do_check(program, opt)
     -- do not attempt to run program? check it fastly
     if opt.norun then
         return os.isfile(program)
-    elseif opt.norunfile and path.is_absolute(program) and os.isfile(program) then
+    elseif not opt.check and opt.norunfile and path.is_absolute(program) and os.isfile(program) then
```
waruqi: please do not modify it.

xmake/core/sandbox/modules/import/lib/detect/find_program.lua
```diff
@@ -64,7 +64,15 @@ function sandbox_lib_detect_find_program._do_check(program, opt)
     elseif type(opt.check) == "table" then
         ok, errors = os.runv(program, opt.check, {envs = opt.envs, shell = opt.shell})
     else
-        ok, errors = sandbox.load(opt.check, program)
+        local ok_or_errors
+        ok, ok_or_errors = sandbox.load(opt.check, program)
```
waruqi: please do not modify it. it only check it once.

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -23,6 +23,47 @@ import("lib.detect.find_program")
...
+                                is_gigabyte = true
+                            end
+                            break
+                        end
+                    end
+                end
+            end
+        end
+        if is_gigabyte then
+            return false
```
waruqi: If the check fails, simply call raise.

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -23,6 +23,47 @@ import("lib.detect.find_program")
...
+        end
+    end
+    return true
```
waruqi: remove it

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -23,6 +23,47 @@ import("lib.detect.find_program")
 import("lib.detect.find_programver")
 import("core.cache.detectcache")
 
+-- check gigabyte gcc
+function _check_gigabyte_gcc(program)
+    -- avoid gcc.exe signed by GIGA-BYTE ref: https://github.com/xmake-io/xmake/issues/5629
+    if is_host("windows") then -- is_plat("windows", "mingw")
```
waruqi: remove unuse comments

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -23,6 +23,47 @@ import("lib.detect.find_program")
...
+        local is_gigabyte = false
+        if program:lower():endswith("gcc.exe") then
+            import("core.base.winos")
```
waruqi: move to header

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -23,6 +23,47 @@ import("lib.detect.find_program")
...
+                            return true
+                        end
+                    end
+                end
+                if path.is_absolute(program) then
+                    if check_signature(program) then
+                        is_gigabyte = true
+                    end
+                else
+                    local paths = path.splitenv(vformat("$(env PATH)"))
```
waruqi: remove this else-block. find_program will return the absolute path whenever possible.

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -54,6 +95,7 @@ end
 function main(opt)
     opt = opt or {}
     opt.norunfile = true
+    opt.check = _check_gigabyte_gcc
```
waruqi: rename to _check_gcc

xmake/core/sandbox/modules/import/lib/detect/find_program.lua
waruqi: `norunfile` always takes precedence over the `check` statement, even if a `check` statement exists.

This is because I've temporarily retained the `check` function for many `find_xxx` functions, even if they are no longer needed for the time being.

xmake/modules/detect/tools/find_gcc.lua
waruqi:
```
if is_host("windows") then
    opt.check = _check_gcc
else
    opt.norunfile = true
end

File: xmake/modules/detect/tools/find_gcc.lua [hunk 16]
```diff
@@ -23,6 +23,47 @@ import("lib.detect.find_program")
...
+                            break
+                        end
+                    end
+                end
+            end
+        end
+        if is_gigabyte then
+            return false
+        end
+    end
```
waruqi, line 39: Since you overridden the check function, if it's not Windows, then check is an empty implementation and it will always pass the gcc check, even if the gcc file doesn't exist.

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -53,7 +100,13 @@ end
 --
 function main(opt)
     opt = opt or {}
-    opt.norunfile = true
+    -- save the original check
+    opt.check_gcc = opt.check
```
waruqi: we need not to save it.

xmake/modules/detect/tools/find_gcc.lua
```diff
@@ -23,6 +23,53 @@ import("lib.detect.find_program")
...
+            return false
+        end
+
+        if not path.is_absolute(program) then
+            local paths = path.splitenv(vformat("$(env PATH)"))
+            if paths then
+                for _, p in ipairs(paths) do
+                    local prog = path.join(p, program)
+                    if os.isfile(prog) and check_signature(prog) then
+                        return false
```
waruqi: find_program do not use any return value of check. 

please use raise() instead of `return false`

see https://github.com/xmake-io/xmake/blob/821545f6a78188e733b0abe84cdfcb6d07c9e082/xmake/modules/detect/tools/find_x64dbg.lua#L54

waruqi:
Does it work now?

luadebug:
> Does it work now?

I can't test for now. I will look into that later maybe. Currently mine Oracle VirtualBox has broken and I am installing Windows 11 24H2 via virt-manager instead of Oracle VirtualBox. Sorry.

luadebug:
> Does it work now?

Well looks good to go...

image

waruqi:
?

luadebug:
> ?

I broke branch accidentally.

Issue #7328, opened 2026-02-17
Incremental build of Qt app on macOS takes too long


serhii-olendarenko-dxc:
### Xmake Version

v3.0.7+20260210

### Operating System Version and Architecture

macOS 15.5

### Describe Bug

I've created a simple app via `xmake create -t qtquick.app`.
The problem is that **every time** I make even the smallest change in QML, all the Qt dependencies get redeployed:
```console
> timeit { xmake build }
[ 27%]: compiling.qt.qrc src/qml.qrc
[ 40%]: linking.release qml-example
[ 54%]: generating.qt.app qml-example.app
[100%]: build ok, spent 196,991s
3min 17sec 196ms 908µs 500ns
```

### Expected Behavior

If the build is incremental, it should be incremental, i.e., rebuild only the things that have to be rebuilt.

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")

target("qml-example")
    add_rules("qt.quickapp")
    add_headerfiles("src/*.h")
    add_files("src/*.cpp")
    add_files("src/qml.qrc")
```

### Additional Information and Error Logs

The reason is that Xmake calls `macdeployqt` with `-always-overwrite` key. I'm not sure why and when it was added (or maybe Qt changed the behavior of this switch on their side), but this makes the development intolerable.

waruqi:
you can try this patch. https://github.com/xmake-io/xmake/pull/7329

I have no time and qt env to test it.

PR #7329, opened 2026-02-18
improve qt deploy for macapp

waruqi:
https://github.com/xmake-io/xmake/issues/7328

### DIFF
--- a/xmake/rules/qt/deploy/macosx.lua
+++ b/xmake/rules/qt/deploy/macosx.lua
@@ -166,7 +166,10 @@ function main(target, opt)
     end
 
     -- do deploy
-    local argv = {target_app, "-always-overwrite"}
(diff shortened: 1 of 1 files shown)


GooRoo:
While the build with this code is faster for sure, gemini is right: at least `Frameworks` and `Resources` in the app bundle are cleaned up every time.

waruqi:
> While the build with this code is faster for sure, gemini is right: at least `Frameworks` and `Resources` in the app bundle are cleaned up every time.

No, we have to remove the .app file and redeploy every time, otherwise it will affect many things, including symbolic links and installation files.

If some files are removed in add_installfiles/add_headerfiles, this can cause some unnecessary files to be incorrectly retained.

GooRoo:
Okay then.

PR #7333, opened 2026-02-20
add winos.file_signature function to retrieve file signature information

luadebug:
Attempt to continue working on https://github.com/xmake-io/xmake/pull/7326 as I accidentally broke branch :(

TODO that still does not work.

Moved here from https://github.com/tboox/tbox/pull/308

Resolves: https://github.com/xmake-io/xmake/issues/5629

Idea is that we can check for signature owner name GIGA-BYTE ... and if that is it and executable name is GCC.exe we should skip/filter out that .exe from find_tool("gcc")/find_gcc function at that case.

Currently I see:
image
Certainly mine wrapper over find_gcc/find_tool might be not correct... :/

image

image

Yet they should return void or either another GCC.exe. Maybe this would work or maybe not.

### DIFF
--- a/core/src/xmake/winos/file_signature.c
+++ b/core/src/xmake/winos/file_signature.c
@@ -37,6 +37,7 @@
 /* //////////////////////////////////////////////////////////////////////////////////////
  * types
  */
+
 // the file signature info type
 typedef struct __tb_file_signature_info_t {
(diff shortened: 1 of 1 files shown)


luadebug:
@waruqi LGTM
image

PR #7338, opened 2026-02-23
feat: add C# support for xmake

JassJam:
# Note
* This is just an initial implementation, stitched together from my old private projects, and use of my modified version of xmake, any feedback and requests are open, please feel free to test it out!
* This was only tested on windows platform for now.

# Inspiration
- #4510
- #6057 
- Some personal private projects (including [Ame](https://github.com/JassJam/Ame), my incomplete game engine).
- Here and there code stitches of some people from xmake's discord server.

# Motivation
Normally one could achieve it directly by just using rules + calling `dotnet` directly, but this was not enough for my greedy self :3, i wanted intellisense support without creating another sln file just for C#, unifying both C++ and C# into one sln file for ease of use and debugging!

# Prerequisites
* First and foremost, `dotnet` tool must be installed in order to use C# projects.
* C# projects require `.csproj` file (contains packages, framework settings, SDK, etc...), so `.csproj` must be part of the source set, and should not be auto generated by xmake, to not over-complicate it (just use `dotnet new ...`) .

# How it works
* Adds the `csharp` rule with `add_rules`.
* Explicitly registers `.csproj` and `.cs` files with `add_files`.
* Supports cross platform (`windows`, `linux`, `macos`) with cross architecture (`x64`, `x86`, `arm64`, `arm`, `riscv64`).


waruqi:
This patch is too large, and I don't have much time to review and test it. Please split several smaller patches if possible. like this:

- [ ] Support basic CLI build, (rules, language and tests)
- [ ] Support vsxmake generator
- [ ] Support vs generator

PR #7343, opened 2026-02-24
print exact verbose logs for Go build process

lewis-yeung:
This PR supersedes #7334. See my [comment][comment] in #7334. Looking through the codebase of xmake, I found a lot of existing `vprint` usages for friendly debugging, not just printing the build commands. Verbose logs are generally for debugging purposes, so it's better for users to know what are being done exactly during the build process.

[comment]: https://github.com/xmake-io/xmake/pull/7334#issuecomment-3939981882

### DIFF
--- a/xmake/rules/go/build/target.lua
+++ b/xmake/rules/go/build/target.lua
@@ -56,16 +56,13 @@ function build_sourcefiles(target, sourcebatch, opt)
     -- trace progress info
     progress.show(opt.progress, "${color.build.target}linking.$(mode) %s", path.filename(targetfile))
 
-    -- trace verbose info
-    vprint(compinst:buildcmd(sourcefiles, targetfile, {target = target, compflags = compflags}))
-
     -- flush io buffer to update progress info
     io.flush()
 
     -- build it (compile and link in one step for Go)
     -- Go builds packages, so we need to change to the source directory
-    local srcdir = path.directory(sourcefiles[1])
     local projectdir = os.projectdir()
+    local srcdir = path.absolute(path.directory(sourcefiles[1]), projectdir)
 
     -- convert targetfile to absolute path before changing directory
     -- targetfile is relative to project directory, so use projectdir as base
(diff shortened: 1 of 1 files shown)


waruqi:
As I mentioned before, verbose logs are not debug logs; they are typically used only to display critical compilation commands. Diagnosis logs are the true debug logs.

xmake/rules/go/build/target.lua
```diff
@@ -82,10 +79,26 @@ function build_sourcefiles(target, sourcebatch, opt)
         table.insert(rel_sourcefiles, rel_sourcefile)
     end
 
-    local oldir = os.cd(srcdir)
+    -- trace verbose info
+    vprint("changing working directory: %s", srcdir)
```
waruqi, line 83: please do not add these logs. If changing directories also needs to be output as logs, then the entire build will be filled with a lot of these logs, which I will not accept.

xmake/rules/go/build/target.lua
```diff
@@ -82,10 +79,26 @@ function build_sourcefiles(target, sourcebatch, opt)
...
-    os.cd(oldir)
+    local ok, errmsg = compinst:build(rel_sourcefiles, abs_targetfile, {target = target, dependinfo = dependinfo, compflags = compflags})
+
+    -- trace verbose info
+    vprint("restoring working directory: %s", original_dir)
+
+    -- restore the original working directory before asserting,
+    -- ensuring a consistent state for other tasks in the same process
+    -- even the build fails
+    os.cd(original_dir)
```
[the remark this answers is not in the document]
waruqi: We don't need to do this; all coroutine tasks will automatically save and restore curdir.

waruqi:
xmake will not print out every detail, which would impact compilation efficiency. Furthermore, most users don't care when they switch to which directory during the build process. If you need this information for debugging, you can temporarily add debug logs to xmake to analyze them.

xmake/rules/go/build/target.lua
lewis-yeung: > Diagnosis logs are the true debug logs.

So, let's just print these logs using `dprint`, should we?

lewis-yeung: @waruqi If you agree, could you please reopen this PR? I've updated my fork.

waruqi: If only print directory information, I think there is no need to use dprint too.

lewis-yeung: @waruqi I mean, for debugging purposes, most users will pay more attention to diagnostic logs for the execution details. For example, a user might manually execute the build commands printed by xmake for debugging purposes. However, because the working directory changes during the Go build process, simply executing the build commands can result in failure. So, providing these intuitive details can reduce confusion during debugging.

lewis-yeung: @waruqi I'm not refuting your point about diagnostic logs, but merely stating the rationale behind my revised PR. Whatever decision you make, thanks for your hard work on this project!

PR #7350, opened 2026-02-25
Implement logic add_deps with {public = false}

### LINKED ISSUE #7079: add_deps  support {public = false}

add_deps can have options like {public = false} which like add_includedirs and other API.

With this config:
```
target('a')
add_includedirs('./a_dir', {public = true})

target('b')
add_deps('a', {public = false})

target('c')
add_deps('b')
```
Target 'c' will not inherit includedir './a_dir' from target 'a'.

_No response_

### Additional context

_No response_

luadebug:
Resolves: https://github.com/xmake-io/xmake/issues/7079

**b.c（直接依赖 — 必须继承）：**
- `#include "*.h"` — 证明包含路径继承（否则无法编译）
- `#ifndef HAS_*` / `#error` — 证明 define 是继承的

**main.c（传递式依赖 — 必须不继承）：**
- `#if __has_include("*.h")` / `#error` — 证明包含路径**不**继承
- `#ifdef HAS_*` / `#error` — 证明定义**不是**继承的

| 场景 | A 类型 | B 类型 | b 得到 -I？ | b 得到 -D？ | c 得到 -I？ | c 得到 -D？ |
|------|--------|--------|-------------|-------------|-------------|-------------|
| 1    | 仅标题 | 静态   | `-Iheaderonly_dir` | `-DHAS_HEADERONLY_A=1` | (无) | (无) |
| 2    | 静态   | 静态   | `-Istatic_static_dir` | `-DHAS_SS_A=2` | (无) | (无) |
| 3    | 共享   | 静态   | `-Ishared_static_dir` | `-DHAS_DS_A=3` | (无) | (无) |
| 4    | 静态   | 共享   | `-Istatic_shared_dir` | `-DHAS_SD_A=4` | (无) | (无) |

### DIFF
--- a/xmake/core/project/target.lua
+++ b/xmake/core/project/target.lua
@@ -263,7 +263,21 @@ function _instance:_build_deps()
             if depinherit == nil then
                 depinherit = t:extraconf("deps", dep:fullname(), "inherit")
             end
-            return depinherit == nil or depinherit
+            if depinherit ~= nil and not depinherit then
+                return false
+            end
+            -- {public = false} only blocks transitive inheritance (not direct)
+            -- @see https://github.com/xmake-io/xmake/issues/7341
+            if t ~= self then
+                local deppublic = t:extraconf("deps", dep:name(), "public")
+                if deppublic == nil then
+                    deppublic = t:extraconf("deps", dep:fullname(), "public")
+                end
+                if deppublic ~= nil and not deppublic then
+                    return false
+                end
+            end
+            return true
         end)
     end
 end
--- a/xmake/core/project/target.lua
+++ b/xmake/core/project/target.lua
@@ -269,10 +269,7 @@ function _instance:_build_deps()
             -- {public = false} only blocks transitive inheritance (not direct)
             -- @see https://github.com/xmake-io/xmake/issues/7079
             if t ~= self then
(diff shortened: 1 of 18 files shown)


xmake/core/project/target.lua
```diff
@@ -263,7 +263,18 @@ function _instance:_build_deps()
...
-            return depinherit == nil or depinherit
+            if depinherit ~= nil and not depinherit then
+                return false
+            end
+            -- {public = false} only blocks transitive inheritance (not direct)
+            -- @see https://github.com/xmake-io/xmake/issues/7079
+            if t ~= self then
+                local deppublic = t:extraconf("deps", dep:name(), "public") or t:extraconf("deps", dep:fullname(), "public")
+                if deppublic ~= nil and not deppublic then
+                    return false
```
waruqi: This is no different from using `{inherit = false}`. 

The point of this issue is that deps themselves should also have visibility configurations such as private/public.

It does not control whether the visibility of defines/incudedirs is inherited.

In other words, `add_deps("xxx", {public = false})` will be a private dependency of the current target, and will not export any of its configuration, including links.

It also does not affect the dependency order of the top-level target compilation.

This is not something that can be easily achieved; there are many details to consider, so I haven't implemented it yet.

PR #7366, opened 2026-03-03
Improve targz and add tarxz pack format

waruqi:
https://github.com/xmake-io/xmake/issues/7361


luadebug:
Can we have .conda?
My bad, it seems already produced and working.

PR #7453, opened 2026-04-03
fix(xcode.application): bundle external dylibs in macOS apps

Akaps316:
## Summary

Bundle non-target macOS dylibs into `xcode.application` app bundles in
addition to shared target and framework dependencies.

This fixes app bundles that link external or package-provided `.dylib`
files and otherwise end up with a binary that depends on `@rpath/...`
entries that are never copied into `Contents/Frameworks`.

Fixes #7452.

## Changes

- keep explicit framework-target bundle copying
- reuse existing target/package shared library collection via `get_target_libfiles()`
- resolve declared shared `links` against merged `linkdirs` to find non-target dylibs
- walk the app binary dylib dependencies and copy non-system dylibs into `Contents/Frameworks`
- add a regression test that builds an external dylib and verifies it is bundled into a macOS app

## Why

`xcode.application` already performs partial app-bundle assembly by rewriting
rpath and copying shared target/framework dependencies.

However, it did not bundle non-target dylibs linked through raw
`add_linkdirs()` / `add_links()` usage, leaving the generated `.app`
incomplete relative to the app binary's load commands.

This patch makes that behavior consistent for macOS app bundles without
changing the separate duplicate-`LC_RPATH` issue, which is intentionally
left out of this change.

## Testing

- ran `build/xmake l tests/run.lua macapp_external_dylib`
- result: `1 test(s) succeed`
- smoke-built existing fixtures:
  - `tests/projects/objc/macapp_with_shared`
  - `tests/projects/objc/macapp_with_framework`

### DIFF
--- a/tests/projects/objc/macapp_external_dylib/test.lua
+++ b/tests/projects/objc/macapp_external_dylib/test.lua
@@ -1,0 +1,41 @@
+function _build_external_dylib()
+    local arch = os.arch()
+    local sdkdir = os.iorunv("xcrun", {"--sdk", "macosx", "--show-sdk-path"}):trim()
+    os.execv("xcrun", {
+        "--sdk", "macosx", "clang",
+        "-dynamiclib",
+        "-target", arch .. "-apple-macos",
+        "-isysroot", sdkdir,
+        "-install_name", "@rpath/libextfoo.dylib",
+        "-o", "ext/libextfoo.dylib",
+        "ext/extfoo.c"
+    })
+end
+
+function main(t)
+    if not is_host("macosx") then
+        return t:skip("wrong host platform")
+    end
+
+    local homedir = path.absolute("home")
+    os.setenv("HOME", homedir)
+    os.mkdir(homedir)
+    os.mkdir(path.join(homedir, ".xmake"))
+
+    _build_external_dylib()
+    local arch = os.arch()
+
+    local xmake = path.absolute(path.join(os.projectdir(), "build", "xmake"))
+    local xmake_program_dir = path.absolute(path.join(os.projectdir(), "xmake"))
+    os.setenv("XMAKE_PROGRAM_FILE", xmake)
+    os.setenv("XMAKE_PROGRAM_DIR", xmake_program_dir)
+
+    os.execv(xmake, {"f", "-p", "macosx", "-a", arch, "-c"})
+    os.execv(xmake, {"-vD"})
+
+    local appdir = path.join("build", "macosx", arch, "release", "demo.app", "Contents", "Frameworks")
+    local dylibfile = path.join(appdir, "libextfoo.dylib")
+    if not os.isfile(dylibfile) then
+        raise("missing external dylib in macOS app bundle: %s", dylibfile)
+    end
+end
--- a/xmake/rules/xcode/application/build.lua
+++ b/xmake/rules/xcode/application/build.lua
@@ -28,7 +28,7 @@ import("private.utils.target", {alias = "target_utils"})
 import("utils.binary.deplibs", {alias = "get_depend_libraries"})
 import("utils.progress")
 
-local function _is_non_system_dylib(libfile)
+function _is_non_system_dylib(libfile)
     return libfile and libfile:endswith(".dylib")
        and not libfile:startswith("/usr/lib/")
        and not libfile:startswith("/System/Library/")
--- a/xmake/rules/xcode/application/build.lua
+++ b/xmake/rules/xcode/application/build.lua
@@ -22,8 +22,42 @@
 import("core.base.option")
 import("core.theme.theme")
 import("core.project.depend")
+import("lib.detect.find_library")
 import("private.tools.codesign")
+import("private.utils.target", {alias = "target_utils"})
+import("utils.binary.deplibs", {alias = "get_depend_libraries"})
 import("utils.progress")
+
+local function _is_non_system_dylib(libfile)
+    return libfile and libfile:endswith(".dylib")
+       and not libfile:startswith("/usr/lib/")
+       and not libfile:startswith("/System/Library/")
+end
+
+local function _get_target_linkdirs(target)
+    local linkdirs = {}
+    for _, values in ipairs(table.wrap(target:get_from("linkdirs", "*"))) do
+        for _, linkdir in ipairs(table.wrap(values)) do
+            table.insert(linkdirs, path.absolute(linkdir))
+        end
+    end
+    return table.unique(linkdirs)
+end
+
+local function _get_target_linklibfiles(target)
+    local linkdirs = _get_target_linkdirs(target)
+    local libfiles = {}
+    for _, values in ipairs(table.wrap(target:get_from("links", "*"))) do
+        for _, link in ipairs(table.wrap(values)) do
+            local libinfo = find_library(link, linkdirs, {plat = target:plat(), kind = "shared"})
+            if libinfo then
+                table.insert(libfiles, path.join(libinfo.linkdir, libinfo.filename))
+            end
+        end
+    end
+    return table.unique(libfiles)
+end
+
 function main (target, opt)
 
     -- get app and resources directory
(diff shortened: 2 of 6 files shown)


tests/projects/objc/macapp_external_dylib/test.lua
```diff
@@ -0,0 +1,39 @@
...
+    local sdkdir = os.iorunv("xcrun", {"--sdk", "macosx", "--show-sdk-path"}):trim()
+    os.execv("xcrun", {
+        "--sdk", "macosx", "clang",
+        "-dynamiclib",
+        "-target", "arm64-apple-macos14.2",
+        "-isysroot", sdkdir,
+        "-install_name", "@rpath/libextfoo.dylib",
+        "-o", "ext/libextfoo.dylib",
+        "ext/extfoo.c"
+    })
```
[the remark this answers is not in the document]
Akaps316: Good point.

I updated the external dylib build step to use `os.arch()` and switched
the target triple to `arch .. "-apple-macos"` so the test is no longer
tied to `arm64` or a specific macOS SDK version.

The PR branch has been updated.

tests/projects/objc/macapp_external_dylib/test.lua
```diff
@@ -0,0 +1,39 @@
...
+
+    local xmake = path.absolute(path.join(os.projectdir(), "build", "xmake"))
+    local xmake_program_dir = path.absolute(path.join(os.projectdir(), "xmake"))
+    os.setenv("XMAKE_PROGRAM_FILE", xmake)
+    os.setenv("XMAKE_PROGRAM_DIR", xmake_program_dir)
+
+    os.execv(xmake, {"f", "-p", "macosx", "-a", "arm64", "-c"})
+    os.execv(xmake, {"-vD"})
+
+    local appdir = "build/macosx/arm64/release/demo.app/Contents/Frameworks"
```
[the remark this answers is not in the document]
Akaps316: Good catch.

I updated the test to use `os.arch()` for the xmake configure step and
changed the bundle path construction to `path.join(...)` so it no longer
assumes `arm64`.

The PR branch has been updated.

xmake/rules/xcode/application/build.lua
```diff
@@ -51,18 +85,39 @@ function main (target, opt)
...
+            end
+        end
+        for _, libfile in ipairs(table.unique(libfiles)) do
+            if not framework_targetfiles[path.absolute(libfile)] then
+                if not os.isdir(frameworksdir) then
+                    os.mkdir(frameworksdir)
                 end
+                os.vcp(libfile, frameworksdir)
             end
         end
```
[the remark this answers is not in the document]
Akaps316: Agreed.

I refactored this block to collect frameworks and dylibs first, create
`Contents/Frameworks` once if needed, and then perform the copy
operations. That keeps the behavior the same while separating collection
from filesystem work.

The PR branch has been updated.

xmake/rules/xcode/application/build.lua
```diff
@@ -22,8 +22,42 @@
 import("core.base.option")
 import("core.theme.theme")
 import("core.project.depend")
+import("lib.detect.find_library")
 import("private.tools.codesign")
+import("private.utils.target", {alias = "target_utils"})
+import("utils.binary.deplibs", {alias = "get_depend_libraries"})
 import("utils.progress")
+
+local function _is_non_system_dylib(libfile)
```
waruqi: `local function` -> `function`

Akaps316: Done, removed `local`.

waruqi:
This method is somewhat of a hack and doesn't guarantee that all dylibs will be reliably bundled into the bundle. 

The fact that macOS also includes system dylibs under /Library/ and the Xcode built-in dylib (/Applications/Xcode.app/) system libraries is completely unreliable.

Additionally, the extra `find_library` will impact build time. If users want to package and generate an app, they shouldn't use `add_linkdirs` to manually link third-party dynamic libraries. Use `add_deps` and `add_packages` whenever possible.

At least for now, I cannot accept the current implementation.

tests/projects/objc/macapp_external_dylib/test.lua
```diff
@@ -0,0 +1,41 @@
...
+    os.mkdir(homedir)
+    os.mkdir(path.join(homedir, ".xmake"))
+
+    _build_external_dylib()
+    local arch = os.arch()
+
+    local xmake = path.absolute(path.join(os.projectdir(), "build", "xmake"))
+    local xmake_program_dir = path.absolute(path.join(os.projectdir(), "xmake"))
+    os.setenv("XMAKE_PROGRAM_FILE", xmake)
+    os.setenv("XMAKE_PROGRAM_DIR", xmake_program_dir)
```
waruqi, line 31: here

tests/projects/objc/macapp_external_dylib/test.lua
```diff
@@ -0,0 +1,41 @@
...
+
+    os.execv(xmake, {"f", "-p", "macosx", "-a", arch, "-c"})
+    os.execv(xmake, {"-vD"})
```
waruqi, line 34: here

xmake/rules/xcode/application/build.lua
```diff
@@ -22,8 +22,42 @@
...
+import("utils.binary.deplibs", {alias = "get_depend_libraries"})
 import("utils.progress")
+
+function _is_non_system_dylib(libfile)
+    return libfile and libfile:endswith(".dylib")
+       and not libfile:startswith("/usr/lib/")
+       and not libfile:startswith("/System/Library/")
+end
+
+local function _get_target_linkdirs(target)
```
waruqi, line 37: here

xmake/rules/xcode/application/build.lua
```diff
@@ -22,8 +22,42 @@
...
+    local linkdirs = {}
+    for _, values in ipairs(table.wrap(target:get_from("linkdirs", "*"))) do
+        for _, linkdir in ipairs(table.wrap(values)) do
+            table.insert(linkdirs, path.absolute(linkdir))
+        end
+    end
+    return table.unique(linkdirs)
+end
+
+local function _get_target_linklibfiles(target)
```
waruqi, line 47: here

xmake/rules/xcode/application/build.lua
```diff
@@ -51,18 +85,46 @@ function main (target, opt)
...
+            local frameworkdir = dep:data("xcode.bundle.rootdir")
+            if dep:rule("xcode.framework") and frameworkdir then
+                table.insert(frameworks_to_copy, frameworkdir)
+                framework_targetfiles[path.absolute(dep:targetfile())] = true
+            end
+        end
+        local libfiles = {}
+        target_utils.get_target_libfiles(target, libfiles, target:targetfile(), {})
+        table.join2(libfiles, _get_target_linklibfiles(target))
+        local dependfiles = get_depend_libraries(target:targetfile(), {
```
waruqi, line 101: Please refer to the implementation details in the installation guide.

The `get_target_libfiles` function has already called `get_depend_libraries` internally. Then, `_get_target_linklibfiles` performs a very time-consuming `find_library` function, which then calls `get_depend_libraries` again.

xmake/rules/xcode/application/build.lua
```diff
@@ -51,18 +85,46 @@ function main (target, opt)
         try { function () os.vrunv("install_name_tool", {"-delete_rpath", "@loader_path", targetfile}) end }
         os.vrunv("install_name_tool", {"-add_rpath", "@executable_path/../Frameworks", targetfile})
 
-        -- copy dependent dynamic libraries and frameworks
+        -- copy dependent frameworks and dynamic libraries
+        local frameworks_to_copy = {}
+        local framework_targetfiles = {}
         for _, dep in ipairs(target:orderdeps()) do
-            if dep:kind() == "shared" then
```
waruqi, line 56: The handling of dependencies on non-framework dynamic libraries is also missing here.

Akaps316:
Understood.

The current approach is too heuristic and duplicates dependency scanning,
and I agree that automatically bundling dylibs from raw `add_linkdirs` /
`add_links` is not the right direction here.

I also see that this refactor is missing the clearer handling for direct
non-framework shared target dependencies.

I’m fine dropping this version rather than pushing the implementation
further in the wrong direction.

PR #7458, opened 2026-04-04
Fix Xcode toolchain ignoring local config precedence

Akaps316:
Fixes #7457

### Description
Updates `xmake/toolchains/xcode/check.lua` so that checking for the `xcode` configuration appropriately defaults first to the `toolchain:config("xcode")` dictionary before resorting to the global `config.get("xcode")`. An exported function `get_xcode()` is provided to handle this check.

Adds a regression toolchain test validating this exact requirement.

### DIFF
--- a/xmake/toolchains/xcode/check.lua
+++ b/xmake/toolchains/xcode/check.lua
@@ -26,6 +26,10 @@ import("private.utils.toolchain", {alias = "toolchain_utils"})
 import("private.utils.executable_path")
 import("private.tools.codesign")
 
+function get_xcode(toolchain)
+    return toolchain:config("xcode") or config.get("xcode")
+end
+
 -- print Xcode SDK summary (single line) and optional codesign infos
 function _show_checkinfo(toolchain, xcode, xcode_sdkver, target_minver)
     if xcode and xcode.sdkdir then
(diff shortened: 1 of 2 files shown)


xmake/toolchains/xcode/check.lua
```diff
@@ -26,6 +26,10 @@ import("private.utils.toolchain", {alias = "toolchain_utils"})
 import("private.utils.executable_path")
 import("private.tools.codesign")
 
+function get_xcode(toolchain)
+    return toolchain:config("xcode") or config.get("xcode")
```
waruqi, line 30: We should not retrieve `toolchain:config()` in `on_check`.

`on_check` is only called during `xmake f -c` and the initial check, and will never use `toolchain:config()`. In `on_check`, you should only set the configuration.

Akaps316: If we can't read toolchain:config() in on_check, what is the recommended way to support configuring a specific Xcode path via set_toolchains("xcode", {xcode = "..."})?

waruqi: you can use set_config, it works for me. you can also use `set_toolchains("xcode", {xcode = "..."})`

Akaps316: Understood. If `set_config("xcode", ...)` and `set_toolchains("xcode", {xcode = "..."})`
cover the intended configuration path here, then this change is not needed.

Thanks for clarifying.

PR #7481, opened 2026-04-11
normalize MinGW import library names on Linux/macOS to avoid linker errors

### LINKED ISSUE #9114: Check Linux => MinGW for case-sensitive syslink over existing xmake.lua packages
As there might be FFMPeg or gamenetworkingsockets are affected by case-sensitive lib linkage flag.

So it looks like Linux => MinGW is sort of case-sensitive during -l**ARG**.

**> checking for c links(minizip, z, crypt32, advapi32, Bcrypt)**

_Originally posted by @luadebug in https://github.com/xmake-io/xmake-repo/pull/9113#discussion_r2695791186_

luadebug:
Resolves: https://github.com/xmake-io/xmake-repo/issues/9114
Maybe it is better to allow it passing with existing uppercase symbols at first and if failed fallback to lowercased, yet I assume MacOS and Linux cross compilation has lowercase syslinks only.

### DIFF
--- a/xmake/modules/core/tools/gcc.lua
+++ b/xmake/modules/core/tools/gcc.lua
@@ -975,16 +975,16 @@ function link(self, objectfiles, targetkind, targetfile, flags, opt)
     end
 
     local program, argv = linkargv(self, objectfiles, targetkind, targetfile, flags, opt)
-    local envs = {envs = self:runenvs(), shell = opt.shell}
+    local run_opts = {envs = self:runenvs(), shell = opt.shell}
     local function _do_link(argv)
         if linker_output then
-            os.execv(program, argv, envs)
+            os.execv(program, argv, run_opts)
         else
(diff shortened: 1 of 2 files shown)


CyanoHao:
`ls mingw-w64-crt-pkg-root/lib | LC_ALL=C sort`  finds following capital names:

```
libCINTIME.a
libPS5UI.a
libPSCRIPT5.a
libUNIDRV.a
libUNIDRVUI.a
```

By the way, should we exclude Windows `... and not is_host("windows")`? There can be case sensitive paths on Windows.

```
fsutil.exe file setCaseSensitiveInfo <path> enable
```

waruqi:
This method is somewhat of a hack and difficult to maintain. It also doesn't support clang. I think there should be a better solution.

waruqi:
try this patch. https://github.com/xmake-io/xmake/pull/7483

PR #7483, opened 2026-04-12
fix syslink case for mingw

waruqi:
https://github.com/xmake-io/xmake-repo/issues/9114

### DIFF
--- a/xmake/core/package/package.lua
+++ b/xmake/core/package/package.lua
@@ -2127,6 +2127,18 @@ function _instance:fetch(opt)
         end
     end
 
+    -- normalize syslinks to lowercase for mingw on case-sensitive host filesystems,
+    -- because mingw import libraries are always lowercase (e.g. libbcrypt.a),
+    -- but windows api headers may declare mixed-case names (e.g. Bcrypt, Crypt32)
+    -- @see https://github.com/xmake-io/xmake/issues/9113
+    if fetchinfo and fetchinfo.syslinks and self:is_plat("mingw") and os.fscase() then
+        local syslinks = table.wrap(fetchinfo.syslinks)
+        for i, link in ipairs(syslinks) do
+            syslinks[i] = link:lower()
+        end
+        fetchinfo.syslinks = table.unwrap(syslinks)
+    end
+
     -- save to cache
     if usecache then
         self._FETCHINFO = fetchinfo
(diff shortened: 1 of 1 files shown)


luadebug:
https://github.com/xmake-io/xmake/pull/7481#issuecomment-4230476297
What do you think about this?

```lua
    if fetchinfo and fetchinfo.syslinks and self:is_plat("mingw") and os.fscase() then
        fetchinfo = table.clone(fetchinfo)
        local linkdirs = table.wrap(fetchinfo.linkdirs)
        local syslinks = table.wrap(fetchinfo.syslinks)
        local results = {}
        for i, link in ipairs(syslinks) do
            local lower = link:lower()
            if lower ~= link then
                -- only lowercase if no original-case import lib exists in the link dirs
                local found = false
                for _, linkdir in ipairs(linkdirs) do
                    if os.isfile(path.join(linkdir, "lib" .. link .. ".a")) or
                       os.isfile(path.join(linkdir, "lib" .. link .. ".dll.a")) then
                        found = true
                        break
                    end
                end
                link = found and link or lower
            end
            results[i] = link
        end
        fetchinfo.syslinks = table.unwrap(results)
    end

```

waruqi:
> [#7481 (comment)](https://github.com/xmake-io/xmake/pull/7481#issuecomment-4230476297) What do you think about this?
> 
> ```lua
>     if fetchinfo and fetchinfo.syslinks and self:is_plat("mingw") and os.fscase() then
>         fetchinfo = table.clone(fetchinfo)
>         local linkdirs = table.wrap(fetchinfo.linkdirs)
>         local syslinks = table.wrap(fetchinfo.syslinks)
>         local results = {}
>         for i, link in ipairs(syslinks) do
>             local lower = link:lower()
>             if lower ~= link then
>                 -- only lowercase if no original-case import lib exists in the link dirs
>                 local found = false
>                 for _, linkdir in ipairs(linkdirs) do
>                     if os.isfile(path.join(linkdir, "lib" .. link .. ".a")) or
>                        os.isfile(path.join(linkdir, "lib" .. link .. ".dll.a")) then
>                         found = true
>                         break
>                     end
>                 end
>                 link = found and link or lower
>             end
>             results[i] = link
>         end
>         fetchinfo.syslinks = table.unwrap(results)
>     end
> ```

`linkdirs` is only used for links, not syslinks. The link directories for system libraries are not in `fetchinfo`.

Issue #5721, opened 2024-10-15
modpost: "_printk" [<...>/<redacted>.ko.ko] undefined


TheAifam5:
### Xmake Version

v2.9.5+20241016

### Operating System Version and Architecture

Gentoo (AMD64)

### Describe Bug

While compiling a kernel module with "-v" (verbose) flag, modpost shows a warnings:

```
WARNING: modpost: "_printk" [build/.objs/<redacted>/linux/x86_64/release/build/linux/x86_64/release/<redacted>.ko.ko] undefined!
WARNING: modpost: "__x86_return_thunk" [build/.objs/<redacted>/linux/x86_64/release/build/linux/x86_64/release/<redacted>.ko.ko] undefined!
WARNING: modpost: "module_layout" [build/.objs/<redacted>/linux/x86_64/release/build/linux/x86_64/release/<redacted>.ko.ko] undefined!
```

### Expected Behavior

There should be no warning. The module compiles fine and runs.

### Project Configuration

xmake.lua:
```lua
option("linux-headers", { showmenu = true, description = "Set linux-headers path." })

target("<redacted>")
	add_rules("platform.linux.driver")
	add_files("src/*.c")
	set_values("linux.driver.linux-headers", "$(linux-headers)")
	set_license("GPL-2.0")
```

entry.c:
```c
#include <linux/init.h>   // Needed for the macros
#include <linux/kernel.h> // Needed for KERN_INFO
#include <linux/module.h> // Needed by all modules

MODULE_LICENSE("GPL-2.0");
MODULE_AUTHOR("<redacted> <redacted>");
MODULE_DESCRIPTION("<redacted>");

static int __init dw_init(void) {
  printk(KERN_INFO "<redacted>: Hello, World!\n");
  return 0;
}

static void __exit dw_exit(void) {
  printk(KERN_INFO "<redacted>: Goodbye, World!\n");
}

module_init(dw_init);
module_exit(dw_exit);
```

### Additional Information and Error Logs

```
[ 50%]: cache compiling.release dw-kmod/src/entry.c
[...]

PR #7501, opened 2026-04-23
fix linux module intermediate filenames

ririyeye:
## Summary
- generate linux kernel module intermediate files from the target basename instead of the final `.ko` path
- make `modpost` consume `hello.o` instead of `hello.ko.o`
- avoid the bogus `*.ko.ko` module name and undefined symbol warnings reported in #5721

## Root cause
`platform.linux.module` previously called `target:objectfile(targetfile)` where `targetfile` already ended with `.ko`. That produced intermediate paths like `hello.ko.o`, which then flowed into `modules.order`/`modpost` and could be rendered as `hello.ko.ko` in warnings.

This patch switches the intermediate basename to `target:basename()`, so the generated files become `hello.o`, `hello.mod`, `hello.mod.c`, and `hello.mod.o`.

## Test plan
- [x] verified `git diff master...HEAD` only contains this single fix commit
- [x] confirmed the generated intermediate names on this branch are based on the module basename, not the final `.ko` path
- [x] attempted to rebuild `tests/projects/linux/driver/hello_custom` against `/lib/modules/$(uname -r)/build`; the local machine currently hits an existing GCC 15 / kernel headers compatibility issue (`false`/`bool` in kernel headers) before the `modpost` stage, so it cannot be used here to fully exercise the warning path

Made with [Cursor](https://cursor.com)

### DIFF
--- a/xmake/rules/platform/linux/module/driver_modules.lua
+++ b/xmake/rules/platform/linux/module/driver_modules.lua
@@ -361,14 +361,15 @@ function link(target, opt)
         if ldflags_o then
             table.join2(argv, ldflags_o)
         end
-        local targetfile_o = target:objectfile(targetfile)
+        local targetfile_base = path.join(path.directory(target:objectfile(targetfile)), target:basename())
+        local targetfile_o = targetfile_base .. ".o"
         table.join2(argv, "-o", targetfile_o)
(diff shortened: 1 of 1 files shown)


waruqi:
ci test has been skipped. please push to dev branch.

PR #7523, opened 2026-05-02
Fix msvc/intel/snippet detect when temp path contains spaces

DavidWang19:
## Summary

- MSVC, Intel oneAPI (icl/icx/ifort/ifx) and `tryrun` C/C++/Fortran snippet detection fail on Windows when the user profile name contains a space (e.g. `C:\Users\Some Name\...`).
- With `xmake f --toolchain=msvc` (or `--toolchain=clang-cl`) the user gets a "toolchain not found" error; with plain `xmake f` xmake silently falls back to MinGW and prints no warning.
- Fix: at every affected call site, switch from `os.run` / `os.iorun` / `os.vrun` to the corresponding `*v` variant, so the temp file path is no longer parsed by `os.argv`.

## Root cause

The detection scripts in `xmake/modules/detect/sdks/` write a generated `.bat` under `os.tmpfile()` (which on Windows lives under the user profile) and run it via `os.iorun(genvcvars_bat)` or `os.run(...)`. These wrappers parse their `cmd` through `os.argv`, a POSIX-shell-style word splitter (`core/src/xmake/os/argv.c:83`). Without surrounding quotes the path gets shredded:

```
"C:\Users\Some Name\AppData\Local\Temp\.xmake\...\_genvcvars.bat"
    -> ["C:\Users\Some", "Name\AppData\...\_genvcvars.bat"]
```

`CreateProcess` then runs against the directory `C:\Users\Some` and fails with `ERROR_BAD_EXE_FORMAT (193)`. In `find_vstudio` the failure is swallowed by the surrounding `try {}` and turned into an empty `vcvarsall = {}`. The same root cause hits Intel oneAPI detection on Windows (`find_iccenv` / `find_icxenv` / `find_ifortenv` / `find_ifxenv`) and any `tryrun` feature probe in `check_cxsnippets` / `check_fcsnippets` (cross-platform but rarely tripped on POSIX).

## Fix

Switch every site that passes a single bare path to its `*v` variant. xmake's C-side `process._openv` already knows how to launch a `.bat` directly, so no `cmd.exe` wrapping is needed.

```diff
- local outdata, errdata = try {function () return os.iorun(genvcvars_bat) end}
+ local outdata, errdata = try {function () return os.iorunv(genvcvars_bat) end}
```

```diff
- os.run(geniclvars_bat)
+ os.runv(geniclvars_bat)
```

```diff
- local output = os.iorun(binaryfile)
+ local output = os.iorunv(binaryfile)
...
- os.vrun(binaryfile)
+ os.vrunv(binaryfile)
```

A short `@note` comment was added at each site so the choice does not get reverted later.

## Test plan

Verified on Windows 11 with a user profile path containing a space, with VS 2026 Community installed.

`xmake lua detect.sdks.find_vstudio` after the fix:

```json
{
  "2026" = {
    version = "18.0",
    vcvarsall_bat = "C:\Program Files\Microsoft Visual Studio\18\Community\VC\Auxiliary\Build\vcvarsall.bat",
    vcvarsall = {
      arm64   = { VCToolsVersion = "14.50.35717", PATH = "...", LIB = "...", INCLUDE = "...", ... },
      arm64ec = { ... },
      x64     = { ... },
      x86     = { ... },
      arm     = { ... }
    }
  }
}
```

Before the fix `xmake lua detect.sdks.find_vstudio` shows

```
{
  "2026" = {
    version = "18.0",
    vcvarsall_bat = "C:\Program Files\Microsoft Visual Studio\18\Community\VC\Auxiliary\Build\vcvarsall.bat",
    vcvarsall = { }
  }
}
```

### DIFF
--- a/xmake/modules/detect/sdks/find_iccenv.lua
+++ b/xmake/modules/detect/sdks/find_iccenv.lua
@@ -55,7 +55,7 @@ function _load_iclvars(iclvars_bat, arch, opt)
     -- run geniclvars.bat
     -- @note we use runv here so the bat path is not split on whitespace by os.argv,
     -- which breaks detection when the temp file lives under a path with spaces.
-    os.runv(geniclvars_bat, {})
+    os.runv(geniclvars_bat)
 
     -- load all envirnoment variables
     local variables = {}
--- a/xmake/modules/detect/sdks/find_iccenv.lua
+++ b/xmake/modules/detect/sdks/find_iccenv.lua
@@ -53,7 +53,9 @@ function _load_iclvars(iclvars_bat, arch, opt)
     file:close()
 
     -- run geniclvars.bat
-    os.run(geniclvars_bat)
+    -- @note we use runv here so the bat path is not split on whitespace by os.argv,
+    -- which breaks detection when the temp file lives under a path with spaces.
+    os.runv(geniclvars_bat, {})
 
     -- load all envirnoment variables
     local variables = {}
--- a/xmake/modules/detect/sdks/find_icxenv.lua
+++ b/xmake/modules/detect/sdks/find_icxenv.lua
@@ -55,7 +55,7 @@ function _load_icxvars(icxvars_bat, arch, opt)
     -- run genicxvars.bat
     -- @note we use runv here so the bat path is not split on whitespace by os.argv,
     -- which breaks detection when the temp file lives under a path with spaces.
-    os.runv(genicxvars_bat, {})
+    os.runv(genicxvars_bat)
 
     -- load all envirnoment variables
     local variables = {}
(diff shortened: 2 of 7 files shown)


waruqi:
?

xmake/modules/detect/sdks/find_iccenv.lua
```diff
@@ -53,7 +53,9 @@ function _load_iclvars(iclvars_bat, arch, opt)
     file:close()
 
     -- run geniclvars.bat
-    os.run(geniclvars_bat)
+    -- @note we use runv here so the bat path is not split on whitespace by os.argv,
+    -- which breaks detection when the temp file lives under a path with spaces.
+    os.runv(geniclvars_bat, {})
```
waruqi: `, {}` is optional

DavidWang19: Refactored.

Issue #7522, opened 2026-05-02
set_arch("x86") does not work with lld-link


na1307:
### Xmake Version

xmake v3.0.8+HEAD.4f45b5a6b, A cross-platform build utility based on Lua

### Operating System Version and Architecture

Windows 11 Build 26100.8246

### Describe Bug

set_arch("x86") works well with the default settings. However, if you set it to use lld-link as the linker, it does not work because it attempts to link targeting x64.

### Expected Behavior

When set_arch("x86") was set, it should have linked to the x86 target correctly even when using lld-link.

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")
set_defaultmode("debug")
set_plat("windows")
set_arch("x86")
--set_toolchains("clang-cl")
set_languages("c17", "cxx23")

target("test")
    add_files("main.cpp")
    set_toolset("ld", "lld-link")

```

### Additional Information and Error Logs

```console
PS E:\Project\asdf> xmake -v
checking for platform ... windows (x64)
checking for cl.exe ... C:\Program Files\Microsoft Visual Studio\18\Community\VC\Tools\MSVC\14.50.35717\bin\HostX64\x64\cl.exe
checking for Microsoft C/C++ Compiler (x64) ... ok
checking for Microsoft Visual Studio (x64) version ... 2026
checking for zig ... no
checking for nim ... no
checking for dotnet ... C:\Program Files\dotnet\dotnet.exe
checking for .NET SDK version ... 10.0.203
checking for .NET SDK directory ... C:\Program Files\dotnet\sdk\10.0.203
checking for cl.exe ... C:\Program Files\Microsoft Visual Studio\18\Community\VC\Tools\MSVC\14.50.35717\bin\HostX64\x86\cl.exe
checking for Microsoft C/C++ Compiler (x86) ... ok
checking for Microsoft Visual Studio (x86) version ... 2026
checking for zig ... no
checking for nim ... no
checking for clang-cl.exe ... C:\Program Files\LLVM\bin\clang-cl.exe
checking for LLVM Clang C/C++ Compiler (x86) ... ok
checking for Microsoft Visual Studio (x86) version ... 2026
checking for clang-cl ... C:\Program Files\LLVM\bin\clang-cl
checking for the c++ compiler (cxx) ... clang-cl
checking for link.exe ... C:\Program Files\Microsoft Visual Studio\18\Community\VC\Tools\MSVC\14.50.35717\bin\HostX64\x86\link.exe
[...]

PR #7524, opened 2026-05-03
add lld-link config

waruqi:
https://github.com/xmake-io/xmake/issues/7522

### DIFF
--- a/xmake/toolchains/clang-cl/load.lua
+++ b/xmake/toolchains/clang-cl/load.lua
@@ -28,19 +28,28 @@ import("private.utils.toolchain", {alias = "toolchain_utils"})
 function main(toolchain)
 
     -- set toolset
+    -- we can use `clang-cl[llvm]` to switch to llvm tools.
+    local use_llvm = toolchain:config("llvm")
     toolchain:set("toolset", "cc",      "clang-cl")
     toolchain:set("toolset", "cxx",     "clang-cl")
-    toolchain:set("toolset", "mrc",     "rc.exe")
     toolchain:set("toolset", "dlltool", "llvm-dlltool")
-    if toolchain:is_arch("x64") then
-        toolchain:set("toolset", "as",  "ml64.exe")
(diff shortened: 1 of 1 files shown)


luadebug:
Can you also check if that going to help for https://github.com/aui-framework/xmake_example_app/blob/master/xmake.lua#L38-L47 so I won't have to place workaround **if clang-cl**.

```lua
        -- Workaround for clang-cl toolchain bug where /WHOLEARCHIVE is ignored in shared library linking
         if is_plat("windows") then
            import("core.tool.toolchain")
            local host_toolchain
            host_toolchain = toolchain.load("clang-cl", {plat = "windows", arch = os.arch()})
            if host_toolchain:check() then
                target:add("ldflags", "/WHOLEARCHIVE:aui.views.lib", { force = true })
                target:add("ldflags", "/WHOLEARCHIVE:aui.xml.lib", { force = true })
                target:add("ldflags", "/WHOLEARCHIVE:aui.image.lib", { force = true })
                target:add("ldflags", "/WHOLEARCHIVE:aui.core.lib", { force = true })
                target:add("shflags", "/WHOLEARCHIVE:aui.views.lib", { force = true })
                target:add("shflags", "/WHOLEARCHIVE:aui.xml.lib", { force = true })
                target:add("shflags", "/WHOLEARCHIVE:aui.image.lib", { force = true })
                target:add("shflags", "/WHOLEARCHIVE:aui.core.lib", { force = true })
            end
        end
```

waruqi:
> https://github.com/aui-framework/xmake_example_app/blob/master/xmake.lua#L38-L47

This is unrelated to the current patch; clang-cl inherits from cl.lua by default and doesn't handle add_linkgroups.

You can improve clang_cl.lua to add it, like this: https://github.com/xmake-io/xmake/blob/d76ec6af28e29921a986662f1d823a3a3a372f4b/xmake/modules/core/tools/gcc.lua#L707

PR #7525, opened 2026-05-03
Fix zig cc stdlib and lto flags

tgsong:
This PR addresses the following issues within the `zigcc` toolchain,

- Fix warnings on `-stdlib=` flags. zig cc does not take this flag, it always uses the bundled libc++.
- Support `build.optimization.lto` policy, expands the check to include zig cc.

### DIFF
--- a/xmake/core/package/package.lua
+++ b/xmake/core/package/package.lua
@@ -2426,9 +2426,9 @@ function _instance:_generate_lto_configs(sourcekind)
     if sourcekind then
         local _, cc = self:tool(sourcekind)
         local cflag = sourcekind == "cxx" and "cxxflags" or "cflags"
-        if cc == "cl" then
+        if cc == "cl" or cc == "clang_cl" then
             configs[cflag] = "-GL"
-        elseif cc == "clang" or cc == "clangxx" or cc == "clang_cl" then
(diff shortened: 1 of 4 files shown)


xmake/modules/core/tools/gcc.lua
```diff
@@ -861,11 +857,7 @@ function nf_pcxxheader(self, pcheaderfile, opt)
     if self:kind() == "cxx" then
         local target = opt.target
         local pcoutputfile = target:pcoutputfile("cxx")
-        if self:name():startswith("clang") then
-            return {"-include", pcheaderfile, "-include-pch", pcoutputfile}
-        else
-            return {"-I", path.directory(pcoutputfile), "-include", path.filename(pcheaderfile)}
-        end
+        return {"-I", path.directory(pcoutputfile), "-include", path.filename(pcheaderfile)}
```
waruqi: Please do not modify anything unrelated to zig.

waruqi:
Ensure minimal changes; each patch should handle only one problem.

tgsong:
> Ensure minimal changes; each patch should handle only one problem.

Refactored GCC pch change into a separate PR.

PR #7534, opened 2026-05-08
feat: include packages in "build.merge_archive"

### LINKED ISSUE #7531: Include links from `add_packages` in `set_policy("build.merge_archive", true)`.

Hi,

I'm trying to build a static library `a` that has some dependency `b` added with `add_packages`. I then want to use static library `a` in other projects which don't use xmake, and without having to install `b`. In fact, this is partly the reason for why `a` exists in the first place; `b` is a bit difficult to install with the other build system.

What I'm effectively trying to do is including `libb.a` in `liba.a` so that `ar -t liba.a` shows both `a.o` and `b.o`.

After digging around previous issues and docs, `set_policy("build.merge_archive", true)` seems to be the closest alternative, ([example](https://xmake.io/examples/cpp/merge-static-libraries.html#automatically-merge-target-libraries)).

Problem is that it only seems to work with `add_deps` and not `add_packages`. Perhaps it is possible to adapt it to also work with packages? All the information seems to be there with `xmake show -t a`. I just can't figure it out myself how to get it to work from looking at the `utils.merge.archive` rule.

Also tried with `set_policy("package.merge_staticlibs", true)`, but it too seems to only work with `add_deps`.

gibbz00:
Fixes #7531 

POC that has some unanswered question left, hence the draft. Among them are:

- Make it a separate policy? (Sorry, I didn't see your suggestion in the issue discussion until now.) 
- Keep previous behavior and only make merging packages opt-in?
- Leave out syslinks and sysdirs? 
- Adding tests.
  - Asserts that packaged links are present in `ar -t`
  - Asserts that the respective links aren't inherited.
  - Inverse applied when policy is disabled.

### DIFF
--- a/xmake/rules/utils/merge_archive/xmake.lua
+++ b/xmake/rules/utils/merge_archive/xmake.lua
@@ -26,6 +26,8 @@ rule("utils.merge.archive")
         --
         -- @see https://github.com/xmake-io/xmake/issues/3404
         if target:policy("build.merge_archive") then
+            target:data_set("inherit.links.export_static", false)
+
             for _, dep in ipairs(target:orderdeps()) do
                 if dep:is_static() then
                     dep:data_set("inherit.links.deplink", false)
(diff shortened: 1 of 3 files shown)


xmake/rules/utils/merge_archive/xmake.lua
```diff
@@ -60,6 +76,15 @@ rule("utils.merge.archive")
                         table.insert(libraryfiles, dep:targetfile())
                     end
                 end
+
+                -- TODO: what about syslinkdirs and syslinks?
```
waruqi, line 80: see https://github.com/xmake-io/xmake/blob/c14b01fd77d4d31376f03263e9833deadb998e4e/xmake/modules/private/action/require/impl/actions/install.lua#L251

waruqi:
https://github.com/xmake-io/xmake/pull/7542

gibbz00:
Perfect, thanks!

PR #7549, opened 2026-05-16
feat: support building multiple targets

bionicbeagle:
Allow `xmake build target1 target2 ...` to build several named targets in a single invocation. Previously the `target` argument only accepted a single value.

The build pipeline already iterated target-name lists internally (get_root_targets and the post-build check both use table.wrap), so the only blockers were the CLI option declaration and a single-name validation call:

- build/xmake.lua: change the `target` argument from "v" (single value) to "vs" (value list), and add usage examples to --help.
- build/main.lua: validate each given name via check_targetname instead of assuming a single string.
- build/main.lua: in the no-xmake.lua third-party (_try_build) path, clean only the first given target, since that path has no concept of multiple xmake targets and trybuild never reads the target name anyway.

Single-target builds, the default (no-arg) build, --rebuild, --shallow, --group and --files all keep their existing behavior; programmatic callers that pass a single string (e.g. `xmake run` -> task.run("build", {target=...})) are unaffected because everything downstream normalizes via table.wrap.

------

### DIFF
--- a/xmake/actions/build/main.lua
+++ b/xmake/actions/build/main.lua
@@ -61,7 +61,10 @@ function _try_build()
     -- rebuild it? do clean first
     local targetname = option.get("target")
     if option.get("rebuild") then
-        task.run("clean", {target = targetname})
+        -- the third-party buildsystem path does not support multiple targets,
+        -- so just clean the first given target (if any)
+        local cleanname = table.wrap(targetname)[1]
+        task.run("clean", {target = cleanname})
     end
 
     -- get the buildsystem tool
@@ -201,7 +204,9 @@ function main(opt)
 
     -- check target name
     if targetname then
-        assert(check_targetname(targetname))
+        for _, name in ipairs(table.wrap(targetname)) do
+            assert(check_targetname(name))
+        end
     end
 
     -- enter project directory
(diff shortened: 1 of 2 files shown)


waruqi:
不考虑，当前的约定就是构建单个  target ，这里行为一改，所有 Actions 都受影响，包括对 install uninstall clean 里，targets 参数行为就不一致了

另外，先提 feature request 评估，不要上来就 pr 

---

Ignoring this, the current convention is to build a single target. Changing this behavior here will affect all Actions, including those in `install`, `uninstall`, and `clean`, causing inconsistent behavior for the `targets` parameter.

Also, submit a feature request for evaluation first, don't immediately submit a pull request.

bionicbeagle:
ok, sorry. The reason for attempting to improve this was because I commonly see agents attempting to build multiple targets frequently and there seems to be no particular reason why it shouldn't be possible. If you have a better suggestion for how this might be implemented I'd be happy to pick it up.

waruqi:
you can use set_group and `xmake build -g group` to build multiple targets.

bionicbeagle:
Thanks, yeah I get that but it requires foresight. Where I've seen this come up is when an agent tries to be surgical about which projects need to be built/verified. Many build systems allow multiple projects to be specified, so that's what they expect to be able to do. They try, fail, and then realise and build one project at a time. So mostly this was an attempt to not have to wait for it to waste time on discovering this over and over.

You can obviously try and guide the agents by informing them that they can only build one project at a time, but they forget and eventually revert back to their "instincts" as it were. I guess I'll just do that.

waruqi:
try this patch. https://github.com/xmake-io/xmake/pull/7607

PR #7558, opened 2026-05-22
Introduced check_alignof support like the one for check_sizeof

### LINKED ISSUE #7557: `alignof` support like `sizeof` to detect types alignment

Modern versions of C and C++ allow the user to specify alignment for fields in a struct.  
The final alignment cannot be trivially assumed, but we don't have this detection feature in xmake.

An integration mirroring the one of sizeof by using alignof

_No response_

### Additional context

_No response_

KaruroChori:
As described in the title.
It closes https://github.com/xmake-io/xmake/issues/7557


xmake/includes/check/check_alignof.lua
```diff
@@ -0,0 +1,155 @@
...
+-- e.g.
+--
+-- check_alignof("ALIGNOF_LONG", "long") => ALIGNOF_LONG=4
+--
+-- configvar_check_alignof("ALIGNOF_LONG", "long") => #define ALIGNOF_LONG 4
+--
+local binary_match_pattern = 'INFO:align%[(%d+)%]'
+
+local check_alignof_template = [[
+#define ALIGN (alignof(${TYPE}))
```
waruqi: Not all compilers and languages support the `alignof` keyword.

We should do more testing

like this 

```c
#if defined(__cplusplus)
    #if __cplusplus >= 201103L
        #define ALIGNOF(type) alignof(type)
    #else
        #define ALIGNOF(type) __alignof(type)
    #endif
#elif defined(_MSC_VER)
    #define ALIGNOF(type) __alignof(type)
#elif defined(__GNUC__) || defined(__clang__)
    #define ALIGNOF(type) __alignof__(type)
#else
    #define ALIGNOF(type) _Alignof(type)
#end 
```

KaruroChori: Right, I forgot C23 has `alignof`, but it was introduced before in C11 as `_Alignof`, which is now deprecated.
I would suggest: 

- `alignof` for version >= C++11 or version >= C23
- `_Alignof` for  version >= C11 and version < C23
- A best effort attempt at matching compiler specific implementations for earlier versions of C++ and C

Issue #7556, opened 2026-05-21
nuget包引入存在问题


ChrisCatCP:
### Xmake 版本

3.0.9

### 操作系统版本和架构

win11

### 描述问题

`nuget::Microsoft.Web.WebView2`可正常引入
`Microsoft.Web.WebView2 1.0.3967.48`会报错

### 期待的结果

正常引入nuget包

### 工程配置

```
add_requires("nuget::Microsoft.Web.WebView2 1.0.3967.48")
```

### 附加信息和错误日志

```
error: @programdir\core\main.lua:274: @programdir\modules\private\action\require\impl\package.lua:461: package(nuget::Microsoft.Web.WebView2): no available schemes with urls and versions
stack traceback:
    [C]: in function 'error'
    [@programdir\core\base\os.lua:1193]: in function 'raiselevel'
    [@programdir\core\sandbox\modules\utils.lua:144]: in function 'assert'
    [@programdir\modules\private\action\require\impl\package.lua:461]: in function '_select_package_version'
    [@programdir\modules\private\action\require\impl\package.lua:1017]: in function '_load_package'
    [@programdir\modules\private\action\require\impl\package.lua:1124]: in function '_load_packages'
    [@programdir\modules\private\action\require\impl\package.lua:1664]: in function 'load_packages'
    [...modules\private\action\require\impl\install_packages.lua:690]: in function '_install_packages'
    [...modules\private\action\require\impl\install_packages.lua:859]:
    [@programdir\modules\private\action\require\install.lua:85]:
    [@programdir\actions\config\main.lua:363]:
    [C]: in function 'xpcall'
    [@programdir\core\base\utils.lua:299]:
    [@programdir\core\base\task.lua:511]: in function 'run'
    [@programdir\core\main.lua:272]: in function 'cotask'
    [@programdir\core\base\scheduler.lua:531]:

stack traceback:
        [C]: in global 'error'
        @programdir\core\base\os.lua:1193: in function 'os.raiselevel'
        (...tail calls...)
        @programdir\core\main.lua:274: in upvalue 'cotask'
        @programdir\core\base\scheduler.lua:531: in function <@programdir\core\base\scheduler.lua:524>
```

bakabaka9405:
[...]

PR #7562, opened 2026-05-25
improve nuget version

waruqi:
https://github.com/xmake-io/xmake/issues/7556#issuecomment-4529016791

### DIFF
--- a/xmake/modules/package/manager/nuget/find_package.lua
+++ b/xmake/modules/package/manager/nuget/find_package.lua
@@ -28,6 +28,19 @@ import("core.project.config")
 import("core.project.target")
 import("private.utils.toolchain", {alias = "toolchain_utils"})
 
+-- convert nuget 4-segment version to semver build metadata
+-- e.g. 1.0.3967.48 -> 1.0.3967+48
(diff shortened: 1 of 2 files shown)


xmake/modules/package/manager/nuget/find_package.lua
```diff
@@ -28,6 +28,19 @@ import("core.project.config")
 import("core.project.target")
 import("private.utils.toolchain", {alias = "toolchain_utils"})
 
+-- convert nuget 4-segment version to semver build metadata
+-- e.g. 1.0.3967.48 -> 1.0.3967+48
+-- @see https://github.com/xmake-io/xmake/issues/7556
+function _to_semver_version(version)
+    if version then
+        local major, minor, patch, build = version:match("^(%d+)%.(%d+)%.(%d+)%.(%d+)$")
```
Wallby: Not sure if it matters to you as developers, but the meaning of these parts in `System.Version` is: major.minor.build.revision
https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-version

Might avoid confusion for future devs to correctly name these.

waruqi: done

Issue #7378, opened 2026-03-06
Improve error/warning handling for invalid package configs and build targets


Willaaaaaaa:
### Is your feature request related to a problem? Please describe.

```xmake
add_rules("mode.debug", "mode.release")

add_requires("ncnn[bulabula]")

target("test")
    add_files("*.cpp")
    add_packages("ncnn")
```

run `xmake` will successfully build the target with no error and warning message. But `bulabula` is an invalid config for `ncnn` package description. Hopefully, when a config is wrongly added, something like a warning/error is given.

And the same checks are hoped to be added in relevant apis like `add_requireconfs()`.

### Describe the solution you'd like

print a warning or raise an error. And although this _can be a language server feature_, I think we'd better provide a more strict and detailed check info for users.

```lua
-- something like this:
wprint("package(xxx): a, b, c configs are ignored, plz check the package description to determine which config(s) you should use")
```

Willaaaaaaa:
# A relevant bug

## Problem

And the same for `xmake build <target>` command. When building an invalid target, xmake won't give any warning/error.

## Current Behavior

```shell
$ xmake build tests  # build any invalid targets
[100%]: build ok, spent 0.007s
```

## Expected behavior:
An invalid target shouldn't be "built ok".
Like `xmake run <target>`, tell users "error: 'xxxx' is not a valid target name for this project."

PR #7584, opened 2026-06-05
add requires api checker

Shiffted:
https://github.com/xmake-io/xmake/issues/7378

Adds checker for `add_requires()` like discussed in https://github.com/xmake-io/xmake/pull/7394

This does not check configs from package deps, e.g.:
```lua
package("foo")
    add_deps("libsdl2", {configs = {invalid_conf = true}})
    on_install(function (package) end)
package_end()

add_requires("foo")
```
will not be checked.

### DIFF
--- a/xmake/modules/private/action/require/impl/package.lua
+++ b/xmake/modules/private/action/require/impl/package.lua
@@ -202,6 +202,56 @@ function _load_package_from_base(package, basename, opt)
     end
 end
 
+-- load the package definition (project, repo, base, system)
+-- @param opt extra options, e.g. {displayname = ..., system = ..., locked_repo = ...}
+-- @return package, packagename, displayname, from_repo
+function _load_package_definition(packagename, requireinfo, opt)
+    opt = opt or {}
+    local displayname = opt.displayname
+
+    -- load package from project first
+    local package
+    if os.isfile(os.projectfile()) then
+        package = _load_package_from_project(packagename)
+        if package and package:namespace() then
+            packagename = package:namespace() .. "::" .. packagename
+            if displayname then
+                displayname = package:namespace() .. "::" .. displayname
+            end
+        end
+    end
+
+    -- load package from repositories
+    local from_repo = false
+    if not package then
+        package = _load_package_from_repository(packagename, {
+            plat = requireinfo.plat,
+            arch = requireinfo.arch,
+            name = requireinfo.reponame,
+            locked_repo = opt.locked_repo})
+        if package then
+            from_repo = true
+        end
+    end
+
+    -- load base package
+    if package and package:get("base") then
+        _load_package_from_base(package, package:get("base"), {
+            name = requireinfo.reponame, locked_repo = opt.locked_repo})
+    end
+
+    -- load package from system
+    local system = requireinfo.system
+    if system == nil then
+        system = opt.system
+    end
+    if not package and (system ~= false or packagename:find("::", 1, true)) then
+        package = _load_package_from_system(packagename)
+    end
+
+    return package, packagename, displayname, from_repo
+end
+
 -- has locked requires?
 function _has_locked_requires(opt)
     opt = opt or {}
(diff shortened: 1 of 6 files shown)


waruqi:
To minimize changes, I don't have time to review large patches.

xmake/modules/private/check/checker.lua
```diff
@@ -30,6 +30,9 @@ function checkers()
             -- package api checkers
             ["api.package.kind"]         = {description = "Check kind configuration in package.", load = true},
             ["api.package.versionfiles"] = {description = "Check versionfiles configuration in package.", download_failure = true},
+            -- requires api checkers
+            ["api.requires.package"]     = {description = "Check package names in add_requires and add_requireconfs."},
+            ["api.requires.configs"]     = {description = "Check package configs in add_requires and add_requireconfs.", load = true},
```
waruqi: Categorized by API name.

api.package.requires
api.package.requireconfs

Shiffted: The package checker does not have the config info and only loads local packages. I could split it into `api.requires.xxx` and `api.requireconfs.xxx`.

---

包检查器没有配置信息，只加载本地包。我可以将其拆分为 `api.requires.xxx` 和 `api.requireconfs.xxx`。

waruqi: 不是，目前的约定是一个 api name 对应一个一个检查，你可以内部分 但文件不用拆这么细了、

Shiffted:
I changed the implementation to minimize changes and added a checker only for `add_requires()`. I will add the `add_requireconfs()` checker in a separate PR.

This now also allows `values` checker to skip instances. The only other previously existing `values` checker always returns a non-nil table, so existing checkers are unaffected.

---

我修改了实现方案，以尽量减少改动，并且仅为 `add_requires()` 添加了检查器。我将在另一个 PR 中添加 `add_requireconfs()` 的检查器。

现在，`values` 检查器也可以跳过实例了。此前唯一存在的另一个 `values` 检查器始终返回一个非空表，因此现有的检查器不受影响。

xmake/modules/private/action/require/impl/package.lua
```diff
@@ -930,45 +980,13 @@ function _load_package(packagename, requireinfo, opt)
...
-    if not package and (system ~= false or packagename:find("::", 1, true)) then
-        package = _load_package_from_system(packagename)
-    end
+    -- load package definition
+    local package, from_repo
+    package, packagename, displayname, from_repo = _load_package_definition(packagename, requireinfo, {
+        displayname = displayname,
+        system = opt.system,
+        locked_repo = locked_requireinfo and locked_requireinfo.repo
+    })
```
waruqi, line 989: 这里为啥要改，好像跟当前 patch 没啥关系把，最小化改动

xmake/modules/private/check/checkers/api/api_checker.lua
```diff
@@ -200,3 +231,39 @@ function check_packages(apiname, opt)
...
+        name = function (self)
+            return require.name
+        end}
+end
+
+-- check api configuration in `add_requires()`
+function check_requires(apiname, opt)
+    opt = opt or {}
+    opt.system = false
+    _check_instances(apiname, nil, function ()
```
waruqi, line 260: 可以把 _check_instances 改成 public api ，然后 requires 相关的都放到 requires.lua 里去

xmake/modules/private/check/checkers/api/api_checker.lua
```diff
@@ -175,6 +179,33 @@ function _check_instances(apiname, instance, instances_func, opt)
...
+            table.join2(sourceinfos, project.get("__sourceinfo_requires"))
+            for _, namespace in ipairs(table.wrap(project.namespaces())) do
+                table.join2(sourceinfos, project.get(namespace .. "::__sourceinfo_requires"))
+            end
+
+            for _, item in ipairs(require_impl.load_requires(table.wrap(requires_str), requires_extra)) do
+                table.insert(requires, {
+                    name = item.name,
+                    info = item.info,
+                    package = require_impl.load_package_definition(item.name, item.info, opt),
```
waruqi, line 199: 最好不要去加载 package ，可能会导致各种 cache 不一致问题，目前 load package 这种时序要求很高，检测阶段去 load，大概率会出一些不可预测的问题

Shiffted: We do need to load packages in some form to know what configs are available I think. I will also need to pass the lock repo info which is currently missing.

---

我认为，我们确实需要以某种形式加载包，才能知道有哪些配置可用。此外，我还需要传递锁定仓库的信息，目前该信息尚缺。

waruqi: No, this will most likely cause many hidden problems. The package API checker itself is called inside the package load, which could be an infinite loop. Due to caching, it may or may not work properly, or it may affect some cache states.

https://github.com/xmake-io/xmake/blob/151db560b3960f677f942357eae8ad3a574e79db/xmake/modules/private/action/require/impl/package.lua#L1098

load package -> check api -> load package

Shiffted: This isn't the `_load_package()` function, its a helper function i created which only contains the `load_package_from_xxx` functions. I can revert that change and put it into the `requires.lua` file to keep the changes minimal.

waruqi: _load_package -> check_api -> check_requires -> _get_requires -> _load_package_definition -> _load_package_from_repository
_load_package -> _load_package_from_repository

Since the cache has now been deleted, this will cause the data to be loaded twice. I'm not sure if there will be any other side effects or performance issues.

xmake/modules/private/check/checkers/api/api_checker.lua
waruqi: move all changes to requires.lua

PR #7586, opened 2026-06-05
Fix show format

Meha555:
Enable `xmake show --format=json`.


Meha555:
I forget to pull newest dev, so this PR mixed with other's commits.

PR #7597, opened 2026-06-10
improve error message for set_toolset with multiple values

Shiffted:
```lua
target("test")
    set_kind("binary")
    add_files("src/*.cpp")
    set_toolset("cxx", "clang", "clang++")
```
would previously result in
```console
error: attempt to call a nil value (method 'find')
```


waruqi:
you should add api checkers in https://github.com/xmake-io/xmake/tree/dev/xmake/modules/private/check/checkers/api/target

waruqi:
try this patch https://github.com/xmake-io/xmake/pull/7630

Shiffted:
> try this patch #7630

Looks good, thanks.

PR #7602, opened 2026-06-11
fix load package memcache

Shiffted:
The cache in the `package.load_from_xxx` functions was not properly set, however I don't think we can cache it here. The instances get mutatet per required package and things like

```lua
add_requires("raylib")
add_requires("raylib~raylib2 >=1.5.1")
```

would not work.

I removed the old cache code and changed it to only cache the description scope info.

---

`package.load_from_xxx` 函数中的缓存未正确设置，但我认为我们无法在此处进行缓存。实例会根据所需包进行修改，例如：

```lua
add_requires("raylib")
add_requires("raylib~raylib2 >=1.5.1")
```

如果进行缓存，上述代码将无法正常工作。

我已移除旧的缓存代码，并将其修改为仅缓存描述作用域信息。

### DIFF
--- a/tests/projects/package/load_from_project/test.lua
+++ b/tests/projects/package/load_from_project/test.lua
@@ -1,0 +1,3 @@
+function main(t)
+    os.exec("xmake l -vD check.lua")
+end
--- a/xmake/core/package/package.lua
+++ b/xmake/core/package/package.lua
@@ -3006,13 +3006,8 @@ end
 -- load the package from the system directories
 function package.load_from_system(packagename)
 
-    -- get it directly from cache first
-    local instance = package._memcache():get2("packages", packagename)
-    if instance then
-        return instance
-    end
-
     -- get package info
+    local instance
     local packageinfo = {}
     local is_thirdparty = false
     if packagename:find("::", 1, true) then
@@ -3073,19 +3068,11 @@ function package.load_from_system(packagename)
         instance:set("parallelize", false)
     end
 
-    -- save instance to the cache
-    package._memcache():set2("packages", instance)
     return instance
 end
 
 -- load the package from the project file
 function package.load_from_project(packagename, project)
-
-    -- get it directly from cache first
-    local instance = package._memcache():get2("packages", packagename)
-    if instance then
-        return instance
-    end
 
     -- load packages (with cache)
     local packages, errors = project.packages()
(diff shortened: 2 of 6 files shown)


xmake/core/package/package.lua
```diff
@@ -3134,55 +3113,59 @@ function package.load_from_repository(packagename, packagedir, opt)
...
 
-    -- new an instance
-    instance = _instance.new(packagename, packageinfo, {scriptdir = path.directory(scriptpath), repo = opt.repo})
+        -- reset plat/arch
+        if opt.plat then
+            package._memcache():set("target_plat", nil)
+        end
+        if opt.arch then
+            package._memcache():set("target_arch", nil)
+        end
```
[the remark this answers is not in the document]
Shiffted: @waruqi

Is this relevant? It was already the case before.

---

这有关系吗？以前就是这样。

waruqi:
I am not sure if this will have other side effects; it could potentially lead to multiple instances of the same package.

Shiffted:
> I am not sure if this will have other side effects; it could potentially lead to multiple instances of the same package.

I don't think this changes anything, both `load_from_repository` and `load_from_system` already created a new scopeinfo on every call before.
`load_from_project` did not but this leads to the following bug:

```lua
package("foo")
    add_configs("baz", {description = "Enable baz.", default = false, type = "boolean"})
    on_install(function(package) end)
    on_load(function(package)
        if package:config("baz") then
            package:add("defines", "FOO_BAZ")
        end
        print(package:get("defines"))
    end)
package_end()

add_requires("foo", {configs = {baz = true}})
add_requires("foo~foo2", {configs = {baz = false}})
```

Currrently for the above both instances will have "FOO_BAZ" defined. With the `:clone()` this is fixed and is now also needed in `load_from_repository` because we cache the scopeinfo. I have also added tests for this now.

tests/projects/package/load_from_project/test.lua
```diff
@@ -0,0 +1,3 @@
+function main(t)
+    os.exec("xmake l -vD check.lua")
```
waruqi, line 2: why not rename check.lua to test.lua directly?

Shiffted: It will try to find the `xmake.lua` in the source root, not the test folder.

PR #7629, opened 2026-06-30
Improve custom source types in Solution Explorer

waruqi:
https://github.com/xmake-io/xmake/pull/7590


usernameak:
Tested it - works with both vstudio and vsxmake generators for me, thanks (and sorry that I couldn't find the time to finish it from my side).

By the way, noticed this comment:
```lua
-- collect the source files not natively supported by vsxmake, and exclude the header/extra files
-- to avoid duplicate display with the include group (add_extrafiles)
```
Interestingly, from my testing - behavior with duplicate items in .vcxproj would be not duplicate display, but a project load error instead.

PR #7636, opened 2026-07-02
Fix cross-target module job ordering for reused BMIs in plain object build

### LINKED ISSUE #7530: build.c++.modules.tryreuse does not create job-level ordering for  .cpp files that import modules from dependency targets

When build.c++.modules.tryreuse is set on a consumer target, cross-target BMI job edges are only created inside build_modules_for_jobgraph (rules/c++/modules/builder.lua), which processes module interface files (.cppm). Regular .cpp implementation files that contain import statements are compiled by the standard object.lua pipeline via on_build_files, which has no connection to those edges.                                                                                                                            
                                                                                                               


Building of regular c++ files in a c++20 modules project should also be included in job ordering with tryreuse is active so that the compilation job only starts when all its required modules have been compiled.

--- 
Claude suggested the following fix, not sure if valid:

When tryreuse is active on a consumer target, the before_build_files phase already accumulates cross-target BMI job edges for .cppm files via _get_jobdeps / the deferred dependent_jobs mechanism. The on_build_files phase (regular .cpp compilation) waits for before_build_files to complete — but only for the current target's own before_build_files group, not for the cross-target BMI jobs that  group depends on. 

The fix: at the end of build_modules_for_jobgraph, after all reuse edges are resolved, insert a synthetic barrier job (e.g. <target>/modules/reused_bmis_ready) that depends on every cross-target BMI job that was added via is_reused. Then order <target>/begin_build (the start of on_build_files) after this barrier. This ensures all reused BMIs — from any dep target — exist before any regular .cpp object file in the consumer starts compiling, without requiring changes to object.lua.
                                                                                                                      


build.fence on provider targets. The current workaround. Correct, but it fully serializes the provider target before any consumer job can start (including linking), eliminating the fine-grained parallelism that tryreuse is designed to provide.                       


### Additional context

_No response_

platontrey:
## Summary

Fixes #7530.

When `build.c++.modules.reuse` is enabled, a regular (non-`.cppm`) source file that imports C++ modules gets "stolen back" into the plain `c++.build` sourcebatch by `rules/c++/modules/scanner.lua` whenever it doesn't itself provide a named module (e.g. a module implementation unit written with a `.cpp` extension, such as `module foo;` without `export`).

That generic jobgraph path (`modules/private/action/build/object.lua`) has no notion of module import dependencies, unlike `rules/c++/modules/builder.lua`'s `build_modules_for_jobgraph`. Under parallel jobgraph scheduling, this lets a reused module's producer job race with the consumer's plain object-file job, intermittently failing with:

```
fatal error: module file 'build/.gens/<Target>/.../SomeModule.pcm' not found: module file not found
```

This PR mirrors `builder.lua`'s `_get_jobdeps` ordering logic (both the immediate `jobgraph:add_orders` path and the deferred `memcache` `"dependent_jobs"` path for producer jobs not yet registered in the graph) inside `object.lua`'s `_add_jobgraph`, scoped to plain source files whose scanned imports are reused from a dependency target.

- Targets without C++ modules are completely unaffected (`modules_support.contains_modules(target)` guard, and the extra imports use `{try = true}`).
- Any unexpected failure in the new lookup only disables the extra ordering for that file (via `try`/`catch`) rather than breaking the build.

## Context / how I found this

I hit this intermittently on a project with ~75 `.cppm` modules shared across 6 targets after enabling `build.c++.modules.reuse` + fixing an unrelated target-feature-flag mismatch (`-msse2` only applied to one target via a package's public cxxflags, not propagated to sibling targets — a separate footgun, happy to file that as its own issue if useful). Once flags were consistent, this job-ordering race was the last remaining failure mode, reproducing on a `.cpp` file that is a module implementation unit (`module arena.vignette;`, no `export`) importing several modules reused from a dependency target.

## Test plan

- [x] Verified the exact failure ("module file '...' not found" for a reused module imported by a plain-`.cpp` module-implementation-unit) no longer reproduces across multiple clean `xmake build -y` runs after this patch, on the project described above (6 targets, ~75 shared modules, `build.c++.modules.reuse` enabled).
- [x] Confirmed targets with no C++ modules are unaffected (guarded by `contains_modules`).
- [ ] Not yet tested against the non-jobgraph `batchjobs` path, MSVC, or a from-scratch minimal reproduction — feedback/review welcome on whether a minimal repro project would help here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/xmake/modules/private/action/build/object.lua
+++ b/xmake/modules/private/action/build/object.lua
@@ -27,6 +27,68 @@ import("private.cache.build_cache")
 import("async.runjobs")
 import("utils.progress")
 import("private.service.distcc_build.client", {alias = "distcc_build_client"})
+import("rules.c++.modules.support", {alias = "modules_support", rootdir = os.programdir(), try = true})
+import("rules.c++.modules.mapper", {alias = "modules_mapper", rootdir = os.programdir(), try = true})
+
+-- @see https://github.com/xmake-io/xmake/issues/7530
+-- When build.c++.modules.reuse is enabled, a regular (non-.cppm) source file
+-- that imports C++ modules gets "stolen back" into this plain c++.build
+-- sourcebatch by rules/c++/modules/scanner.lua whenever it doesn't itself
+-- provide a named module (e.g. a module implementation unit written as a
+-- .cpp file). This generic jobgraph path has no notion of module import
+-- dependencies, unlike rules/c++/modules/builder.lua's build_modules_for_jobgraph,
+-- so under parallel jobgraph scheduling a reused module's producer job can
+-- race with this job and fail with "module file '...' not found". This
+-- mirrors builder.lua's _get_jobdeps ordering logic for that case only;
+-- targets without C++ modules are completely unaffected.
+local function _add_module_reuse_jobdeps(target, jobgraph, sourcefile, jobname)
+    if not (modules_support and modules_mapper) then
+        return
+    end
+    try
+    {
+        function()
+            if not modules_support.contains_modules(target) then
+                return
+            end
+            local module = modules_mapper.get(target, sourcefile)
+            if not (module and module.deps) then
+                return
+            end
+            local moduletype = modules_support.has_two_phase_compilation_support(target) and "bmi" or "onephase"
+            for dep_name, dep in pairs(module.deps) do
+                local dep_key = dep.headerunit and (dep_name .. dep.key) or dep_name
+                local dep_module = modules_mapper.get(target, dep_key)
+                if dep_module then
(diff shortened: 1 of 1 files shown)


waruqi:
I haven't had much time to examine this patch in detail recently, but I believe all C++ module-related changes should be placed in `rules/c++/modules` rather than modifying global internal modules like `xmake/modules/private/action/build/object.lua`. Doing so would lead to increased coupling and make the code harder to maintain.

Additionally, this looks like AI-generated code and does not fully grasp xmake's current code architecture and style.

waruqi:
try this patch. https://github.com/xmake-io/xmake/pull/7692

PR #7643, opened 2026-07-10
fix: avoid stack buffer overflow for long paths in path translation

luadebug:
Investigate https://github.com/xmake-io/xmake/issues/6962

```lua
local pats = {}
for i = 1, 200 do pats[i] = "ThirdParty/any/subdir_" .. i .. "/**" end
path.translate(os.projectdir() .. "/**|" .. table.concat(pats, "|"))  -- crash
```

### DIFF
--- a/core/src/xmake/path/translate.c
+++ b/core/src/xmake/path/translate.c
@@ -69,17 +69,13 @@ tb_int_t xm_path_translate(lua_State *lua) {
             return 1;
         }
         maxn = (tb_size_t)path_size + TB_PATH_MAXN;
-        data = (tb_char_t *)tb_malloc(maxn);
-        tb_check_return_val(data, 0);
+        data = (tb_char_t *)lua_newuserdata(lua, maxn);
     }
     tb_size_t size = tb_path_translate_to(path, (tb_size_t)path_size, data, maxn, normalize);
     if (size) {
         lua_pushlstring(lua, data, (size_t)size);
     } else {
         lua_pushnil(lua);
-    }
-    if (data != buff) {
-        tb_free(data);
     }
     return 1;
 }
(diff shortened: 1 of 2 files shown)


core/src/xmake/path/translate.c
```diff
@@ -53,8 +53,25 @@ tb_int_t xm_path_translate(lua_State *lua) {
...
+    tb_size_t  maxn = sizeof(buff);
+    if (path_size + 1 > maxn) {
+        if (path[0] == '~') {
+            lua_pushnil(lua);
+            return 1;
+        }
+        maxn = (tb_size_t)path_size + TB_PATH_MAXN;
+        data = (tb_char_t *)lua_newuserdata(lua, maxn);
+    }
+    tb_size_t size = tb_path_translate_to(path, (tb_size_t)path_size, data, maxn, normalize);
```
waruqi, line 74: This did not fix the issue with tb_path_translate_to.

I think we should improve `tb_path_translate_to` to check for `maxn` and return 0 for paths longer than `maxn`. Then try using heap memory.

Additionally, for temporary heap buffers, we should use `malloc` instead of `lua_newuserdata`.

```c
    tb_size_t size = tb_path_translate_to(path, (tb_size_t)path_size, data, maxn, normalize);
    if (size) {
    } else {
        maxn = (tb_size_t)path_size + TB_PATH_MAXN;
        data = (tb_char_t *)malloc(maxn);
        size = tb_path_translate_to(path, (tb_size_t)path_size, data, maxn, normalize);
    }

    // push result

    // free data
```

waruqi:
please wait for this patch. https://github.com/tboox/tbox/pull/316

waruqi:
I have improved it. https://github.com/xmake-io/xmake/pull/7644

PR #7676, opened 2026-07-29
fix masm symbol flags for embed/edit levels (A4018)

### LINKED ISSUE #7675: ml.lua maps the embed/edit symbol levels to -Z7/-ZI, which MASM does not support (A4018)
### Xmake Version

3.0.8 (also present on master)

Windows 11, x64, MSVC 14.51 (VS 2026) — but applies to all MASM versions

### Describe Bug

`set_symbols("debug", "embed")` (and `"debug", "edit"`) produces invalid flags for `ml.exe`/`ml64.exe`. MASM only supports `-Zi`/`-Zd` for debug info; `-Z7` and `-ZI` are cl.exe-only flags ([ML and ML64 command-line reference](https://learn.microsoft.com/en-us/cpp/assembler/masm/ml-and-ml64-command-line-reference)).

In `xmake/modules/core/tools/ml.lua` (inherited by `ml64.lua`), `nf_symbols` maps `embed` → `-Z7` and `edit` → `-ZI`, and `mapflags` maps `-g` → `-Z7`.

This fails in two ways, depending on warning settings:

1. **Without `/WX`** (the worse case): ml64 emits `warning A4018: invalid command-line option : /Z7`, **ignores the flag, and exits 0** — the build "succeeds" but the assembled objects silently contain **no debug info at all**, even though the user asked for it.
2. **With warnings-as-errors** (`set_warnings("...", "error")` → `-WX`): A4018 becomes a hard build failure.

This makes `set_symbols("debug", "embed")` unusable project-wide in any codebase with mixed C++/MASM sources (e.g. BLAKE3's assembly kernels) without a per-target override.

### Expected Behavior

MASM's `-Zi` already writes debug info into the object file itself (MASM has no compile-time PDB concept), so semantically `-Zi` *is* the embedded form — `embed` and `edit` should degrade to `-Zi` for ml/ml64:

```lua
-- symbols: masm only supports -Zi/-Zd; -Z7 and -ZI are cl.exe flags
-- that ml/ml64 reject (A4018). -Zi already embeds debug info in the
-- object (masm has no compile-time pdb), so "embed"/"edit" degrade to it.
function nf_symbols(self, levels)
    if hashset.from(levels):has("debug") then
        return {"-Zi"}
    end
end
```

plus `["-g"] = "-Zi"` in `mapflags`. The ARM assembler modules (`armasm_msvc`, `armasm64_msvc`) are separate and unaffected.

I have this fix ready and validated (ml64 14.51: no A4018, builds clean under `-WX`, cl.exe still correctly gets `-Z7` for the C/C++ files) and can submit a PR.

### Project Configuration

```lua
add_rules("mode.debug")
set_symbols("debug", "embed")
set_warnings("allextra", "error") -- makes the failure a hard error; without it the flag is silently dropped

target("t")
    set_kind("binary")
    add_files("src/main.c", "src/answer.asm")
```

`xmake f -m debug -a x64 -p windows && xmake -rv` shows:

```
ml64.exe -c -nologo -Z7 -WX -Fo... src\answer.asm
MASM : warning A4018:invalid command-line option : /Z7
```

bionicbeagle:
Fixes #7675

MASM (`ml.exe`/`ml64.exe`) only supports `-Zi`/`-Zd` for debug info; the cl.exe flags `-Z7` and `-ZI` are rejected with `warning A4018: invalid command-line option` ([ML and ML64 command-line reference](https://learn.microsoft.com/en-us/cpp/assembler/masm/ml-and-ml64-command-line-reference)). Because A4018 is a warning, the outcome depends on warning settings:

- without `-WX`: the flag is silently dropped and the assembled objects contain **no debug info**, despite `set_symbols("debug", "embed")` asking for it
- with `set_warnings(..., "error")`: hard build failure

Since MASM's `-Zi` writes debug info directly into the object file (there is no compile-time PDB for MASM), `-Zi` already is the embedded form — this PR makes the `embed` and `edit` levels degrade to `-Zi` in `nf_symbols`, and fixes the same problem in the gcc-style `mapflags` entry for `-g`.

`ml64.lua` inherits `ml.lua`, so both assemblers are covered; the ARM assembler modules are separate and unaffected.

Verified on ml64 14.51 (VS 2026) with a mixed C/MASM x64 project using `set_symbols("debug", "embed")` + `set_warnings("allextra", "error")`: before — `MASM : warning A4018:invalid command-line option : /Z7` (build failure under `-WX`); after — ml64 receives `-Zi -WX` with no diagnostics, cl.exe still correctly receives `-Z7`, build and run OK.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/xmake/modules/core/tools/ml.lua
+++ b/xmake/modules/core/tools/ml.lua
@@ -35,7 +35,7 @@ function init(self)
     self:set("mapflags",
     {
         -- symbols
-        ["-g"]                      = "-Z7"
+        ["-g"]                      = "-Zi"
     ,   ["-fvisibility=.*"]         = ""
 
         -- warnings
@@ -55,20 +55,14 @@ function init(self)
 end
 
 -- make the symbol flags
+--
+-- masm only supports -Zi/-Zd, the -Z7 and -ZI flags of cl.exe are rejected (A4018).
+-- -Zi already embeds debug info in the object file (masm has no compile-time pdb),
+-- so the "embed" and "edit" levels degrade to it.
 function nf_symbols(self, levels)
-    local flags = nil
-    local values = hashset.from(levels)
-    if values:has("debug") then
-        flags = {}
-        if values:has("edit") then
-            table.insert(flags, "-ZI")
-        elseif values:has("embed") then
-            table.insert(flags, "-Z7")
-        else
-            table.insert(flags, "-Zi")
-        end
+    if hashset.from(levels):has("debug") then
+        return {"-Zi"}
     end
-    return flags
 end
 
 -- make the warning flag


waruqi:
please push to dev branch next.

Issue #7678, opened 2026-07-29
Adding functions to set the running context for packaged appimages (and more)


KaruroChori:
### Is your feature request related to a problem? Please describe.

Adding `set_runenv` and `set_runargs` to xpack. Right now it is extremely annoying to enforce some contextual information for the runnner.  
In the case of app-images I would be forced unpacking the generated one, replacing the AppRun file and pack it back.
All workaround I tested have not been successful sadly. 

### Describe the solution you'd like

Being able to write something like this 
```
xpack("dustbox")
    set_formats("appimage")
    set_title("Dustbox")
    set_description("Interactive Lua playground/sandbox for graphic engine prototyping")
    set_iconfile("assets/package-full/icons/dustbox_256x256.png")
    add_targets("dustbox-bin", "fake-engine")
    set_runenv("DUSTBOX_BASEDIR", "$HERE/usr/share/dustbox")
    set_runargs("--plugin", "$HERE/usr/lib/libfake-engine.so")

    before_package(function (package)
        os.iorun("xmake doxygen srcdir=include")
    end)
```

leading to this entry point for the appimage 

```
#!/bin/bash
HERE="$(dirname "$(readlink -f "${0}")")"
export DUSTBOX_BASEDIR="$HERE/usr/share/dustbox"
exec "$HERE/usr/bin/dustbox" --plugin "$HERE/usr/lib/libfake-engine.so" "$@"
```

And similarly so for other packaging targets.

### Describe alternatives you've considered

As discussed before, the only way I was able to make it work was unpacking, replacing the file and repacking the appimage.

### Additional context

I have a functioning demo implementation covering appimages.

https://github.com/KaruroChori/xmake/tree/feature/ctx-appimage

waruqi:
You can submit a PR directly, but support for other package formats must be added as well, since this is a universal interface.

Casinius:
I submited it for you , you can see [PR7727](https://github.com/xmake-io/xmake/pull/7727) , glad to have more advice. :)

PR #7681, opened 2026-07-30
Adding contextual environment and args for packaged applications

KaruroChori:
It addresses #7678.

It seems like xpack is not being specifically tested in the testsuite or CI as other parts of xmake, so I cannot validate these changes properly, except for appimages.
Also, I have no experience whatsoever with windows and macos; I had to let my LLM of choice go wild on those and hope for the best.

Please, feel free to make changes; I left the branch open for other maintainers.


xmake/plugins/pack/deb/main.lua
```diff
@@ -245,8 +245,17 @@ function _pack_deb(debuild, package)
         end
     end
 
-    -- archive install files
+    -- install launcher wrapper script if runenvs/runargs are set
     local rootdir = package:source_rootdir()
+    local launcher_exe = import("plugins.pack.launcher").main_executable(package)
```
waruqi: move import to head

waruqi:
Did you test all pack formats?

KaruroChori:
No, I was only able to properly implement and test appimages, I just sketched out the rest as requested in the issue, but I was not able to complete it as I had "infrastructural" limitations.
 
Now I am pushing some commits which fixes several problems, introduce proper testing + CI and passes all pack formats I can locally test (and hopefully they will pass CI as well, alongside wix and dmg which cannot be tested locally in any reasonable way)

.github/workflows/pack.yml
```diff
@@ -0,0 +1,75 @@
+name: Pack
+
```
waruqi, line 2: Do not add separate CI configurations; currently, tests are run uniformly across all platform CIs.

https://github.com/xmake-io/xmake/tree/dev/tests/plugins/pack

KaruroChori: My concern is that some packing operations, as far as I understand, cannot be run on all platforms uniformly. Linux can run **most** packing mechanism, and test most of the resulting packages to see if working accordingly.
However testing nsis would require wine deps which are pretty heavy for a CI, while neither wix nor dmg are buildable or trivial to test on linux.
How would you prefer to see that being approached?

waruqi: NSIS and WiX only need to run on Windows, while DMG files only need to be tested and run on macOS.

KaruroChori: Sorry for the delay, I had a spike in workload. I am resuming revision of this PR to move it to the finish line shortly.

I can restrict the testing matrix just for the sake of simplifying the CI, but I can "confirm" that it should be possible for a person with the right local setup to use most of the other combinations even if not covered.
Maybe that is better reflected in documentation in case.

KaruroChori:
Ok it seems to work as "intended" now. 
Mostly, I just need to disable nsis for the mingw runs due to failing deps, but that should be it.

I removed the custom CI jobs and reworked tests to be part of the current test-suite.
I also tidied up the commit history a bit, as most were just rework stuff for the failing CI.

PR #7698, opened 2026-08-08
fix(semver): correct version selection

### LINKED ISSUE #7697: xrepo installs the wrong version of packages
### Xmake Version

3.1.0

macOS 26.5.2

### Describe Bug

xrepo installed the wrong version of a package when a specific version was specified:

```console
$ xrepo install "sqlite3 3.51.0+300"
in xmake-repo:
  -> sqlite3 3.51.0+0 [license:Public Domain]
please input: y (y/n/m)
y
  => download https://sqlite.org/2025/sqlite-autoconf-3510000.tar.gz .. ok
  => install sqlite3 3.51.0+0 .. ok

$ xrepo install "sqlite3 3.53.0+200"
in xmake-repo:
  -> sqlite3 3.53.0+0 [license:Public Domain]
please input: y (y/n/m)
y
  => install sqlite3 3.53.0+0 .. ok
```

### Expected Behavior

sqlite3 3.51.0+300 and 3.53.0+200 should be installed.

### Project Configuration

None.

```console
$ xrepo install -vD "sqlite3 3.51.0+300"
/opt/homebrew/bin/xmake f -c --require=n -vD
checking for platform ... macosx (arm64)
checking for Xcode SDK ... /Applications/Xcode.app (sdk: 26.5, arm64-apple-macos26.5)
checking for zig ... /opt/homebrew/bin/zig
checkinfo: cannot runv(nim --version), No such file or directory
checking for nim ... no
checkinfo: cannot runv(dotnet --version), No such file or directory
checking for dotnet ... no
checking for .NET SDK directory ... no
configure
{
    ndk_stdcxx = true
    arch = arm64
    xcode = /Applications/Xcode.app
    kind = static
    clean = true
    ccache = true
    plat = macosx
    builddir = build
    mode = release
    host = macosx
}
/opt/homebrew/bin/xmake require -v -D -j 12 --extra={system=false} "sqlite3 3.51.0+300"
checking for git ... /usr/bin/git
checking for gzip ... /usr/bin/gzip
checking for tar ... /usr/bin/tar
/usr/bin/git rev-parse HEAD
finding sqlite3 from xmake ..
checking for xmake::sqlite3 ... sqlite3 3.51.0+0
```

apocelipes:
semver_rangen() accepts an exact version as a range, and SemVer comparison ignores build metadata. This can select a different version than requested.

Try exact string matches for versions and tags before falling back to the range selection.

Fixes #7697

### DIFF
--- a/core/src/xmake/semver/select.c
+++ b/core/src/xmake/semver/select.c
@@ -49,6 +49,24 @@ static tb_bool_t xm_semver_is_exact_version(tb_char_t const *version_str, tb_siz
     }
     return tb_false;
 }
+static tb_long_t xm_semver_compare_with_build(semver_t const *left, semver_t const *right) {
+    tb_long_t result = semver_pcmp(left, right);
+    // xmake-repo orders build metadata as package revisions
+    return result ? result : tb_strcmp(left->build.raw ? left->build.raw : "",
+                                       right->build.raw ? right->build.raw : "");
+}
+static semver_t const *xm_semvers_find_newest(semvers_t const *versions) {
+    tb_assert_and_check_return_val(versions && versions->length, tb_null);
+
+    semver_t const *newest = &versions->data[0];
+    tb_size_t i = 0;
+    for (i = 1; i < versions->length; ++i) {
+        if (xm_semver_compare_with_build(&versions->data[i], newest) > 0) {
+            newest = &versions->data[i];
+        }
+    }
+    return newest;
+}
 static tb_bool_t xm_semver_select_from_versions_tags1(
     lua_State *lua, tb_int_t fromidx, semver_t *semver, semver_range_t const *range, semvers_t *matches) {
     // clear matches
@@ -75,22 +93,16 @@ static tb_bool_t xm_semver_select_from_versions_tags1(
     // no matches?
     tb_check_return_val(matches->length, tb_false);
 
-    // sort matches
-    semvers_psort(matches);
-
     // get the newest version
-    semver_t top = semvers_ppop(matches);
+    semver_t const *top = xm_semvers_find_newest(matches);
     lua_createtable(lua, 0, 2);
 
     // return results
-    lua_pushstring(lua, top.raw);
+    lua_pushstring(lua, top->raw);
     lua_setfield(lua, -2, "version");
 
     lua_pushstring(lua, fromidx == 2 ? "version" : "tag");
     lua_setfield(lua, -2, "source");
-
-    // exit the popped semver
-    semver_dtor(&top);
 
     return tb_true;
 }
@@ -167,21 +179,17 @@ static tb_bool_t xm_semver_select_latest_from_versions_tags(lua_State *lua,
     }
     tb_check_return_val(matches->length, tb_false);
 
-    // sort matches
-    semvers_psort(matches);
-
     // get the newest match
-    semver_t top = semvers_ppop(matches);
+    semver_t const *top = xm_semvers_find_newest(matches);
     lua_createtable(lua, 0, 2);
 
     // return results
-    lua_pushstring(lua, top.raw);
+    lua_pushstring(lua, top->raw);
     lua_setfield(lua, -2, "version");
 
     lua_pushstring(lua, fromidx == 2 ? "version" : "tag");
     lua_setfield(lua, -2, "source");
 
-    semver_dtor(&top);
     return tb_true;
 }
 
--- a/core/src/xmake/semver/select.c
+++ b/core/src/xmake/semver/select.c
@@ -216,26 +216,27 @@ tb_int_t xm_semver_select(lua_State *lua) {
         is_range = semver_rangen(&range, range_str, range_len) == 0;
         is_exact = xm_semver_is_exact_version(range_str, range_len);
 
-        // attempt to select version from the versions list first
-        if ((is_exact || !is_range) &&
-            xm_semver_select_from_versions_tags2(lua, 2, range_str, range_len)) {
-            ok = tb_true;
-            break;
-        }
-        if (is_range && xm_semver_select_from_versions_tags1(lua, 2, &semver, &range, &matches)) {
-            ok = tb_true;
-            break;
+        // matching order: versions exact -> tags exact -> versions range -> tags range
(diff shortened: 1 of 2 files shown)


waruqi:
Thanks! The analysis is right — build metadata does not participate in precedence per the semver spec (§10), and a bare version string is itself a valid range for semver_rangen, so 3.53.0+200 fell into range matching and became equivalent to 3.53.0+0. Trying an exact string match first is the correct fix, and I like that all existing fallbacks are preserved (v1.2.3, 1.2, ^1.2.3 still take the old path), and that xm_semver_is_exact_version uses the strict semvern rather than semver_tryn so only a fully valid version takes the new branch.

Two things before we merge:

1. "exact beats fuzzy" should hold across both lists

The current order is:

versions exact → versions range → tags exact → tags range

So if versions contains 3.53.0+0 while tags contains an exact 3.53.0+200, step 2 matches and the exact tag match is never reached. I think the intended priority is:

versions exact → tags exact → versions range → tags range

It is only a reordering of the if blocks. The case is uncommon (both lists populated, only the tag matching exactly), but the current order can still return a build-metadata mismatch when an exact match exists.

2. A loose request still resolves to an arbitrary revision

With the current binary:

loose 3.53.0, versions = {3.53.0+0, 3.53.0+100, 3.53.0+200}   ->  3.53.0+0
loose 3.53.0, versions = {3.53.0+200, 3.53.0+100, 3.53.0+0}   ->  3.53.0+200

Same set, different input order, different result — because the three compare equal and semvers_psort has nothing to break the tie with. In xmake-repo +N is effectively the package revision, so add_requires("sqlite3 3.53.0") should resolve to the newest revision rather than an order-dependent one.

Could you use build metadata as a tiebreaker for otherwise-equal versions in the sort? That makes the result deterministic and matches what users expect. If you would rather keep this PR focused on the exact-request case, that is fine too — please open a follow-up issue so we do not lose it.

Minor: the exact match is a raw strncmp, so requesting 1.2.3+7 when the repo stores v1.2.3+7 still misses and falls back to the range path. Skipping a leading v/= before comparing would cover it — the added tests only exercise the reverse direction (requesting 1.2.3 against a stored v1.2.3+7).

The added test cases look good otherwise.

apocelipes:
@waruqi 

Done. All changes:

1. Reordered to versions exact → tags exact → versions range → tags range
2. Skipping a leading v/= before comparing. This will convert the original equality range ( =1.2.3 ) to an exact match, but it will not affect existing behavior.
3. When `semver_pcmp` returns the same value, compare the metadata using lexicographical order; this ensures that 'v1.2.3+100' is less than 'v1.2.3+200'. This is consistent with the order used in #6955.

waruqi:
Thanks for the follow-ups. Dropping the sort in favor of a linear max scan is the right fix — it also cleans up ownership (nothing is popped, so semvers_dtor frees everything), and it removes the real non-determinism, since semver_pcmp returns 0 for versions differing only in build metadata and qsort is not stable. Guarding the exact-match path with is_exact is also correct — ^1.2.3 and 1.2 still take the range path.

Three things to fix:

1. Build metadata is compared with tb_strcmp, so 1.0.0+9 sorts above 1.0.0+10. xmake-repo already has mkl 2025.0.1+5 vs +14; the +0/+100/+200 families only work because the widths line up. semver_id_t already carries numeric/num — please walk the build id chain and compare numerically when both sides are numeric. Also note build.raw isn't independently terminated; it only works because build metadata is the last component, so a length-bounded compare would be safer.
2. The v/= prefix skipping is unconditional, so it also applies to non-version strings (the !is_range path). A query of next now matches a tag named vnext. It should only normalize when is_exact.
3. The comment "semver comparison ignores build metadata, so compare the normalized strings" doesn't match the code — only the v/= prefix is stripped, not build metadata.

The new tests all use same-width build numbers, which is exactly what hides issue 1; please add select("latest", {"1.0.0+9", "1.0.0+10"}) → 1.0.0+10.

One question: the tests lock in that 3.53.0+999 against {"3.53.0+100"} returns +100. That's spec-conformant, but if build metadata is meant to be the package revision, asking for one revision and getting another is surprising — is that intended?

apocelipes:
> Thanks for the follow-ups. Dropping the sort in favor of a linear max scan is the right fix — it also cleans up ownership (nothing is popped, so semvers_dtor frees everything), and it removes the real non-determinism, since semver_pcmp returns 0 for versions differing only in build metadata and qsort is not stable. Guarding the exact-match path with is_exact is also correct — ^1.2.3 and 1.2 still take the range path.
> 
> Three things to fix:
> 
> 1. Build metadata is compared with tb_strcmp, so 1.0.0+9 sorts above 1.0.0+10. xmake-repo already has mkl 2025.0.1+5 vs +14; the +0/+100/+200 families only work because the widths line up. semver_id_t already carries numeric/num — please walk the build id chain and compare numerically when both sides are numeric. Also note build.raw isn't independently terminated; it only works because build metadata is the last component, so a length-bounded compare would be safer.
> 2. The v/= prefix skipping is unconditional, so it also applies to non-version strings (the !is_range path). A query of next now matches a tag named vnext. It should only normalize when is_exact.
> 3. The comment "semver comparison ignores build metadata, so compare the normalized strings" doesn't match the code — only the v/= prefix is stripped, not build metadata.
> 
> The new tests all use same-width build numbers, which is exactly what hides issue 1; please add select("latest", {"1.0.0+9", "1.0.0+10"}) → 1.0.0+10.
> 

@waruqi  All done.

> One question: the tests lock in that 3.53.0+999 against {"3.53.0+100"} returns +100. That's spec-conformant, but if build metadata is meant to be the package revision, asking for one revision and getting another is surprising — is that intended?

Good point. Returning a different build revision for an explicitly requested revision would be surprising, even though SemVer precedence ignores build metadata.

I’ve changed the behavior so that an exact version query containing build metadata is treated as an exact package revision request. If that revision is not present, selection returns nil instead of falling back to range matching. Therefore:

```lua
 select("3.53.0+999", {"3.53.0+100"}) -- nil
```

Existing behavior is preserved otherwise: an exact version without build metadata can still select the latest build revision of that version, and range/tag queries are unchanged. I’ve updated the tests accordingly.

PR #7715, opened 2026-08-17
improve: skip msvc/vstudio detection for projects with standalone toolchains

Left-Hand:
this pr makes xmake skip the unnecessary msvc/vstudio detection when the project already configures its own standalone toolchains, e.g. `set_toolchains("mingw")`.

currently `platform:check()` always runs the platform default toolchain list (`msvc`, `clang`, ...) at config time, and both the msvc and clang checks trigger the full vstudio detection (vswhere/registry/vcvarsall). that takes a few seconds on every recheck even if we only build with mingw, and on machines without vs it's never cached so it keeps re-running.

i moved the platform check into `check_target_toolchains()` where the target toolchains are already known, and skip the platform default list when all enabled targets have their own standalone toolchains. if any target still relies on the platform toolchains, the check runs as before, so the default msvc workflow is unaffected.

tested with a simple mingw project on windows (no vs installed): config time drops from ~1.1s to ~0.4s, and the "checking for Microsoft Visual Studio" lines are gone. the default project (no `set_toolchains`) still detects vs exactly as before, and `--toolchain=mingw` / rebuild / `-p mingw` all behave the same.

### DIFF
--- a/xmake/actions/config/main.lua
+++ b/xmake/actions/config/main.lua
@@ -25,7 +25,6 @@ import("core.base.hashset")
 import("core.tool.toolchain")
 import("core.project.config")
 import("core.project.project")
-import("core.platform.platform")
 import("private.detect.find_platform")
 import("core.cache.localcache")
 import("core.cache.detectcache")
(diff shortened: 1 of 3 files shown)


waruqi:
Currently, the platform toolchain must be checked, even if the user configures `set_toolchains`, the platform toolchain may not be used. The current patch is too hacky and will affect many things.

PR #7719, opened 2026-08-18
Propagate package vector extensions to consumers

### LINKED ISSUE #7718: Support exporting languages, vectorexts, and forceincludes from package recipes

XMake packages can export common consumer usage requirements:

```lua
package:add("includedirs", ...)
package:add("defines", ...)
package:add("cflags", ...)
package:add("cxxflags", ...)
package:add("links", ...)
package:add("syslinks", ...)
```

However, packages cannot export these portable target requirements:

```lua
package:add("languages", "c++23")
package:add("vectorexts", "sse2")
package:add("forceincludes", "include/package/config.hpp")
```

These calls are accepted without an error or warning, but the values are not serialized into the installed package manifest and do not reach targets using `add_packages()`.

The corresponding target APIs already exist:

```lua
add_languages("c++23")
add_vectorexts("sse2")
add_forceincludes("include/package/config.hpp")
```

Consumers must therefore know and repeat requirements that should belong to the package, particularly when public headers:

- Require a particular language standard.
- Enable code paths for a particular instruction set.
- Require a configuration header to be force-included.

Allow package recipes to export these usage requirements:

```lua
package:add("languages", "c++23")
package:add("vectorexts", "sse2")
package:add("forceincludes", "include/package/config.hpp")
```

Example:

```lua
package("usage-requirements-probe")
    set_kind("library", {headeronly = true})
    set_sourcedir(path.join(os.scriptdir(), "src"))

    on_load(function (package)
        package:add("includedirs", "include")
        package:add("languages", "c++23")
        package:add("vectorexts", "sse2")
        package:add("forceincludes", "include/probe/config.hpp")
    end)

    on_install(function (package)
        os.cp("include", package:installdir())
    end)
```

Consumer:

```lua
add_requires("usage-requirements-probe")

target("example")
    set_kind("binary")
    add_files("main.cpp")
    add_packages("usage-requirements-probe")
```

These properties should:

- Be serialized into the installed package manifest.
- Be returned as package fetch information.
- Be visible through `target:pkg("name"):get(...)`.
- Be applied automatically by `add_packages()`.
- Support package components.
- Propagate through dependent targets when `add_packages(..., {public = true})` is used.
- Work for both locally built and precompiled packages.

Vector extensions should continue using XMake’s existing toolchain-specific mapping.

Forced-include paths should be resolved relative to the installed package directory. Requirements belonging to a package component should only be applied when that component is selected.

It would also be helpful to warn when `package:add()` is given a property that will not be serialized, since unsupported names and spelling mistakes currently become silent no-ops. It's acceptible to warn in verbose mode too, I guess.

Consumers can repeat the requirements manually:

```lua
target("example")
    add_packages("usage-requirements-probe")
    add_languages("c++23")
    add_vectorexts("sse2")
    add_forceincludes("path/to/config.hpp")
```

This duplicates package knowledge in every consumer and makes it easy for projects to use the package incorrectly.

A consumer can currently pass vector extensions through package-binding configuration:

```lua
add_packages("usage-requirements-probe", {
    vectorexts = "sse2"
})
```

This works, but the consumer must still know the package’s requirements.

The package can export raw compiler flags:

```lua
package:add("cxxflags", "-msse2")
```

However, this embeds compiler-specific flags and bypasses XMake’s portable `add_vectorexts()` mapping. The same problem applies to manually exporting language-standard or forced-include flags.

A project-owned wrapper target could apply these properties publicly, but it prevents straightforward standalone `add_requires()`/`add_packages()` consumption and must be recreated by every project.

### Additional context

Tested with:

```text
XMake: v3.1.0+HEAD.96ad28edb
Platform: Windows
Architectures: x86 and x64
Toolchain: clang-cl
```

Some official xmake-repo recipes already attempt to export language requirements using `package:add("languages", ...)`, including:

- [cairomm](https://github.com/xmake-io/xmake-repo/blob/dev/packages/c/cairomm/xmake.lua)
- [glibmm](https://github.com/xmake-io/xmake-repo/blob/dev/packages/g/glibmm/xmake.lua)
- [pangomm](https://github.com/xmake-io/xmake-repo/blob/dev/packages/p/pangomm/xmake.lua)

Northn:
## Summary

Allow package recipes and package components to export these portable usage requirements:

- `vectorexts`

These values are now serialized into installed package metadata, returned when fetching packages, applied by `add_packages()`, and propagated through public target dependencies.

## Changes

- Register `package.add_vectorexts`.
- Preserve package-level and component-level usage requirements.
- Apply requirements only from selected components.
- Propagate the requirements through targets using public package dependencies.
- Add an end-to-end package usage-requirements regression test.
- Verify that selected vector extensions reach consumers while unselected component values do not.

## Testing

- [x] `xmake l tests/run.lua projects/package/usage_requirements`
- [x] Tested on Windows x64 with Visual Studio 2022/MSVC.

Closes #7718

### DIFF
--- a/tests/projects/package/usage_requirements/repo/packages/u/usage-requirements/src/README.md
+++ b/tests/projects/package/usage_requirements/repo/packages/u/usage-requirements/src/README.md
@@ -1,0 +1,1 @@
+Local source fixture for the usage-requirements package tests.
--- a/tests/projects/package/usage_requirements/repo/packages/u/usage-requirements/src/README.md
+++ b/tests/projects/package/usage_requirements/repo/packages/u/usage-requirements/src/README.md
@@ -1,1 +1,0 @@
-Local source fixture for the usage-requirements package tests.
--- a/tests/projects/package/usage_requirements/repo/packages/u/usage-requirements/xmake.lua
+++ b/tests/projects/package/usage_requirements/repo/packages/u/usage-requirements/xmake.lua
@@ -1,0 +1,22 @@
+package("usage-requirements")
+    set_kind("library", {headeronly = true})
+    set_sourcedir(path.join(os.scriptdir(), "src"))
+
+    add_components("enabled")
+    add_components("disabled")
+
+    on_load(function (package)
+        package:add("vectorexts", "avx")
+    end)
+
+    on_component("enabled", function (package, component)
+        component:add("vectorexts", "avx2")
+    end)
+
+    on_component("disabled", function (package, component)
+        component:add("vectorexts", "avx512")
+    end)
+
+    on_install(function (package)
+        os.cp("README.md", package:installdir())
+    end)
--- a/tests/projects/package/usage_requirements/src/main.cpp
+++ b/tests/projects/package/usage_requirements/src/main.cpp
@@ -1,0 +1,4 @@
+int main()
+{
+    return 0;
+}
--- a/tests/projects/package/usage_requirements/src/main.cpp
+++ b/tests/projects/package/usage_requirements/src/main.cpp
@@ -1,4 +1,3 @@
-int main()
-{
+int main() {
     return 0;
 }
--- a/tests/projects/package/usage_requirements/src/usage.cpp
+++ b/tests/projects/package/usage_requirements/src/usage.cpp
@@ -1,0 +1,8 @@
+#if !defined(__AVX2__) && !defined(_M_AVX2) && !defined(__AVX2)
+#    error AVX2 is required by the package
+#endif
+
+int usage()
+{
+    return 0;
+}
--- a/tests/projects/package/usage_requirements/src/usage.cpp
+++ b/tests/projects/package/usage_requirements/src/usage.cpp
@@ -2,7 +2,6 @@
 #    error AVX2 is required by the package
 #endif
 
-int usage()
-{
+int usage() {
     return 0;
 }
--- a/tests/projects/package/usage_requirements/test.lua
+++ b/tests/projects/package/usage_requirements/test.lua
@@ -1,0 +1,5 @@
+function main(t)
+    if os.subarch():startswith("x") or os.subarch() == "i386" then
+        t:build()
+    end
+end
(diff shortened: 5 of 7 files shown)


waruqi:
I don't think we need to support forceincludes. Why not just use `#include` files directly?

Northn:
> I don't think we need to support forceincludes. Why not just use `#include` files directly?

i'm unsure about them too, kinda powerful stuff, should be removed i guess

waruqi:
Furthermore, I don't recommend exporting languages. Unlike `defines`, it doesn't just affect the current package; it has a global impact. If two packages export different languages, such as C++11 and C++20, it will cause conflicts. This is why I haven't supported it.

Users should configure it themselves.

Northn:
> Furthermore, I don't recommend exporting languages. Unlike `defines`, it doesn't just affect the current package; it has a global impact. If two packages export different languages, such as C++11 and C++20, it will cause conflicts. This is why I haven't supported it.
> 
> Users should configure it themselves.

I already know about this and was planning to make a next issue about this, reporting this problem, because right now the `add_languages` doesn't work in the way as it should: it just appends each command so the compiler will pick the last provided value: `-std=c++20 -std=c++17`: the compiler will pick cpp17
But that's not a blocker for this PR.. I guess

waruqi:
No, it will affect the package. Currently, the languages ​​in the package always conflict, and the order cannot be controlled, which is also difficult to resolve. Therefore, I can only restrict the export of languages ​​from a single package.

Northn:
> No, it will affect the package. Currently, the languages ​​in the package always conflict, and the order cannot be controlled, which is also difficult to resolve. Therefore, I can only restrict the export of languages ​​from a single package.

What if we treat the package specified language as the minimal required to build, instead of appending to the targets commands list? And if the compiler/project already have set it lower than the required ones then it would raise an error.

Anyways, the language part isn't even a priority, this PR was initially made to add `vectorexts` stuff, because this one is really useful.

waruqi:
> What if we treat the package specified language as the minimal required to build, instead of appending to the targets commands list? And if the compiler/project already have set it lower than the required ones then it would raise an error.

This is complex and inconsistent with the implementation of behaviors such as add_defines and vectorexts. It requires a lot of special logic and is quite hacky, so we will not consider it for now.

Northn:
Should be fine now, kepy only `vectorexts`.

> This is complex and inconsistent with the implementation of behaviors such as add_defines and vectorexts. It requires a lot of special logic and is quite hacky, so we will not consider it for now.

Should I create another issue reporting the problem?

Issue #7724, opened 2026-08-21
Support referencing external (non-NuGet, non-project) assemblies in C# targets


Omega172:
## Problem

The C# rule's generated `.csproj` only supports two kinds of assembly references:

- `ProjectReference` (via `add_deps` on another xmake target)
- `PackageReference` (via `add_requires("nuget::...")`)

There is currently no way to reference a **prebuilt, external assembly that is neither** - e.g. a third-party SDK's managed DLL that ships outside NuGet (a game's own `Assembly-CSharp.dll` for modding, a vendor SDK, etc). I confirmed this is a hard gap, not just undocumented, by testing every plausible existing mechanism against a real external DLL:

- `add_linkdirs(...)` + `add_links("Name")` (the native C/C++ convention): silently ignored, produces no `Reference` in the generated csproj, compile fails with `CS0246`.
- `add_files("path/to/Foo.dll")`: hard error at config time, `unknown source file`.
- Reading `xmake/rules/csharp/generator/itemgroups.lua`: only three `<ItemGroup>` kinds are ever registered (`Compile`, `ProjectReference`, `PackageReference`) - there's no code path that can produce a `<Reference>` element under any configuration.

## Proposed API

Add `csharp.references`, a list-type `csharp.*` value (same convention as the other list options like `csharp.warnings_as_errors`), taking paths to prebuilt assemblies:

```lua
target("app")
    set_kind("binary")
    add_files("src/*.cs")
    add_values("csharp.references", "path/to/Foo.dll", "path/to/Bar.dll")
```

This generates, for each path, a standard MSBuild reference item:

```xml
<ItemGroup>
  <Reference Include="Foo">
    <HintPath>path/to/Foo.dll</HintPath>
  </Reference>
  <Reference Include="Bar">
    <HintPath>path/to/Bar.dll</HintPath>
  </Reference>
</ItemGroup>
```

- `Include` is derived from the file's basename (`path.basename`), matching what Visual Studio itself writes when you use "Add Reference" on a browsed file.
- Paths are resolved relative to the project directory (matching how source files are already handled) and rendered relative to the generated `.csproj`'s directory (via the existing `_normalize_relative` helper).
- Only `HintPath` is emitted in this first pass; no `Private`/`CopyLocal` control. Happy to add that as a documented option (e.g. `csharp.references_private` or per-entry syntax) as a follow-up if wanted - keeping the initial surface minimal.

## Implementation

I have a tested, working implementation ready:

[...]

PR #7725, opened 2026-08-21
feat(csharp): support referencing external assemblies via csharp.references

Omega172:
## Summary

Adds `csharp.references`, letting a C# target reference a prebuilt external assembly (not a NuGet package, not another xmake target) via a standard `<Reference Include="Name"><HintPath>...</HintPath></Reference>` item in the generated `.csproj`.

Design proposal and rationale: #7724

**Opening as a draft** per CONTRIBUTING.md's requirement that public API additions be proposed and approved in an issue before a PR is opened for merge. The implementation is complete and tested so the diff is concrete for discussion, but I'm not requesting merge until the design in #7724 is discussed.

## What changed

- `xmake/rules/csharp/generator/csproj.lua`: item-group renderer now supports items with nested child elements (needed for `HintPath`, which is a child element, not an attribute). Backward compatible - existing `Compile`/`ProjectReference`/`PackageReference` items don't set `children` and render exactly as before.
- `xmake/rules/csharp/generator/itemgroups.lua`: new `reference` item group, reading `target:values("csharp.references")`.
- `xmake/rules/csharp/xmake.lua`: documents the new option.
- `tests/projects/csharp/assembly_reference/`: new test project referencing a prebuilt fixture library (`lib/MyLib.dll`), verifying the reference resolves at both compile time and run time.

## Test plan

- [x] `xmake l tests/run.lua projects/csharp` on Windows: all 8 C# tests pass (7 existing + 1 new), no regressions.
- [x] Manually verified the generated `.csproj` contains the expected `<Reference>`/`<HintPath>` block.
- [x] Manually verified `xmake run app` actually executes code from the referenced external assembly (not just that it compiles).
- [x] Manually verified the three previously-plausible-but-broken workarounds (`add_linkdirs`+`add_links`, `add_files` on a `.dll`) either silently no-op or hard-error, confirming this was a genuine gap, not a documentation issue.


waruqi:
Does it ready now?

PR #7726, opened 2026-08-21
Improve MinGW toolchain support for Clang and libc++

damster101:
This patch primarily enables support for LLVM libc++ runtimes (c++_static, c++_shared) in the mingw toolchain, and updates C++20 module rules so that `import std;` resolves to LLVM libc++ instead of hardcoded GNU libstdc++ when using Clang on MinGW (llvm-mingw), which caused build failures before. Also, configuring for MinGW on Linux or macOS without a cross-compiler present would previously result in fallback to host-native /usr/bin/gcc. Now unprefixed fallback is only used on Windows/MSYS2/Cygwin hosts. The toolchain check now also automatically selects clang when pointing at llvm-mingw or a clang-only installation, avoiding the need to manually pass clang = true.


xmake/toolchains/mingw/check.lua
```diff
@@ -44,6 +44,24 @@ function main(toolchain)
...
+            if not os.isexec(clang) then
+                return false
+            end
+        elseif use_clang == false then
+            if not os.isexec(gcc) then
+                return false
+            end
+        elseif (not os.isexec(gcc) or (mingw.sdkdir and mingw.sdkdir:find("llvm-mingw", 1, true)))
+            and os.isexec(clang) then
+            toolchain:config_set("clang", true)
```
waruqi: 这里我觉得没必要，写的太复杂，如果 clang config 已经配置，on_load 本身就能自动切，还会做检测。

如果是 llvm-mingw ，直接配置

```lua
    set_toolchains("mingw[clang]@llvm-mingw")
```

就行了。https://github.com/xmake-io/xmake/blob/05b1ec467142a02aad0304f1361c2b59491fffcd/tests/projects/package/toolchain_llvm_mingw/xmake.lua#L7

你这还硬编码检测路径，太 hack 了。

damster101: I agree.

PR #7740, opened 2026-08-27
fix(tty): fix Windows redirected stdout/stderr and enable 256 colors on Win10+

luadebug:
When `tbox` is built without `TB_CONFIG_FORCE_UTF8`, redirected stdout/stderr streams on Windows (e.g. in CI / pipes) fail to initialize because `tbox` lacks an ANSI => Console Output CodePage (COCP) charset converter. This causes `tb_stdfile_write` to silently drop output chunks.
I came up with this workaround because of https://github.com/conda-forge/xmake-feedstock/pull/2 github CI workflows were unable to print XMake TTY answers (xmake --version the text color, text itself was missing) in github CI workflow logs, so I placed this workaround and now its being show up on https://github.com/conda-forge/xmake-feedstock/actions/runs/33080533205/job/98546199488 
image

### DIFF
--- a/core/src/xmake/io/file_write.c
+++ b/core/src/xmake/io/file_write.c
@@ -109,8 +109,36 @@ static tb_void_t xm_io_file_write_std(xm_io_file_t *file, tb_byte_t const *data,
     tb_size_t type = (file->type & ~XM_IO_FILE_FLAG_TTY);
     tb_check_return(type != XM_IO_FILE_TYPE_STDIN);
 
+#if defined(TB_CONFIG_OS_WINDOWS)
+    HANDLE handle = GetStdHandle(type == XM_IO_FILE_TYPE_STDERR ? STD_ERROR_HANDLE : STD_OUTPUT_HANDLE);
+    if (handle != INVALID_HANDLE_VALUE && handle != NULL) {
+        DWORD mode;
+        if (GetConsoleMode(handle, &mode)) {
+            tb_size_t wsize = (size + 1) * sizeof(tb_wchar_t);
+            tb_wchar_t *wbuf = (tb_wchar_t *)tb_malloc(wsize);
+            if (wbuf) {
+                int wlen = MultiByteToWideChar(CP_UTF8, 0, (char const *)data, (int)size, (wchar_t *)wbuf, (int)(wsize / sizeof(tb_wchar_t)));
+                if (wlen > 0) {
+                    DWORD written = 0;
+                    WriteConsoleW(handle, (wchar_t *)wbuf, (DWORD)wlen, &written, NULL);
+                }
+                tb_free(wbuf);
+            }
+        } else {
+            DWORD written = 0;
+            WriteFile(handle, data, (DWORD)size, &written, NULL);
+        }
+    }
+#else
     // write data to stdout/stderr
-    tb_stdfile_write(file->u.std_ref, data, size);
+    if (file->u.std_ref) {
+        tb_stdfile_write(file->u.std_ref, data, size);
+    } else {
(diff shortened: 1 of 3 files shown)


core/src/xmake/io/file_write.c
```diff
@@ -109,8 +109,36 @@ static tb_void_t xm_io_file_write_std(xm_io_file_t *file, tb_byte_t const *data,
     tb_size_t type = (file->type & ~XM_IO_FILE_FLAG_TTY);
     tb_check_return(type != XM_IO_FILE_TYPE_STDIN);
 
+#if defined(TB_CONFIG_OS_WINDOWS)
```
waruqi: Why not improve tb_stdfile_write and use a more unified interface?

core/src/xmake/io/file_write.c
```diff
@@ -109,8 +109,36 @@ static tb_void_t xm_io_file_write_std(xm_io_file_t *file, tb_byte_t const *data,
...
+    if (handle != INVALID_HANDLE_VALUE && handle != NULL) {
+        DWORD mode;
+        if (GetConsoleMode(handle, &mode)) {
+            tb_size_t wsize = (size + 1) * sizeof(tb_wchar_t);
+            tb_wchar_t *wbuf = (tb_wchar_t *)tb_malloc(wsize);
+            if (wbuf) {
+                int wlen = MultiByteToWideChar(CP_UTF8, 0, (char const *)data, (int)size, (wchar_t *)wbuf, (int)(wsize / sizeof(tb_wchar_t)));
+                if (wlen > 0) {
+                    DWORD written = 0;
+                    WriteConsoleW(handle, (wchar_t *)wbuf, (DWORD)wlen, &written, NULL);
```
waruqi: There are already similar implementations, and I'm not sure what the difference is.

https://github.com/tboox/tbox/blob/b3af9a9e3a714eae58e8aea82f1479592c567c74/src/tbox/platform/windows/stdfile.c#L138-L148

xmake/core/base/tty.lua
```diff
@@ -584,7 +584,12 @@ function tty.has_color256()
         -- detect it from system
         if has_color256 == nil then
             if os.host() == "windows" then
-                has_color256 = false
+                local winos = require("base/winos")
+                if winos.version():lt("win10") then
+                    has_color256 = false
+                else
+                    has_color256 = true
+                end
```
luadebug, line 592: I think to keep this part, as it is solution for text color in console. But revert all another changes and provide tbox changes in `tbox/src/tbox/platform/windows/stdfile.c` try to

 ```c
            if (is_console)
            {
                file->ofstream = tb_stream_init_filter_from_charset(file->ostream, encoding, TB_CHARSET_TYPE_UTF16 | TB_CHARSET_TYPE_LE);
                tb_assert_and_check_break(file->ofstream);
            }
            else if (encoding == TB_CHARSET_TYPE_UTF8)
            {
                // redirect (console output cp)
                file->ofstream = tb_stream_init_filter_from_charset(file->ostream, encoding, TB_CHARSET_TYPE_COCP);
                tb_assert_and_check_break(file->ofstream);
            }
            else
            {
                // no charset filter needed for ANSI passthrough
                file->ofstream = file->ostream;
            }
 ```

But I dont know if it would help or not need some testing.

luadebug: It looks like this works! According to https://github.com/conda-forge/xmake-feedstock/pull/2 and already merged https://github.com/conda-forge/tbox-feedstock/pull/9

So I made PR for tbox https://github.com/tboox/tbox/pull/321

I am not sure this solution going with has_color256 = true is good maybe I can just use env var in CI.

PR #7755, opened 2026-09-07
abort config when platform check fails

damster101:


waruqi:
why? This could break many projects, as the platform toolchain is not used if the user has set `set_toolchains` globally.

damster101:
> why? This could break many projects, as the platform toolchain is not used if the user has set `set_toolchains` globally.

Ooh, I had not considered that `instance_plat:check()` in `actions/config/main.lua` only inspects the platform defaults and CLI `--toolchain` parameter, and is unaware of `set_toolchains()` declared at global scope or on targets in `xmake.lua`... Aborting unconditionally there could indeed break projects when the host/platform default toolchain is absent...
It wasn't meant to improve anything too important, so I'm just gonna close.

PR #7756, opened 2026-09-08
restrict mingw toolchain to mingw platform

damster101:


waruqi:
Why? Some users may customize their platform based on the MinGW toolchain. The MinGW platform is always tied to the MinGW toolchain, but the MinGW toolchain is not necessarily only used with the built-in MinGW platform.

damster101:
> Why? Some users may customize their platform based on the MinGW toolchain. The MinGW platform is always tied to the MinGW toolchain, but the MinGW toolchain is not necessarily only used with the built-in MinGW platform.

That makes sense... I had not considered custom platforms reusing the MinGW toolchain as a possibility. Thanks for clarifying. Closing this.

PR #7758, opened 2026-09-08
add downloader option for download tasks and do not use aria2 UserAgent

luadebug:
https://github.com/xmake-io/xmake-repo/pull/10794

https://github.com/xmake-io/xmake-repo/actions/runs/34200510849/job/101977977356?pr=10794

403 - Forbidden it seems Cloudflare does not like aria2 UserAgent maybe?

### DIFF
--- a/xmake/actions/require/xmake.lua
+++ b/xmake/actions/require/xmake.lua
@@ -99,6 +99,8 @@ task("require")
             {nil, "packagedir",  "kv", "packages", "Set the packages directory for exporting, importing and downloading."},
             {nil, "debugdir",    "kv", nil,        "Set the source directory of the current package for debugging."},
             {nil, "extra",       "kv", nil,        "Set the extra info of packages."},
+            {nil, "downloader",  "kv", nil,        "Set the download tool.",
+                                                   values = {"aria2", "curl", "wget", "powershell", "pwsh"}},
             {},
             {nil, "requires",    "vs", nil,        "The package requires.",
                                                    "e.g.",
--- a/xmake/modules/net/http/download.lua
+++ b/xmake/modules/net/http/download.lua
@@ -299,103 +299,80 @@ function _powershell_download(tool, url, outputfile, opt)
     os.vrunv(tool.program, argv)
 end
 
--- downloaders mapping
-local _DOWNLOADERS = {
-    aria2 = _aria2_download,
-    curl = _curl_download,
-    wget = _wget_download,
-    powershell = _powershell_download,
-    pwsh = _powershell_download
-}
-
--- find download tool by name
-function _find_download_tool(name)
-    if name == "powershell" then
-        if is_host("windows") then
-            return find_tool("pwsh") or find_tool("powershell")
+-- get user-configured downloader
+function _get_downloader(opt)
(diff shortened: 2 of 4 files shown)


xmake/modules/net/http/download.lua
```diff
@@ -297,36 +299,103 @@ function _powershell_download(tool, url, outputfile, opt)
...
 
--- download url with the first available tool (aria2/curl/wget/powershell)
-function _download(url, outputfile, opt)
-
-    -- attempt to download url using aria2 first (multi-threaded, fastest)
-    local tool = find_tool("aria2", {version = true})
-    if tool then
-        return _aria2_download(tool, url, outputfile, opt)
+-- downloaders mapping
+local _DOWNLOADERS = {
```
waruqi: do not define global vars.

xmake/modules/net/http/download.lua
```diff
@@ -297,36 +299,103 @@ function _powershell_download(tool, url, outputfile, opt)
...
+        end
+
+        -- clean up partial outputfile and aria2 control files before next attempt
+        os.tryrm(outputfile)
+        os.tryrm(outputfile .. ".aria2")
+
+        -- fallback to next available tool
+        local next_tool = tools[i + 1]
+        if next_tool then
+            wprint("downloading %s with %s failed, falling back to %s ..", url, tool_info.name, next_tool.name)
```
waruqi: `wprint` does not output in real-time—which isn't necessary—or you could switch to `dprint` for diagnostic logging.

xmake/modules/net/http/download.lua
```diff
@@ -297,36 +299,103 @@ function _powershell_download(tool, url, outputfile, opt)
...
+            local download_fn = _DOWNLOADERS[name]
+            if download_fn then
+                table.insert(tools, {name = name, tool = tool, download = download_fn})
+            end
+        end
+    end
+    assert(#tools > 0, "no available download tool found (%s)!", table.concat(downloaders, ", "))
+
+    local errors = {}
+    for i, tool_info in ipairs(tools) do
```
waruqi: I consider this unnecessary, as it can result in invalid or unstable links; using different downloaders repeatedly can interfere with the package's download and installation. Since the package configuration already specifies a mirror URL, it is usually sufficient to simply verify connectivity using a single, reliable downloader.

luadebug:
@waruqi thank you for review. I hope this would get us closer to resolve https://repology.org/repository/xrepo/problems since websites often use Cloudflare to protect, and it is often checking for UserAgent I guess.

xmake/actions/require/xmake.lua
```diff
@@ -99,6 +99,8 @@ task("require")
             {nil, "packagedir",  "kv", "packages", "Set the packages directory for exporting, importing and downloading."},
             {nil, "debugdir",    "kv", nil,        "Set the source directory of the current package for debugging."},
             {nil, "extra",       "kv", nil,        "Set the extra info of packages."},
+            {nil, "downloader",  "kv", nil,        "Set the download tool.",
```
waruqi, line 102: we need not set it in require menu

Issue #7720, opened 2026-08-20
Xmake selects the wrong MSVC version


aubymori:
### Xmake Version

3.1.0+master.de769d557

### Operating System Version and Architecture

Windows 10 version 21H2

### Describe Bug

Xmake selects the wrong MSVC toolkit version when attempting to find MSVC. I had a complete toolkit `14.51.36231` set as the default via `C:\Program Files\Microsoft Visual Studio\18\Community\VC\Auxiliary\Build\Microsoft.VCToolsVersion.v145.default.txt`. Xmake picked the higher version `14.52.36615`, which did not have all files needed, causing it to fail to find MSVC.

### Expected Behavior

It should pick the default version, or at the very least, check that all the needed tools are there and move on to the next highest version if they are not.

### Project Configuration

```lua
add_rules("mode.debug", "mode.release")

target("xmaketest")
    set_toolchains("msvc")
    set_kind("binary")
    add_files("src/*.c")
```

### Additional Information and Error Logs

```console
C:\Users\Aubrey\src\xmaketest>xmake f -a x86
checking for platform ... windows (x86)
Detecting Visual Studio environment (cached); run `xmake global --clean` first if vs updated.
checking for Microsoft Visual Studio (x86) version ... no
checking for LLVM Clang C/C++ Compiler (x86) ... ok
checking for Microsoft Visual Studio (x86) version ... 2022
error: toolchain("msvc"): not found!
```

Moving the incomplete version out of the `C:\Program Files\Microsoft Visual Studio\18\Community\VC\Tools\MSVC` directory results in it finding 14.51.36231 just fine.

waruqi:
you can debug https://github.com/xmake-io/xmake/blob/dev/xmake/modules/detect/sdks/find_vstudio.lua

PR #7761, opened 2026-09-10
Keep the default MSVC environment when a newer toolset is incomplete

Hanabi9248:
Related to #7720.

After loading the default MSVC environment, detection tries the newest installed toolset. If that attempt returns an environment without INCLUDE or LIB, it currently replaces the usable default and downstream toolchain checks reject it.

Keep the default unless the newer result contains PATH, INCLUDE and LIB, matching the existing toolchain checks. Copy the retry options so automatic selection does not become an explicit toolset request for subsequent architectures. A valid newer environment still takes precedence.

Added a Windows batch fixture through the public detector, covering missing INCLUDE, missing LIB, successful upgrade and an explicitly selected older version. It runs in a child Xmake process with an isolated detection cache. `xmake lua tests/run.lua modules/detect` passes with Xmake 3.1.1 on Windows; the regression fails before the change. No real compiler build or full Xmake suite was run. This does not address an installation whose environment is complete but whose compiler executable is missing.

### DIFF
--- a/xmake/modules/detect/sdks/find_vstudio.lua
+++ b/xmake/modules/detect/sdks/find_vstudio.lua
@@ -321,7 +321,7 @@ function _load_vcvarsall_impl(vcvarsall, vsver, arch, opt)
 
     -- check if the environment variables are truncated
     _check_vcvarsall_env(variables)
-    if not variables.PATH then
+    if not (variables.PATH and variables.INCLUDE and variables.LIB) then
         return
     end
 
@@ -435,7 +435,7 @@ function _load_vcvarsall(vcvarsall, vsver, arch, opt)
             local latest_opt = table.clone(opt)
             latest_opt.toolset = _strip_toolset_ver(latest_toolset)
             local latest_result = _load_vcvarsall_impl(vcvarsall, vsver, arch, latest_opt)
-            if latest_result and latest_result.PATH and latest_result.INCLUDE and latest_result.LIB then
+            if latest_result then
                 result = latest_result
             end
         end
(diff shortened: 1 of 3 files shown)


xmake/modules/detect/sdks/find_vstudio.lua
```diff
@@ -432,8 +432,12 @@ function _load_vcvarsall(vcvarsall, vsver, arch, opt)
             end
         end
         if latest_toolset and VCToolsVersion and semver.compare(latest_toolset, VCToolsVersion) > 0 then
-            opt.toolset = _strip_toolset_ver(latest_toolset)
-            result = _load_vcvarsall_impl(vcvarsall, vsver, arch, opt)
+            local latest_opt = table.clone(opt)
+            latest_opt.toolset = _strip_toolset_ver(latest_toolset)
+            local latest_result = _load_vcvarsall_impl(vcvarsall, vsver, arch, latest_opt)
```
waruqi, line 437: 为什么不直接在 _load_vcvarsall_impl 里面检测，如果不完整 直接返回 nil

Hanabi9248:
已移到 `_load_vcvarsall_impl` 中检查，缺少 PATH、INCLUDE 或 LIB 时直接返回 nil；外层只处理新版本加载失败后的回退。

同时补了显式指定新工具集时的用例：环境不完整应返回 nil，完整时仍使用指定版本。Windows 下 `xmake lua tests/run.lua modules/detect` 通过；新增用例在修改前会因接受了不完整环境而失败。

PR #7772, opened 2026-09-17
qt: avoid debug suffix for Harmony libraries

KnightJun:
When using `qt.shared` with the Harmony platform in debug mode, xmake generates Qt library names with the `_debug` suffix:

- `Qt5Core_debug`
- `Qt5Gui_debug`
- `Qt5Widgets_debug`

However, the Harmony Qt SDK uses Unix-style library names without the debug suffix:

- `libQt5Core.so`
- `libQt5Gui.so`
- `libQt5Widgets.so`

This causes the linker to fail because the generated library names do not exist.


xmake/rules/qt/load.lua
```diff
@@ -37,7 +37,7 @@ function _link(target, linkdirs, framework, qt_sdkver, infix)
             else
                 debug_suffix = "d"
             end
-        elseif target:is_plat("android") or target:is_plat("linux") or target:is_plat("cross") then
+        elseif target:is_plat("android") or target:is_plat("linux") or target:is_plat("cross") or target:is_plat("harmony") then
```
waruqi: target:is_plat("android", "linux", "cross", "harmony")

PR #7775, opened 2026-09-17
Better support for wasi outside of the browser

### LINKED ISSUE #7774: Better support for wasm/wasm-wasi

Right now support for these targets is not ideal and very web-centric.
I have a set of working patches I have been using over the last month to get "parity" on these platforms.

Proposals:

- To add toolchain arch `wasm32` and `wasm64`.
- To introduce crossplatform target `wasm-wasi` like we do already for any other native platform supported.

There is none to be honest, proper wasi support is just missing outside the assumption of web usage.

### Additional context

I will be linking a PR shortly.

KaruroChori:
It closes #7774

Basically, some patches I added to make wasi usable zia zigcc on my system.
I further added some testing infrastructure and guards for CI, as I did not want to introduce a dependency for a wasm runtime for all runners (but it is needed to fully test this path if not running locally).

This allows to output clean wasm files without all the html bundle.


waruqi:
We are not considering adding support for the WASI platform, as this would increase the maintenance workload for xmake-repo packages and the demand on CI testing resources. Currently, all existing WASM packages are incompatible with WASI.

KaruroChori:
I understand you don't want to get an endless stream of bug reports for what is well beyond your control, and all things considered, a pretty minor target. Still, I sincerely think there are ways to integrate such support without the need of extra workload on your end nor more complex CI pipelines.
If you mind is already set on that, I cannot do much and I will just run my internal fork of xmake.
It creates a lot of friction in distributing software which is wasm-centric, but other native builds would still work fine for the "general" public, so it could be worse :smile: .

Still, if you have some time to spare, here are some of my thoughts.
I don't think anyone should expect every single xmake-repo package to work in wasm-wasi or any other arch triplet to be honest. Using syscalls outside posix is already enough to prevent it from working outside linux, and just the presence of inline assembly disallows portability on less common architectures, unless fallbacks are added explicitly. 
Many common libraries will handle this fine, but I doubt this extends to all of them.

One thing I really like about xmake-repo, is how easy it is for anyone to manage their own registries of packages. I do that a lot to provide our public and private libraries, as well as to repackage many small libraries which would never match the criteria for the main registry, but we use internally. 
I think it would be reasonable to explicitly guard some platforms simply as "not supported" at the registry level, preventing compilation for "bad targets" and avoiding any downstream confusion for users.
This I assume would result in no extra bug reports, CI or packaging effort, while keeping the core of xmake more flexible and allowing its users to manage their own registries as they need.

waruqi:
Many packages include logic to detect the WebAssembly (Wasm) platform—for example:

https://github.com/xmake-io/xmake-repo/blob/eef02822a0c266883394db1b2e18a07f30e7c17e/packages/b/binn/xmake.lua#L23

https://github.com/xmake-io/xmake-repo/blob/eef02822a0c266883394db1b2e18a07f30e7c17e/packages/c/ctrl-c/xmake.lua#L11

Currently, a large number of packages rely on explicit WASM platform checks; adding a new platform would double the workload and maintenance effort.

PR #7783, opened 2026-09-19
c++ modules: recompute module DAG and sourcebatch on objectdir change

damster101:
Switching build mode, platform, architecture, or output directory should invalidate the module dependency DAG and object file paths so they are recomputed, avoiding build failure from stale paths from a previous configuration.


waruqi:
try this patch. https://github.com/xmake-io/xmake/pull/7788

damster101:
> try this patch. #7788

Yes, I think that's sufficient. And more optimal.

